PR #68, opened 2020-09-11
Fix some errors under Tweakfiles section

Faalagorn:
* Use block of code instead of single line code for the example
* `PROTON_NO_ESYNC` was stated twice instead of `FSYNC`


frostworx:
Thank you! :)

PR #107, opened 2020-10-16
RU-Localisation update

IDeathByte:
Also i see lost checkbox name in all locales in Game and Default settings tabs:
lost_check

Now i have installed version from AUR and i didn`t see fixed it or not


frostworx:
thanks for both the update and the hint about the missing checkbox.
will fix it soon.

Issue #116, opened 2020-10-24
Compatibilitytools.d is in another directory on Debian Buster


kadogo:
Hello at first thanks for your big work on STL.

It make me discover applications I didn't know and it really ease all the configuration. I just start playing with it and it's already pretty amazing.

It's not a big issue, I will self make a PR myself later if you want but I just want to be sure it's not an issue from my setup (it's a Debian Buster with some unstable stuff so it's not really the most stable system).

When I did `stl compat add` it installed STL in `/home/user/.steam/compatibilitytools.d/` instead of `/home/user/.steam/root/compatibilitytools.d/`. I must move it with the older compatibiitytools from Steam so that it could be detected.

I not know if it's related to may setup or maybe because I had old version of compatibilitytools but if you think it's normal I can take a look to add a check if the directory already exist when we use `stl compat add`

Have a good evening.
Cheers

frostworx:
Hello!
Thanks a lot for the friendly feedback! Despite the amount of stars it is sometimes hard to tell if people like or even use my program, because everything is pretty silent.

The absolute path `"$SCTS"` which `stl compat add` currently uses is
`~/.steam/steam/compatibilitytools.d/SteamTinkerLaunch/`.

_(which is set up from `"$STEAMCOMPATOOLS/$PROGNAME"`,
where `"$STEAMCOMPATOOLS"` is `"$HSS/compatibilitytools.d"` and `"$HSS"` in turn is **`"$HOME/.steam/steam"`**
and `"$PROGNAME"` is SteamTinkerLaunch)_

On my system all directories but "steamcmd" in "$HOME/.steam" are symlinks _(I'd say autocreated)_
and both `"$HOME/.steam/steam"` and `"$HOME/.steam/root"` point to `"$HOME/.local/share/Steam"`
where `"compatibilitytools.d"` lies. Not unlikely that all those confusing symlinks are the cause of the issue here.

Hard to tell if using `"$HOME/.steam/steam"` for `"$HSS"` is the/an official way to go, but if using `"$HOME/.steam/root"`
for `"$HSS"` would fix it for you, I'd be happy to change it.
Would be nice if you would test if that fixes the issue, by changing the variable here:
https://github.com/frostworx/steamtinkerlaunch/blob/master/stl#L128
Have a nice sunday.
cheers

frostworx:
I just found 
[...]

PR #121, opened 2020-10-31
Make mangohud variable export the libs

kadogo:
Hello,

# Changes

I not think the MangoHud was working properly because I didn't find where `LD_PRELOAD` and `LD_LIBRARY_PATH` where updated so that MangoHud was used.

I added in the global.conf a way to define where the libraries are in prevision of the use of https://github.com/kadogo/easy-builder that will (I hope) let anyone build the library or use them easily without any installation.

I added before all the test the loading of the libs if `MANGOHUD` is equal to 1 and that add the DLSYM lib if it's set to 1 in the game configuration. Maybe it could be skipped in at some time but I'm not sure where it work or not yet (Steam Linux Runtime, Proton...). If I understand the code there are warnings if there are a chance that it not work.

In the game configuration I added `MANGOHUD_DLSYM` and `MANGOHUDARCH`. The first one is used for some OpenGL games if I understand it correctly and the second is used to load the right library. In the MangoHud wrapper they detect the architecture automatically by doing multiple symlink (https://github.com/flightlessmango/MangoHud/blob/master/bin/mangohud-setup.sh#L57).

# Possible improvement

Maybe we could avoid the `MANGOHUDARCH` by using the auto-detection if MangoHud is installed.
I just added the variables for the configuration file but not in the GUI because I don't know zenity/yad yet.

If you want make modification you are free to do.
Cheers.


frostworx:
Thanks, I will review it in the middle of next week, as I'm offline for some days.
Already checked it quickly and have some questions (likely I missed something):

- I assume 32 bit MangoHUD is used for 32bit games right? Would be nice to automatically pick the right one by using an exe architecture auto-detection (like stl does in the ReShade installation using file`)`
- Where do you disable MangoHUD per game?
- Would be nice to have the new translation lines at the bottom of every language file, to make translation a bit easier.
- Maybe more when I reviewed the patches

kadogo:
Sure take your time no problem

> I assume 32 bit MangoHUD is used for 32bit games right? Would be nice to automatically pick the right one by using an exe architecture auto-detection (like stl does in the ReShade installation using file`)`

I see 2 issues and another where defining the lib is still helping and 1 I didn't thought about it and that the user will need to check.

*  I had a first version with file but I had an issue with "Baba is you" where we launch a script instead of a binary (we can still make it by verify if the file is have an '64' as guess with a way to override it with `MANGOHUDARCH` ?)

```
# The game script define if we use one or other
file ~/SteamLibrary/steamapps/common/Baba\ Is\ You/run.sh 
/home/user/SteamLibrary/steamapps/common/Baba Is You/run.sh: POSIX shell script, ASCII text executable

file ~/SteamLibrary/steamapps/common/Baba\ Is\ You/bin64/Chowdren 
/home/user/SteamLibrary/steamapps/common/Baba Is You/bin64/Chowdren: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.24, BuildID[sha1]=36c0f2cd9bfe3e9007b39cc0c23362e7aae7e015, stripped

file ~/SteamLibrary/steamapps/common/Baba\ Is\ You/bin32/Chowdren 
/home/user/SteamLibrary/steamapps/common/Baba Is You/bin32/Chowdren: ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.24, BuildID[sha1]=38d2a261dccf4d8caf592ae9a50151b460fdc0aa, stripped
```

*  The games that are make LD_PRELOAD or LD_LIBRARY_PATH empty. I'm not sure there are much like that but it would be the user work to make a bug or check the script and maybe use a `CUSTOMCMD` (didn't test yet).

> Where do you disable MangoHUD per game?

It's still done with `MANGAHUD`. The loading is happening only if it's set to 1.

> Would be nice to have the new translation lines at the bottom of every language file, to make translation a bit easier.

I can already update that in a further commit so it would be already done.

> Maybe more when I reviewed the patches

Let me know and if you have some idea too. For now it's more like a proof-of-concept that is a bit rough on the edge ^^

frostworx:
Heh, I really didn't read through it carefully and thought you removed the `MANGAHUD` option.
I consider adding a generic `getArch` function which checks file `$1` and returns either 32 or 64.
Then, if `$1` is none of both an additional check could be called which checks usual suspect neighbor files in the `dirname` of `$1`
(f.e. dll or exe files). The success rate probably will be pretty high and it would help both MangoHUD and ReShade.
Do you have an example for games emptying LD_PRELOAD and LD_LIBRARY_PATH variables?
A little check for the correct path for MANGOHUD_LIB and MANGOHUD_LIB32 would be nice instead of initially hardcoding them (and I would fallback to the default `/usr/lib/` and `/usr/lib32/` if not found :) )

kadogo:
Sure the getArch could make it easier (but maybe it would not be needed anymore).

I remark that it didn't work on Proton games and vkcube. So I guess I have somekind of issue related to the vulkan library. I will try to have some answers on the MangoHud repo. The good news is that it work on opengl games and glxgears without the need of the architecture variable.

For the game that could empty LD_PRELOAD and LD_LIBRARY_PATH it was just mentionned in MangoHud repo but I didn't saw one for now. I will keep you updated if I find the issue about the vkcube part.

kadogo:
I fixed for the AppImage it needed the correct vulklan implicit layer JSON and few environement variables so that vulkan detect where are the implicit layer and which he needs to activate MangoHud.

I think we will need to copy the implicit vulkan layer (or asking the user to copy it) if they want to use the libs without any installation.

frostworx:
*back*
I planned to improve the "architecture detection" anyway, so no matter if it is of any use for MangoHUD I'll rewrite it.
I'll leave this pull request open until above issues are ready for merge.
steamtinkerlaunch won't move or copy around any vulkan layers for MangoHUD, so need to try to write a function for it.

kadogo:
It would be good if we need the users to tell where the different part are ?

It can work through env variable so steamtinkerlaunch not need to start managing files. If you prefer that it make the operation itself we can still do that. I will update the MR based on what you decide.

frostworx:
Sorry, not sure what you require from me currently _(pretty sure we simply have a two-way language communication problem)_. What do you mean with
`It would be good if we need the users to tell where the different part are`?

_(Currently I write a generic function for #126 as it seems to be a fundamental glitch on debian based distros (so #116 was just a minor part/fix of the whole broken path))_
Then depending on how you want to handle mangohud architecture I'll next either rewrite the current architecture check or keep on with #125 Depending on that decision the Pull Request needs to be changed obviously.

The current Pull Request also is not mergable because
- a preCheckPathFunction which replaced the hardcoded paths [here](https://github.com/frostworx/steamtinkerlaunch/pull/121/commits/6e41817bc5d298bcf8004b6f59c51d067f65a345) like described [here](https://github.com/frostworx/steamtinkerlaunch/pull/121#issuecomment-719911659) is missing
- translation lines[ are missing](https://github.com/frostworx/steamtinkerlaunch/pull/121#issuecomment-719902859)

The whole merge should happen at once to keep dis/appearing variables and settings as low as possible.
I'm sure you agree with that.

kadogo:
> Sorry, not sure what you require from me currently _(pretty sure we simply have a two-way language communication problem)_. What do you mean with
> `It would be good if we need the users to tell where the different part are`?

My question is about how steamtinkerlaunch must deal with libraries.
Example, I have the mangohud libs in a directory that is not in the system library path. We use steamtinkerlaunch to copy them in the system library path or we use environnement variables to load them if Mangohud is enabled?

In my opinion, the environnement variables look better because we let the user decide what to do with his libraries. We can still have a default library path if not provided.

> The whole merge should happen at once to keep dis/appearing variables and settings as low as possible.
> I'm sure you agree with that.

Yes, completely.

frostworx:
Ah ic, thanks for explanation!
No, in no way steamtinkerlaunch is allowed to install anything systemwide (it would require root priviledges anyway to do so).
What I mean is that the best thing to do for the mangohud libs example would simply be to check for a valid path for `libMangoHud.so` (or `libMangoHud_dlsym.so`) sequentially in `/usr/lib/mangohud/lib/` (and `/usr/lib/mangohud/lib32/`)
and `/usr/lib/ `(and `/usr/lib32/`) until it is found and fill the MANGOHUD_LIB and MANGOHUD_LIB32 variable with the found file
 (the same would be valid for the json files in /usr/share/vulkan/implicit_layer.d/MangoHud.json and /usr/share/vulkan/implicit_layer.d/MangoHud.x86.json ).
If it was not found at all I'd "hardcode"  `/usr/lib/` and log/notify a warning.

kadogo:
> Ah ic, thanks for explanation!
No problem, I know that I go sometimes to fast and my english don't help ^^

It sound good for me and if the user want another path he can just edit global.conf or something like that?

frostworx:
Hey! Yes, exactly!
With your current pull request the variables are defined in _default_template_
If there's no reason to make this configurable per game (the library path should be globally the same usually, right? Or are custom libraries used sometimes?) it would be better to write them into the global conf instead.
Do you need help with adding the missing functions? Wouldn't be a problem, I'm very happy that you're helping, but I'd need some time to fix my other open stuff :)

kadogo:
Take your time I will see when I have time to update the PR it must not be very hard ^^ Thanks.

I think global would be good. I not see any case where we must change it per game and if it happen, it will be for another issue or PR ^^

frostworx:
Hehe, sounds like a good plan! Thanks again! :)

kadogo:
I added the lines for the langages.

How do you see the searching function for the libs? I'm not sure it would be usefull for systems path but I'm not sure about it.

frostworx:
Thanks for the changes, they have version conflicts though. Would be nice if you'd rebase them based on current master.
Pardon, I don't understand what you mean with your question/concern

kadogo:
> What I mean is that the best thing to do for the mangohud libs example would simply be to check for a valid path for `libMangoHud.so` (or `libMangoHud_dlsym.so`) sequentially in `/usr/lib/mangohud/lib/` (and `/usr/lib/mangohud/lib32/`)
> and `/usr/lib/ `(and `/usr/lib32/`) until it is found and fill the MANGOHUD_LIB and MANGOHUD_LIB32 variable with the found file
> (the same would be valid for the json files in /usr/share/vulkan/implicit_layer.d/MangoHud.json and /usr/share/vulkan/implicit_layer.d/MangoHud.x86.json ).

I mean how you see this function to work, I was thinking to just set the default path because that is decided by the system. Or maybe you have few paths that think would be checked between differents distro?

frostworx:
Ah I see!
Yes exactly - on Arch linux the paths are 
```
/usr/lib32/libMangoHud_dlsym.so
/usr/lib32/libMangoHud.so
```
and
```
/usr/lib/libMangoHud_dlsym.so
/usr/lib/libMangoHud.so
```
A check function would avoid distribution specific problems, which we had earlier :)

kadogo:
Ok I will use Arch and debian path for lib and vulkan trying to make something.

I tried to rebase (I not do that often I hope it's ok)

frostworx:
Thanks for the update! :)
If you need any help just let me know!

frostworx:
Absolutely no hurry of course, but you don't wait for me doing something, do you? :)

kadogo:
No don't worry it's just me that I'm slow ^^

I just need to take (and make) some time to write the code that is needed ^^

frostworx:
Hehe ok, just wanted to be sure. Thank you for the reply :)

kadogo:
I have in mind to try to make goverlay working in an AppImage too so I'm a bit busy trying to make it work.
The change to do on the PR are pretty easy but taking the time to write the code is always hard xD

frostworx:
A positive effect of the pandemic is that I have lots of free time. That's why I even started expanding steamtinkerlaunch to its current size. It is always good to find time for things that make fun :)
Either way stay safe!

kadogo:
Thanks same for you, not big change for me I mainly do remote work.

Because of my desktop that use Debian I prefer to avoid breaking anything (or having compiling hassle ^^) it's the reason I started the AppImage repo.

I'm sure the Dockerfile are not well written or well secured but goal is having something to use without too much issues at end ^^

frostworx:
Thanks! My wife and me are leading a withdrawn life anyway so no big changes for us as well :)

AppImage comes in handy for many things for sure, but working around a sane package management in either way will never feel right for me :)
But maybe I'm just too old for this shit (R. Murtaugh) ;)

kadogo:
A real package would be better but yeah with Debian I not really have much hope because their goal that are stil having long tested software.

For an example vkBasalt needed GCC9 at east and Buster is only on GCC8 ^^

frostworx:
Hehe, true. Debian likes to "test" software until it is as old as a good whiskey - or simply let it age in a barrel for that time :)
But strange to see that debian derivates don't seem to update their packages as well.

kadogo:
I think it's really related to the use of the applications and who that want maintain them (and that's the hardest part ^^)

frostworx:
I know maintaining packages can be an ungrateful job, but a yad package is pretty much self-maintaining.

frostworx:
Just have seen that I have closed this PR while cleaning up messed up commits.
This wasn't intended, but as it stalled anyway, I guess it's fine to just reopen it whenever the time has come.
stay safe!

frostworx:
Just have seen you were recently active on github, so I'm glad you're fine. Farewell!

kadogo:
Sorry it's just that for now I'm not much about playing so it's "staying" there but I not forget about it to be honnest ^^

I hope you are well too

frostworx:
Thanks for the reply, I'm really glad you're fine!
And yes, I'm fine as well, thanks!

Issue #138, opened 2020-11-07
STL exits due to missing language file even after downloading one?


michaelneverwins:
### Problem
I don't know much about the language files required by STL, but I've noticed something unusual: even though the script apparently has the ability to fetch missing language files from this repository, it still exits with an error after doing so. (Then, the next time STL runs, it works with no complaints about language files.)

### Details
I can reproduce the error consistently by deleting my `~/.config/stl` directory, but if you have anything important in there, moving it works too:
```bash
mv ~/.config/stl ~/.config/stl_backup
```
Then, when I try to bring up the settings window,
```bash
stl settings
```
I get the following output:
```
ERROR loadLangFile - language file '/home/michael/.config/stl/downloads/SteamTinkerLaunch/lang/english.txt' missing - downloading ' https://github.com/frostworx/steamtinkerlaunch' and trying again
Sat Nov  7 16:51:27 EST 2020 INFO - gitPullStl - updating '/home/michael/.config/stl/downloads/SteamTinkerLaunch' from 'https://github.com/frostworx/steamtinkerlaunch'
Cloning into '/home/michael/.config/stl/downloads/SteamTinkerLaunch'...
remote: Enumerating objects: 1026, done.
remote: Counting objects: 100% (1026/1026), done.
remote: Compressing objects: 100% (201/201), done.
remote: Total 1800 (delta 916), reused 905 (delta 814), pack-reused 774
Receiving objects: 100% (1800/1800), 859.82 KiB | 6.77 MiB/s, done.
Resolving deltas: 100% (1337/1337), done.
###############################
ERROR - no language file could be loaded! For the initial setup at least one file (default english) is required
ERROR - You can ether copy a valid file to '/home/michael/.config/stl/lang' or '/usr/share/stl/lang' or provide an absolute path via command line using the lang= option
###############################
```
It looks like the STL script, having failed to find a language file, acquired `english.txt` by downloading this repository but then exited with an error anyway. When I run it again,
```bash
stl settings
```
then there is no error about language files, and the settings window opens as expected.

I'm using the latest SteamTinkerLaunch script, with YAD 7.3, on Linux Mint 19.3.

### Workarounds
The only way I know how to avoid this error, if I'm running STL for the first time with no existing `~/.config/stl`, is to download the file to some valid location manually beforehand, e.g.:
```bash
mv ~/.config/stl ~/.config/stl_backup2
[...]

PR #139, opened 2020-11-08
Load default language after git pull

kadogo:
Hello,

Related to https://github.com/frostworx/steamtinkerlaunch/issues/138 I had the same issue so I took a look and I think it just miss that it need to try again to load the language.

Possible issue is maybe that it could have a loop if there is an issue. Pretty hard that the issue can happen but maybe you have a better idea to fix that properly. Maybe with a counter or a way to do only the pull one time, I let you modify the PR if needed.

### DIFF
--- a/stl
+++ b/stl
@@ -625,6 +625,7 @@ function loadLangFile {
 					echo "ERROR" "${FUNCNAME[0]} - language file '$LAFI' missing - downloading ' $PROJECTPAGE' and trying again"
 					gitPullStl
 					${FUNCNAME[0]}
+                                        loadLangFile "$STLDEFLANG"
 				fi
 			fi
 		fi


frostworx:
Thanks, I'll check this soon!
I need coffee first :)

frostworx:
Just pushed https://github.com/frostworx/steamtinkerlaunch/commit/a5c18f49b398bc7b8a46611cfafccd52ed8f4ef7
which hopefully fixed this and #138 
[Probably](https://github.com/frostworx/steamtinkerlaunch/issues/138#issuecomment-723549136) the bug recently introduced

frostworx:
The fix above fixed the initial bug #138
so this pull request is no longer required.
Thanks for trying to fix it though!
*closing here*

PR #206, opened 2021-03-30
A bit of russian translation

Kron4ek:


frostworx:
Nice, thanks! Maybe this motivates @IDeathByte to start translating again as well :)

PR #227, opened 2021-05-01
Polish Translation

Faalagorn:
Everything is translated, but wasn't extensively tested. Nevertheless STL seems to run fine with it loaded.


frostworx:
Wow, thank you very much @Faalagorn! :)

PR #231, opened 2021-05-01
Even more fixes for Polish

Faalagorn:
Even more fixes, hopefully for now it's all.


frostworx:
Thanks, manually added those which are still available in my offline language files.

PR #237, opened 2021-05-17
Update polish.txt

Faalagorn:
Update Polish for 6.8 plus minor fixes to prior strings. Should be fine to merge, though it's untested, so most likely contains errors.


frostworx:
Thanks a lot @Faalagorn!

frostworx:
Wow, that was even pretty huge :)

PR #239, opened 2021-05-20
Fix yad version check if statement

alisonjenkins:
I was having stl report that my version of yad was '0'. Having checked
the code I found that the $YAD variable was just set to 'yad' which
meant that it would be checking for a file called yad in the current
directory. Swapping to command -v which will use the PATH variable to
determine if yad is present.

### DIFF
--- a/stl
+++ b/stl
@@ -9054,7 +9054,7 @@ function checkIntDeps {
 		MINYAD="7.2"
 		YADVER="0"
 
-		if [ -f "$YAD" ]; then
+		if command -v yad &>/dev/null; then
 			YADVER="$("$YAD" --version | tail -n1 | cut -d ' ' -f1)"
 		fi
 


frostworx:
Hello Alan,
nice to hear from you! Thanks for reporting, this is definitely a problem, but the fix is not 100% correct, as the initial idea is to allow to choose from appimage and conty yad versions _(see [here](https://github.com/frostworx/steamtinkerlaunch/wiki/Custom-Yad))_
so YAD should be defined in global.conf.
Maybe the best solution would be to remove the global YAD="yad" completely and when adding an early function which checks if YAD is configured in global.conf. If yes do nothing, if no set the default
YAD="$(which yad)"
What do you think?

frostworx:
Just pushed a possible fix. Please check if this fixes it for you as well.

frostworx:
Oh, wait, don't miss my previous comment! It might be already fixed in current master!
See here:
https://github.com/frostworx/steamtinkerlaunch/commit/031c8324d6f1bcfff2343f9d9bbc2f88c49ad70c

frostworx:
Thanks for your friendly feedback, mate! :)

PR #252, opened 2021-06-05
Change ReShade WINEDLLOVERRIDES to add to variable

### LINKED ISSUE #251: WINEDLLOVERRIDES set by ReShade
Related to #205. If ReShade is enabled as an option in the Game Menu, then `WINEDLLOVERRIDES` is set (as needed for ReShade) which will overwrite the variable set in the config file. For example, if you need a dll override for a particular game (or a mod), but then also enable ReShade, game will not run with both. A workaround is to use the ReShade overrides in the game config along with whatever else you need (don't enable ReShade through the menus). Minor wrinkle is if you are using this for Depth3D VR stuff, will need to manually open `vr-video-player`.

Solution I think is easy, changing https://github.com/frostworx/steamtinkerlaunch/blob/0cdf8bda1688b9c616661732182068a93c6a5596/stl#L5607 to

`export WINEDLLOVERRIDES="$WINEDLLOVERRIDES;dxgi=n,b;d3d9=n,b;d3dcompiler_47=n,b"`

I'll make a PR for that to review at your leisure.

(Returning to #205, it would be nice to have at least an empty box in the Game Menu for specifying `WINEDLLOVERRIDES`)

(I know you are on break, again no rush! Just spent more time playing with ReShade.)

podiki:
Fixes #251. I didn't see anywhere else where this variable is set, so this should do it. Works in my test (Dark Souls, using DSFix which needs a dinput override, along with enabling ReShade in Game Menu).

### DIFF
--- a/stl
+++ b/stl
@@ -5604,7 +5604,7 @@ function useReshade {
 			fi
 
 			writelog "INFO" "${FUNCNAME[0]} - Setting WINEDLLOVERRIDES for ReShade: dxgi=n,b;d3d9=n,b;d3dcompiler_47=n,b"
-			export WINEDLLOVERRIDES="dxgi=n,b;d3d9=n,b;d3dcompiler_47=n,b"
+			export WINEDLLOVERRIDES="$WINEDLLOVERRIDES;dxgi=n,b;d3d9=n,b;d3dcompiler_47=n,b"
 		else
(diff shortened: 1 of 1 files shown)


frostworx:
Makes sense. thx!

PR #285, opened 2021-08-06
Update polish.txt

Faalagorn:
Rough translation of missing strings to Polish (largely untested, as always) + one typo fixed


frostworx:
Thanks for keeping the momentum going, mate! :)

PR #314, opened 2021-10-03
Update polish.txt

Faalagorn:
Rough Polish update, as always untested, but hopefully should work :)


frostworx:
Thank you very much! I won't merge it yet as I have many changes as well offline, which will be committed _soon_.
Saves some work to get both together.

Faalagorn:
Not sure if you want to keep the `stl` lowercase for Polish as well, as discussed in #315, but I see that "Allow edits by maintainers" is set here, so either I can do it in separate commit, or you should be able to edit my PR if you want to :)

frostworx:
Thanks for the heads-up! I'll take care of it.
Would be great _(no must of course!)_ if you would translate the upcoming additional new lines as well before the next release, so we have at least 3 complete language sets _(I even wondered if it even does make much sense to keep the other ones, as adding all new lines to them to keep all in sync is useless extra-work)_

frostworx:
Thanks a lot, just merged it with my major version bump:
https://www.reddit.com/r/SteamTinkerLaunch/comments/q1785a/vortex_mod_manager_mod_organizer_2_specialk_added/

PR #315, opened 2021-10-03
Update english.txt

Faalagorn:
Fixed some issues found while translating to Polish


frostworx:
Thanks! :) Reverting STL to stl though :P

Faalagorn:
> Thanks! :) Reverting STL to stl though :P

Fine, as long it is consistent – was late when I was doing that so didn't do much looking around :eyes:

frostworx:
> Fine, as long it is consistent

Should be - I never write it uppercase, because the program itself is lowercase

Faalagorn:
> > Fine, as long it is consistent
> 
> Should be - I never write it uppercase, because the program itself is lowercase

Well, I see that you merged it now, so I guess you have to change it in separate PR; I realized I didn't have the "Allow edits by maintainers" checked though, so you probably couldn't anyway. :P

EDIT: and yeah, just checked, it was the only instance of `stl` in messages.

frostworx:
Yeah, no problem. I already added your changes _(minus STL)_ in my offline english.txt, but your changes in the polish.txt were too huge to do the same there :)

PR #318, opened 2021-10-09
Adds British English + minor changes to english.txt

jake770321:
Unsure about the change on line 346 as it removes 'XXX' but it currently isn't being replaced with a variable and shows as "The own project page of 'XXX'"


frostworx:
Yay, this is great! Thank you very much for your contribution, and also thanks for the hint with the missing variable, 
will fix this shortly.

frostworx:
Fixed the missing variable :) Thanks a lot! :) https://github.com/frostworx/steamtinkerlaunch/commit/48fbcb477b31bd20e02ce96f49d79b2093edd251

frostworx:
Just renamed englishUK.txt to eng-UK.txt, to avoid removing a false positive duplicate.
Could be solved better, but I'm short of time.
I hope this is ok for you.

frostworx:
Just fixed it properly :)
https://github.com/frostworx/steamtinkerlaunch/commit/0dbfbb668db7cdd45b88318eb25fa8475da31fa1

jake770321:
got 2 quick questions. hope you don't mind answering

1. Should i be copying over new entries from english.txt when/if they get added 

2. What happens if an entry is missing? Does it just revert back to english.txt or will things break?

frostworx:
Thanks for asking!
1. No, I take care for all language files always having all new entries available in english at the same position (usually everything new will be appended). This makes it easier for translators to find untranslated lines.
2. No, it would not revert back to english.txt, so things _could_ break. Most lines in the language files are "informational" though, so I guess it wouldn't be fatal in the most cases when a line is missing.
Feel free to ask if anything is still unclear.

PR #327, opened 2021-10-21
Update polish.txt

Faalagorn:
As always, largely untested. Fixed some issues with existing translation too


frostworx:
As always, thanks a lot @Faalagorn! :)

PR #334, opened 2021-10-27
TRANSLATION translated to Dutch - line 301

JSvanVeen:
Saw I made a mistake with some extra quotes in one line, so removed them and checked for other mistakes with some regex. Other than that translated up to line 301 so it wouldn't be just a 1 change PR.


frostworx:
Thank you very much, @JSvanVeen! 👍
Yesterday I added two more lines and "translated" all languages myself to avoid another comment in the translation issue.
If the Dutch one is too bad, please send a PR or just tell me what it would be correctly.
For
`GUI_CUPROTOCOMPAT="Extract custom Proton in compatibilitytools.d"`
I added the dutch line
`GUI_CUPROTOCOMPAT="Uitpakken aangepaste Proton archieven in compatibilitytools.d"`
_(probably totally wrong grammar - sorry...)_

PR #343, opened 2021-11-08
Add `stl vortex url <url>` command

MattSturgeon:
This PR adds the `url` command present for `mo2` to `vortex`. The MIME Desktop Entry now makes use of this feature, rather than hard-coding wine variables and exe paths.

I also made it so the MIME Desktop Entries use the absolute path to stl, rather than `$0`. This is because during my testing I was running `./stl vortex start` which meant the generated .desktop file contained `Exec=./stl vortex u %u` rather than `Exec=/home/matt/Dev/steamtinkerlaunch/stl vortex u %u`. In normal usage this'll be `Exec=/usr/bin/stl vortex u %u`, which should be more robust than relying on `PATH`.

I'm not very familiar with your codebase, so please let me know if there's anything you'd like done differently or anything I've neglected to do.


frostworx:
Oh nice, thank you for the contribution! I haven't tested it _(yet)_, but your changes look all good! 👍

Issue #200, opened 2021-03-22
cheatengine


frostworx:
apparently the ce download was removed from the github page https://github.com/cheat-engine/cheat-engine/releases.
feel free to PR sane patches with a sane dl url. [this](https://github.com/frostworx/steamtinkerlaunch/issues/160#issuecomment-736897776) might help.
#199 #198

frostworx:
the md5 of my old download is 
`8ce5de1b284bc9f865865902fea09489  CheatEngine72.exe`

frostworx:
https://github.com/cheat-engine/cheat-engine/issues/1523

frostworx:
After reading [this](https://www.cheatengine.org/downloads.php) I doubt that a "a clean install file" will become available again on github.

esotericDisciple:
Uh, no. The game menu says 7.1 is the cheatengine version. I haven't used cheatengine in four months, so I've might have forgotten how to use steamtinkerlaunch entirely, but the docs on getting cheatengine to work weren't helpful at all; there's no screenshots for the process at all...

neurodiverseEsoteric:
So basically this issue can't be fixed and therefore steamtinkerlaunch is not longer usable?

frostworx:
Looks like there's no other valid download url for the removed "clean install file", and those who want to use CE via steamtinkerlaunch either have it downloaded before or are unwilling to help _(and/or unable to read)_.
Of course I could simply host the last "clean install file" myself without any legal issues and use that as source,
but there is apparently no reason to do so currently.
There's still a chance that upstream reacts on my issue of course.

frostworx:
Someone reminded me in the upstream issue that there's still the rar file to download _(already linked [above](https://github.com/frostworx/steamtinkerlaunch/issues/200#issue-837364978) ='this' )_.
If someone is intererested in creating a PR for the download, extract and unrar dependency check _(instead of innoextract)_
a (currently) valid wget command would be
`wget -O CheatEngine72.rar --user-agent="Mozilla" --referer=https://www.cheatengine.org/download.php 'https://www.cheatengine.org/download/CheatEngine7.2_MissingSetup.rar'`
Should be very easy to change, so this might be your chance to actually contribute instead of only expecting a "fix".

neurodiverseEsoteric:
I've never used stl language/code before...

frostworx:
It is pure bash.
[...]

PR #360, opened 2021-11-26
CheatEngine: Fallback to WayBack Machine Archive for 7.2

SaifRushdHadad:
A quick hotfix to support installation of ad-less cheat engine 7.2 installer mirror.

We utilize the Internet Archive's Wayback Machine (https://archive.org/web/) mirror (https://web.archive.org/web/20210118065254if_/https://github.com/cheat-engine/cheat-engine/releases/download/7.2/CheatEngine72.exe) to ensure that we do not have to rely on community members to repackage or redistribute the 7.2 installer. 

The only notable potential downside are:
1) end-users being subject to being rate limited due to large amounts of download requests coming from their IP.
2) dependency on web archive being available and up
3) ability to access and resolve web archive domain name

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -10,6 +10,7 @@ PROGVERS="v8.0"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
+CEWAYBACKURL="https://web.archive.org/web/20210118065254if_"
 PROJECTPAGE="$GHURL/frostworx/${PROGNAME,,}"
 PPW="$PROJECTPAGE/wiki"
 CURWIKI="$PPW"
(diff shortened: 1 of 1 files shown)


SaifRushdHadad:
Resolves: #200

frostworx:
This is indeed a good idea! I use wayback pretty often, but never looked into it as CE solution.
Thank you very much for your contribution!
Merging it untested, but I'm pretty confident that it works :)

frostworx:
Hello @SaifRushdHadad!
I just notices, that for whatever reason your pull request no longer is in the current master.
I have no idea how this happened, and it wasn't my intention to revert the commit.
I'll re-add it shortly.
Sorry for that! :(

frostworx:
*fixed* : https://github.com/frostworx/steamtinkerlaunch/commit/c945f200224e740e418abb899cb74b656f70d977
Sorry again :(

PR #362, opened 2021-11-28
Add protonlist.txt to migrateCfgs function

ghost:
The _migrateCfgs_ function is missing protonlist.txt and migration results in an empty file. All existing contents are lost because the _getAvailableProtonVersions_ function removes invalid and/or non-existent paths and the parent directory has been renamed.


frostworx:
thx!

PR #397, opened 2022-01-09
Update English

Faalagorn:
Update English and UK English based on the fixed I found when translating to Polish.


jake770321:
Shouldn't "SHA56" be "SHA256"?

frostworx:
Great, thanks for both the translation @Faalagorn and heads up @jake770321

PR #404, opened 2022-01-24
Add app name to default notifier

### LINKED ISSUE #403: Add app-name for KDE Plasma notifications
SteamTinkerLaunch notifications on KDE Plasma display the process as "notify-send". It would be nicer if it said "SteamTinkerLaunch" as it's more in line with what a user would expect. Below is a small example of a notification I was able to trigger on my system, though for what it's worth, most applications do not display this name correctly on KDE Plasma and the only one I can actually think of that properly displays its process name is Firefox (as just "Firefox" or "Firefox Developer Edition", etc).

Screenshot_20220124_211008

I took a look into how this might be implemented and I'm not sure exactly how this is done. It seems like this information can be taken in from a "desktop" file. An example I was testing was `notify-send foo bar -h STRING:desktop-entry:steamtinkerlaunch`, which still displays "notify-send" as the process name, but will display the steamtinkerlaunch icon, which does not normally happen with SteamTinkerLaunch notifications at all either (and may not be desired behaviour, either).

It is very much just a "nice to have" feature but would add extra clarity as to where notifications are coming from for users.

sonic2kk:
With `notify-send`, the default notifier, the `-a` flag can be used to set the name of the application. This is set to the existing `$PROGNAME`. It is also possible to use an `--app-name=` syntax, but going with `-a` keeps in line with the existing `-i` syntax for adding the icon to the notification.

This flag is not mentioned in the manual for `notify-send` on my Arch Linux installations, but it is mention on [this](http://vaskovsky.net/notify-send/linux.html) page and I can verify that the flag works. Discussion and testing on other distributions may be required to make sure this doesn't raise errors or unwanted behaviours.

Example of a game launch notification with this change on KDE Plasma:
Notification with -a flag on KDE Plasma

Alternatively, if this change is undesirable, steps for adding this flag manually should be included in the [Notifier wiki page](https://github.com/frostworx/steamtinkerlaunch/wiki/Notifier).

Closes #403.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -2226,7 +2226,7 @@ function setDefaultCfgValues {
 		if [ -z "$MAXASK" ]								; then	MAXASK="3"; fi
 		if [ -z "$BROWSER" ]							; then	BROWSER="$(command -v "firefox")"; fi
 		if [ -z "$NOTY" ]								; then	NOTY="$(command -v "notify-send")"; fi
-		if [ -z "$NOTYARGS" ]							; then	NOTYARGS="-i $STLICON"; fi
+		if [ -z "$NOTYARGS" ]							; then	NOTYARGS="-i $STLICON -a $PROGNAME"; fi
 		if [ -z "$USENOTIFIER" ]						; then	USENOTIFIER="1"; fi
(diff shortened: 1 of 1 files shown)


frostworx:
Thanks for the fix!
For whatever reason the -a parameter doesn't have any effect on my Arch xfce4, but that is no reason to not merge it. :)

PR #409, opened 2022-02-04
french translation update

tl-pierre:


frostworx:
thanks a lot @tl-pierre 👍

PR #433, opened 2022-03-29
Set WINE_FULLSCREEN_FSR_STRENGTH default value to 2

sonic2kk:
I may not have implemented this correctly but I wanted to make an attempt at this, get feedback if it is not implemented correctly and also get a discussion going :smile: 

The default sharpening value for FSR right now is 5, however [the recommended default](https://github.com/ValveSoftware/wine/pull/116) for this patch is 2. Gamescope [also uses](https://github.com/Plagman/gamescope/pull/360/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R146) 2.

This PR aims to set the default value to 2 based on this. Feedback and guidance welcome if I did not change the correct value, I couldn't get this to display but think that might be a different problem.


frostworx:
Good catch! Thanks for the PR! :)

PR #434, opened 2022-03-29
Add FSR option to Gamescope

sonic2kk:
Gamescope now allows users to use FidelityFX SuperResolution (FSR) 1.0 by adding the `-U` flag. This PR adds the ability to enable Gamescope's FSR in the Gamescope menu. This can be toggled with the Super+U shortcut, so for any PR testing you can use this to see the difference :smile: I think this could be a useful way for more people to access FSR that want to use Valve's official Proton build. I also believes the Steam Deck uses Gamescope FSR.

[](https://postimg.cc/Q9vH5Xrk)

[The default sharpening for Gamescope FSR is 2](https://github.com/Plagman/gamescope/pull/360/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R146), judging by the code, though this is not officially documented anywhere. You can adjust Gamescope's FSR sharpening with the keyboard shortcuts, but I'm not sure if there's a way to pass a value. If there is, then I don't think it is documented. Using a command line `gamescope -U 5 -- <game>` doesn't work.

Games tested, though it should not be game specific:
- Cookie Clicker
- NieR:Automata
- Yakuza 0

I have a few concerns though. I am unsure what the expected behaviour is if Integer Scaling and FSR are used together. I suspect whichever comes first would take priority, and based on my understanding of how STL works, this new FSR flag would take priority. Is there a way and should one of these filters be disabled if the other is in use?

My other concern is how this works with builds of Wine which have the Wine FSR patch in the fshack patch (wine-tkg allows this I think, and Proton-GE did and may still have this patch too). This may need further research.

The German translation line included is courtesy of a friend. Do translation lines need to be included for all the language files even if they are not translated? If so, I can add these no problem. Feedback on the phrasing of the menu items is also appreciated, if more information should be included (e.g., that it is FSR 1.0 as opposed to [FSR 2.0](https://gpuopen.com/fidelityfx-superresolution-2/), which is not supported by Gamescope or Wine yet.

The implementation of the FSR checkbox borrows heavily from the code of other Gamescope menu items from STL, and some of my own investigating :sweat_smile: I'll be happy to make any changes to improve the code quality. All feedback welcome :)


frostworx:
Wow, thanks a lot! Please give me a bit tolook deeper into this.
Would you mind me adding a prt of your description to the wiki when merged?
yes, please bump every translation file with the english variables, to make sure everything is sorted correctly and people who want to translate something have every line ready.

sonic2kk:
I'd be happy to add any information to the wiki! Take all the review time necessary. Thank you for your work on SteamTinkerLaunch. The translation files should be bumped now :partying_face:

frostworx:
This looks very good! Thanks a lot! :)
Just noticed that the current steamtinkerlauch [gamescope wiki](https://github.com/frostworx/steamtinkerlaunch/wiki/GameScope) is pretty much empty anyway, so I'm not really sure if it makes sense to describe your new function :)
_(of course you're very welcome to improve any wiki you want :D)_
Would you mind me using your above fresh screenshot for that wiki entry?
*merging your PR*

sonic2kk:
I don't mind at all, go ahead 😄 Thanks!

I'll take a look at the wiki and see if there's anything I can contribute for Gamescope that might help users 👍

frostworx:
thx! done :)

PR #438, opened 2022-03-30
Add option for Nearest Neighbor filtering

sonic2kk:
Gamescope offers nearest neighbor filtering with `-n`. This PR adds the option to enable nearest neighbor filtering in the Gamescope menu. The difference is minor between the two, and FSR is probably more desirable in most situations I imagine.

I feel Integer Scaling and Nearest Neighbor are loosely related, so I put them together in the UI, code (hence the re-ordering of `GSARR`) and the language files. I didn't do this with #434 as I felt FSR is a different "kind" of technology that's kept separate. I can move it if desired :)

[](https://postimg.cc/Tyq4B80g)

I investigated to find any technical differences between nearest neighbor and integer scaling, I wanted to make sure they were entirely different before continuing. I didn't find much, so I decided to dive into some testing and comparisons to see for myself :smile: 

### Gamescope Arguments Priority and Behaviour
I tested by launching Gamescope outside of SteamTinkerLaunch, with `-i` and then `-n` to check for differences. And indeed there was! If you pass `-i`, `-n`, and `-U` (for Integer Scaling, Nearest Neighbor and FSR respectively) it seems the option passed last takes priority. So if running `gamescope -i -U -n -- supertuxkart`, Nearest Neighbor (`-n`) would be used. Related, if you pass `-n` and then press `Super+U` to enable FSR, and then again to disable FSR it goes back to Nearest Neighbor. The same applies when going from Integer Scaling (e.g., passing `-i`) to Nearest Neighbor (`Super+N`); it will go back to using Integer Scaling once Nearest Neighbor has been disabled.

With how `GSARR` appends Gamescope arguments, **the priority for Gamescope arguments will be Integer Scaling, Nearest Neighbor and FSR**. Following from discussion in #434, I plan to add this information to the wiki, but will wait for decision on this PR in case information from here needs included. 

### SuperTuxKart (Native Linux) - No SteamTinkerLaunch
Here is a comparison of SuperTuxKart, running with Gamescope's default internal resolution of 720p and fullscreened on a 4k display. I launched using `gamescope -n -- supertuxkart` and `gamescope -i -- supertuxkart`. The image is PNG to preserve as much quality as possible for detailed comparison. The difference, I think, is most visible on text.

[](https://postimg.cc/bZt3bX2k)

### DuckTales Remastered (GE-Proton7-9)  - SteamTinkerLaunch
Here is another comparison, this time running the Steam game DuckTales Remastered using this PR's Nearest Neighbor checkbox. I ran with and without STL to show how it should look normally, and then how it looks with this PR to demonstrate it in action :smile: Zooming in on the "DuckTales" title image shows the difference, though to me I did not really see much of a difference.

[](https://postimg.cc/8jyqwV5Y)

Scaling up from 720p to 4k is quite the extreme example but my hope is that it demonstrates the difference between the two, and the justification for making this PR. The `-n` flag was recently updated to use `-i` for Integer Scaling as per a Gamescope update (#435). This PR will bring back the option to use the `-n` flag for those that were using the filtering method.

Lots of background here, hopefully it justifies the logic for the feature :smile: As always, feedback is welcome! Still learning the ropes with shell scripts.


frostworx:
Wow, impressive description! 👍 _(you're always very welcome to rewrite the whole wiki 😀)_
Thanks a lot, again for your contributions!

Found a typo in your translation changes: GUI_GSNN="Nea**r**est Neighbour"
would be great if you'd fix it, before I merge it.

sonic2kk:
Ah! Should be fixed now, sorry for the embarrassing typos :sweat_smile: 

I'm not on my PC at the moment but I can update the wiki with a corrected screenshot and information over the weekend (fingers crossed :crossed_fingers:)

EDIT: I forgot one of the language files, force-pushed an update. *Now* it should be fixed :smile:

frostworx:
wow, that was fast! thanks a lot! :)
No hurry with the wiki, but of course it would be great if you really wanted to update it :)

Issue #440, opened 2022-04-01
Custom Gamescope arguments get overwritten


sonic2kk:
## Problem
If you pass a custom argument to Gamescope from the Game Menu and then edit some options in the Gamescope menu, the custom Gamescope arguments get overwritten by only the flags selected in the Gamescope UI. Normally, this wouldn't really be a problem, as those that edit Gamescope options probably aren't going to use the Gamescope menu anyway, so it would be a relatively minor issue.

## Real world use-case
What makes this slightly less of a minor issue is a bug with Gamescope and Proton 7.0 (Plagman/gamescope#438, Plagman/gamescope#420) that can be resolved by passing the `-e` flag. This flag is for "Steam integration," though there is not meant to be a reason for a user to use this flag as far as I know (I saw it from someone on the Gamescope project, but I can't find the comment now). The bug also affects GE releases of Proton.

The official description of the flag in `gamescope --help` is as follows:
```
-e, --steam                    enable Steam integration
```
It is probably not really the responsibility of a project like SteamTinkerLaunch to manage working around bugs in upstream like this. Trying to solve this specific issue has a few challenges too:
- Passing `-e` breaks non-Steam games
- Passing `-e` silently may work fine for Steam games for now, but its behaviour may change in the future, so passing it even silently may be a problem
  - If a user removed it too and their input stopped, they may be confused as to why it stopped
- Having an option for `-e` in the Gamescope menu may be confusing, because the behaviour may change in the future while for now it is used to workaround a bug

## Potential Solutions
With this use-case out of the way, my main reason for creating this issue is to at least document that SteamTinkerLaunch currently overwrites any custom Gamescope arguments once the Gamescope UI is used and "Done" is clicked (triggering an update of the Gamescope arguments).

Fixing this may be challenging, but I see a few possibilities:
- Implement a setting for every Gamescope option, hiding more advanced settings under a dropdown (is this even possible with yad?). This would require some work and may be unreasonable, most of the Gamescope flags are probably not useful to the average player.
- Display a Gamescope arguments textbox on the Gamescope menu. This may be problematic as whenever an option is selected (such as enabling fullscreen), the box would need to update. To my understanding, with the way the Gamescope menu works currently, a display value would need to be used for this textbox and then that would be written out as the actual Gamescope argument value. I'm not even sure if this kind of behaviour is feasibly possible with yad, and is giving me a headache just thinking about it :exploding_head: 
[...]

PR #441, opened 2022-04-02
Add option for Steam integration with -e flag, for working around a Gamescope bug

sonic2kk:
Following on from discussion in #440, this PR adds the ability to toggle the `-e` flag for Steam integration. This serves to work around a bug in Gamescope with Steam games that prevents mouse and controller input with Proton 7.0 (https://github.com/Plagman/gamescope/issues/438, https://github.com/Plagman/gamescope/issues/420). I am not sure how long this bug will be present for or when the `-e` flag might start to have some other behaviour, but for now it is mostly harmless.

[](https://postimg.cc/VSc7Tbz5)

I tested using a few versions of Proton I had on my system:
- With Steam games using Gamescope+Proton 7.0, this fixes a bug preventing mouse and controller input
- With Steam games using Gamescope+GE-Proton7-10, this fixes the bug too
- With Steam games using Gamescope+Proton 7.1-GE-2 (an older release of GE than above), the mentioned bug is **not** present but the flag does not seem to have any impact either way, so it is harmless to keep in place here as far as I can tell.
- With Native Linux games on Steam using Gamescope, the `-e` flag makes no difference.

Proton games tested:
- Cookie Clicker
- Tales of Berseria

Native Linux Steam games tested:
- Portal
- Doki Doki Literature Club

A note from my general testing: Launching a game with `gamescope -e -- supertuxkart` from Terminal causes Gamescope to hang, and the game does not launch. So games using `-e` that are non-Steam games might fail. Trying to test this seems to be tricky, as messing around with adding native non-Steam games to Steam and launching them with `steamtinkerlaunch %command%` likes to make Steam crash. I have not tried non-Steam Proton/Wine games. **However, as this feature is meant primarily to fix Steam games with the input bug, I did not consider these cases of major importance**. To compensate for this though, a note has been added in the tooltip that this may break non-Steam games.

I thought about setting this to be enabled by default, but was unsure if this was a good idea. On one hand, it will probably (for now) affect a good amount of users. But on the other hand, this behaviour is a bug and one that is subject to be fixed anytime. Similarly, the behaviour of the `-e` flag may change with a Gamescope update to have undesirable effects. Also, it does not affect all versions of Proton, so it may just create clutter having it enabled by default. Also, admittedly, I had difficulty finding a way to enable it by default, and then came to the conclusion that it may not be the right behaviour to begin with. I struggled to find a way to enable it by default but allow the user to toggle it from the menu and also remove it manually from the Game Menu. I figured as well, that a user typically won't enable Gamescope without configuring it at all. It can be toggled from the Game Menu but setting the resolution at the very least is probably what most users will do instead of running a game at the base 720p window. So they should hopefully be exposed to the option and can know to enable it.

Either way, I'll also document more information about this flag on the wiki for further visibility :slightly_smiling_face: 

Feedback on whether this should be enabled by default (and maybe some guidance on how to do it :sweat_smile:), the GUI text, and any other areas are always welcome!


frostworx:
Wow, you're fast! Thanks a lot, again! 
I think it is perfect like it is now and should not be enabled by default,
because this would override current upstream configurations.

With your solution, everybody has the opportunity to enable it self-responsible per game on demand.

I doubt that there are that many users, who would like to enable it for all games anyway,
but in theory, the best solution which comes into my mind for that would be to add another command line parameter to
`steamtinkerlaunch gamescope` _(f.e. just `steamtinkerlaunch gamescope default`)_,
which would write the configuration options into `STLDEFGAMECFG` _(the default template)_  instead of the game-specific `STLGAMECFG`. For all game configs already existing, the `GAMESCOPE_ARGS` can be overridden already using `steamtinkerlaunch set GAMESCOPE_ARGS [args]`
_(I barely use that option, but I assume it still works 😀)_

_PS: can't wait to see your wiki, although I'm already a bit envious, because it will be the best part of the whole wiki,
and it took me a lot of time to write all that mess 🤓_

PR #452, opened 2022-04-15
Enable support for winesync

r-vd-woude:
Just a quick pull request, I don't expect you will merge, but I thought it might be an interesting idea.

This will allow environment variables to be exported that can enable winesync to run. Winesync is an alternative for fsync, the latter can crash some games.

Winesync has a few downsides. It needs a seperate kernel module called "winesync" and a Wine (or Proton) build not based on staging and with esync and fsync disabled.

Anyway, interested in how you think about it.


frostworx:
Thanks for the contribution! I see no reason why this should not be merged.
_(maybe it would be a good idea to also check if the winesync kernel module is loaded and log a warning if it isn't)_

Issue #450, opened 2022-04-12
Question: How to set arbitary env vars through GUI


IcyAlmond:
I've found https://github.com/frostworx/steamtinkerlaunch/wiki/ENV-Variables but and it seems to work if i set it through the configuration file. But is there a way to add custom environmental variables through the GUI? I wasn't able to find it if there is

frostworx:
No, adding them manually is the only option.
Imho not worth the time to add a gui option (of course I'd accept PRs anyway)

frostworx:
just merged the new feature and also updated the [ENV Variables](https://github.com/frostworx/steamtinkerlaunch/wiki/ENV-Variables) wiki.
global and per-game configs are autocreated and ignored if empty.

please test and report back @Latrolage 

edit: thanks here as well for the [initial commit](https://github.com/frostworx/steamtinkerlaunch/pull/453), @sonic2kk ;)

IcyAlmond:
The new configuration file works for env variables without issue.

PR #453, opened 2022-04-16
Add gamecfg file to game files menu

sonic2kk:
This PR adds an option in the "Game Files" menu to open and edit the game configuration (`$STLGAMEDIRID/$AID.conf`) and global configuration file (`$STLCFGDIR/global.conf`), with the primary use case in mind being easy access to setting/adjusting environment variables (See #450). 

Screenshot_20220416_220130

The path to the game-specific config file uses the game's AppID, so the path shows up as `/home/gaben/.config/steamtinkerlaunch/gamecfgs/id/60.conf`. This *hopefully* should not cause confusion for users, as the AppID is displayed in various places in the SteamTinkerLaunch UI, it's displayed on that very Game Files menu beside the game name, and other paths display the AppID (even though those other paths don't have any other names, so there is no real choice of what name to display there). The name of the file in a text editor may cause confusion, but the top of the configuration file lists the game's name. It may also be somewhat cleaner to link to the "real" file and not to a link file. But regardless, if it is preferred to link to the file with the game's name, that is no problem.

# Rationale
The rational for putting this setting in the Game Files menu was for a few reasons:
- The functionality of selecting and opening game-related files was already present in the logic for this menu, meaning no additional code was needed.
- Users can open and edit the config files with whatever text editor they like, but can do it from the SteamTinkerLaunch UI. They don't have to do it before launching a game.
- Users don't have to know or remember where a game's config file is
- Having what is essentially a shortcut to open the config file from this menu lowers the maintenance cost of this feature

# Reservations
I have a few reservations about this PR, however. Having access to the global config file in this menu might not be correct. I had a browse through the UI and couldn't think of where to put it. Initially, I thought to have buttons in the Game Menu and Global Menu to do exactly what this PR does, but decided to use the Game Files menu instead. Perhaps it isn't even desirable to have such direct access to the global config like this? Is it even an intuitive place for the global config file? Further discussion may be needed.

Another problem I foresee is this. Using the Game Files menu, as mentioned, keeps maintenance cost down, but this comes at a price. Users can easily edit the config file but they can't see what edits they have made *inside the SteamTinkerLaunch UI*. If they set an option, they may forget what they set and what was there by default. To improve this, perhaps the last few lines of the config file could be changed to look like the following (newlines included intentionally as a guide for formatting, but are optional :slightly_smiling_face:):

```

## Custom environment variables can be set here (without export).

```

This would invite users to keep their custom environment variables together, at the bottom of the file, and help them remember where to find them. I didn't add functionality like this in this PR as I'm not sure how to do it quite honestly, plus it may not even be desired and if it was added in a PR, is probably best to be reviewed separately from the UI change.

A menu option that may go hand-in-hand with this proposed change would be a checkbox to disable exporting of custom environment variables, in the Game Menu and Global Menu. That would be tricky to implement though, in a scenario where there may be global environment variables but the user wants them disabled for a specific game (potentially for debugging purposes, for example).

<hr>

After some thought, this was the most straightforward solution I could come up with to add this feature. It keeps the option out of the way for users that don't need it, while being in a fairly intuitive place for users that want it. Hopefully the approach is acceptable :smile: I'm also happy to update the relevant wiki pages if this PR is merged. All feedback is welcome!


frostworx:
Thanks for the PR! Basically a good idea indeed to simply offer the main configs in the already available editor menu.
On the other hand your concerns regarding manual edits vs. representating state of the gui entries are also valid.
What do you think if we simply add a completely new config file, which is exclusively meant for manual env-variables? It could be simply sourced shortly before the game launch _(no need to source it earlier, as it doesn't appear in the regular menus)_ and can be offered in the editor menu as well easily. This could be also split into a per-game custom-vars config and an additional global custom-vars.conf. Imho this would make it easier for the user to find their custom variables, and avoid possible complications with auto-recreating the main config files.

sonic2kk:
Definitely, I think the idea of a separate env var file on a per-game basis and then a separate global env var file is a much, much cleaner solution. I hadn't even thought about your last point before, re-creating config files and then losing custom variables could be a real pain, especially if there's a scenario where several might be needed (or wanted) for a game!

I'm happy enough to close this PR then in favor of a solution closer to what you described, it would do a better job of solving more of the problem described in the mentioned issue too (clearer for the user like you mentioned and potentially better UI integration).

frostworx:
If you want you could also leave the PR open and I'll merge it, when I find the time to add the new custom configs. Most of your PR could still be used for it.

sonic2kk:
Sounds good then, I suppose the paths will just need changed to point to the new configs :)

frostworx:
yeah, this should be enough indeed :)
I think I will add it soon.

frostworx:
why wait - I'll add it now :) *merging*

frostworx:
just merged the new feature and also updated the [ENV Variables](https://github.com/frostworx/steamtinkerlaunch/wiki/ENV-Variables) wiki.
global and per-game configs are autocreated and ignored if empty.

PR #466, opened 2022-05-17
Further improve the Chinese translation

ling0412:


frostworx:
Wow, you're fast! Thank you very much! :)

PR #471, opened 2022-05-25
Add Gamescope option for NVIDIA Image Scaling v1.0.2

sonic2kk:
While I wait (im)patiently for my Steam Deck order invitation, I figured I'd make an attempt at implementing an option for a newly added Gamescope feature! :smile: 

# Overview
This PR adds a checkbox to enable Gamescope's new NVIDIA Image Scaling v1.0.2 (NIS) feature, available in the latest Gamescope git. This is an upscaling technology developed by Nvidia that serves essentially the same purpose to AMD's FidelityFX SuperResolution (FSR): Making games at lower internal resolutions look better at higher display resolutions. It works on AMD, Intel and Nvidia GPUs. It can be enabled by passing the `-Y` flag or the more verbose `--nis-upscaling` option. This PR uses the shorthand `-Y` flag, keeping with the pattern of other launch options.

Screenshot_20220525_065412

# Comparison
## With/Without NIS
Here is an example of how NIS looks with a sharpness of 2 ([the default](https://github.com/Plagman/gamescope/commit/a515153a93f877d5e306a59568ba701120ac95c8#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R152) as the upscaling values are shared between NIS and FSR now, more details below), scaling up from an internal resolution of 720p to a display resolution of 4k. Note the text on each side of the image to see the difference:

NIS comparison sonic mania stl

<details>
<summary>Perhaps a clearer example, though it may hurt your eyes. Left is without NIS, right is with NIS enabled.</summary>
STL Gamescope NIS example
</details>

<hr>

## NIS Sharpness Comparison
This PR also renames the "FSR Sharpness" spinner to say "Upscaler Sharpness", as this setting also applies to NIS and is [internally referred to as Upscaler Sharpness](https://github.com/Plagman/gamescope/commit/a515153a93f877d5e306a59568ba701120ac95c8#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R151-R152). Here is a comparison of NIS with a sharpness value of 2 (the default) and a sharpness value of 20, the minimum upscaler sharpness [allowed by Gamescope](https://github.com/Plagman/gamescope/commit/a515153a93f877d5e306a59568ba701120ac95c8#diff-b0f748a3fc316ec1a44fb378d68c31a12acecc1d6f7b57a05a82e1137d6dff4cR167).

NIS comparison 20 2

The `gamescope --help` screen also says the values for the upscaler range from 0 to 20 and the Gamescope code also seems to reflect this as well, though I am not an expert on Gamescope or C++. Using the Meta+I and Meta+O keyboard shortcuts to increase/decrease sharpness also works for NIS.

# UI Changes
As a related change, the `--fsr-sharpness` option, though it still works when applied to NIS, has been renamed to use the new `--sharpness` flag, as this flag makes it clearer to users that this flag works for NIS as well when they look at the Gamescope arguments in the Game Menu. Previously, this flag was checked for and added when FSR was enabled, but this is now in a separate if block broken out from the FSR check. Instead, the sharpness value is passed only when FSR *or* NIS *or* both are enabled.

# Edge Case Considerations
I believe even before this change there is a behaviour where if FSR is disabled, the sharpness value is reset to 2. But this IMO should be expected behaviour, as the sharpness option should not be passed to Gamescope if no upscaler is selected. This means the user will lose their selected sharpness value but this value is either defaulted to 2 or read from the `--sharpness` value passed to Gamescope. If neither FSR nor NIS is selected, no sharpness value is passed to Gamescope, and so the next time the Gamescope menu is opened, the spinner will default to a value of 2. This is theoretically fixable, but probably a lot more effort than it's worth for such a small problem. Though it could be annoying if a user has fine-tuned a specific sharpness value for a specific game, I think that's more of an edge case than it's worth to pursue fixing.

In the same vein of edge cases, there is a possibility that a user may want separate sharpness values for FSR and NIS. For example, a sharpness value of 2 might be fine with FSR, but far too much with NIS (this is my own personal opinion for some games I tested tonight with this PR). However this means SteamTinkerLaunch would need to track two separate sharpness variables: One for FSR and one for NIS, and this would be entirely separate from the sharpness value stored and sent to Gamescope's `--sharpness` flag. and then apply the appropriate one based on whichever upscaler is selected. I did not pursue this avenue for a few reasons:
- This is adding logic and checks on top of Gamescope separate from what Gamescope itself offers, which may be beneficial for the IMO small selection of users that would want this feature, but may be confusing and could be more costly to maintain. I thought it best to stick with what is offered by Gamescope and simply offer a frontend for configuring Gamescope, rather than trying to essentially create a new feature
- A user would probably typically not switch between using FSR and NIS with the same game
- Similarly, FSR and NIS conflict with each other, so a user would not switch between them frequently
- Re-using the existing sharpness spinner was less effort :stuck_out_tongue: 

The internal names for the sharpness value (`GSFSRS`, Gamescope FSR Sharpness) was not changed, as this could be a breaking change and is not strictly necessary. It should still be clear based on the code what purpose this variable services.

# Potential NIS Quirks
NIS has a strange behaviour that is not exhibited by other Gamescope filters. For example, if you enable Nearest Neighbor and then enable FSR, Nearest Neighbor will get disabled but when you disable FSR, Nearest Neighbor will resume being used. However, FSR will not be applied on top of the Nearest Neighbor effect as far as I can tell, or Integer Scaling for that matter. However, NIS *will* be applied on top of Nearest Neighbor (and vice versa), but it *conflicts with FSR*. Whichever flag is passed last is the one that will be used, so for example `gamescope -U -Y -- supertuxkart` will prefer NIS over FSR, but if you disable NIS with Meta+Y (the Gamescope shortcut to disable it), it will not fall back to FSR. This is different from other behaviour, example `gamescope -i -N -U -- supertuxkart` will prefer FSR over the other filters, but when FSR is disabled with Meta+U, Gamescope will fall back to Nearest Neighbor. And if that is disabled using Meta+N, it will fall back to Integer Scaling. The behaviour with NIS is that it will *overwrite* FSR but it will *be applied on top of* other filters like Integer Scaling and Nearest Neighbor. This is not documented Gamescope behaviour, but I updated the translation for English to make a note of this potential NIS/FSR conflict. I did not mention that NIS can be applied on top of other filters, as I felt that is better suited for the wiki and is likely not to be as jarring as NIS/FSR overwriting each other.

A side effect of this is that if a user were to enable both FSR and NIS in SteamTinkerLaunch, NIS would *always* be preferred as it is always applied last. IMO this is fine as it follows the logical order of the UI, where NIS is below FSR.

<hr>

# Other Minor Changes
Finally, this PR also updates the labels for FSR to explicitly mention that it is FSR 1.0, as opposed to [FSR 2.0](https://gpuopen.com/fidelityfx-superresolution-2/) which is not (yet) supported by Gamescope. This versioning is also more consistent with the NIS option, which explicitly lists its version as v1.0.2. I can't find much on the exact versions of NIS, but I assume there will be updates in the future, and Gamescope may not update right away *if at all* to this new version. Therefore I thought it best to explicitly list the version, just as Gamescope itself does in its documentation. This may add a slight maintenance cost as if/when these versions change, STL will need to be updated to reflect this change. It can be removed if it is preferred to leave the version out :+1: 

I was not able to update some of the translations as I was not sure to add the versioning number to them. For example I was not  sure how to add it to the Chinese translation fully. Also, with some translations such as the German translation, I did not update the existing translated strings to include the new information about FSR and NIS potentially conflicting as I was not sure if that would be preferred to add untranslated parts to existing translated strings. There shouldn't be any strings missing from the translations but they may not all be fully up-to-date. Feedback on this is welcome!

**Note**: The version of Gamescope that this is available in and some versions before that, have been slightly problematic for me with some games, and I have an open issue for it on the Gamescope repo. To be clear, this is unrelated to SteamTinkerLaunch and this PR, just for anyone in the future that might be having issues with Gamescope and stumbles across this :slightly_smiling_face: 

<hr>

As usual, if this PR is accepted, I am happy to update the Gamescope wiki to reflect the new addition. All feedback is appreciated.

Thanks!


frostworx:
Wow, thank you very much for your great contributions again! The wiki is amazing! :)
Before I merge, would you please be so kind and check if all language files are "synchronized" _(so every variable is in the same line in each file)_ This might be already the case, but the line numbers shown in the diff look a bit off.
Also it would be great if you also bumped the 
`PROGVERS` _(in line 9)_, so the configs get autoupdated with the new `GSNIS` default value.

frostworx:
just bumped PROGVERS to "v10.0.20220525" with [this](https://github.com/frostworx/steamtinkerlaunch/commit/fd7a661d6199b78f063e677b21711a8e59c3100c) small commit, so depending on our timing you don't "have" to change it anymore :)

sonic2kk:
Ah, so I think I see where the confusion is coming from. Looking at the diffs, it seems like the translations in every file to add the NIS strings are on lines 614-615. Lines 612-613 were not updated for every string, as these are FSR strings that I did not update for every string as I did not want to mess with translations. So the lines for FSR in other languages were not touched, making it seem like the diffs are a bit off.

As well as this, some of the language files had an update on Line 600-601, which is an updated translation for the Upscaler spinner. It applies to both FSR and NIS. I didn't change all of the language files though as it would result in a partial translation, or removing the translation entirely, which I didn't really want to mess with. For example in the [Chinese translation](https://github.com/frostworx/steamtinkerlaunch/pull/471/files#diff-cd4e54f08f5e507f39025f5389919f436f6e167e967e4dd777ca445cb717ac40R602-R603) I didn't update the translation to specify that the spinner now applies to FSR and NIS, as it is translated into Chinese and I did not want to remove the work of the translators.

Related, it seems like the order of the strings in the Chinese file is slightly different than the other files, as it seems like in every other translation file the spinner text is on Line 600-601, but for the Chinese strings it's on Line 602-603. The length of both the Chinese translation and English translation are identical, so it seems like just the order of the strings has changed.

So from what I can see, the diffs are fine, it's just that some of the translations have more diffs than others. If I've missed something or handled the translation files poorly please let me know :smile: 

EDIT: And thanks for taking care of the version :) If it needs updated again it's no problem!

frostworx:
oh, thank you very much for the extra work in looking into and explaining the language file "problem" :)

frostworx:
❤

sonic2kk:
Thanks! I updated the wiki to contain the new information as well :partying_face:

frostworx:
yay, thank you very much! 🎉

PR #473, opened 2022-05-26
update chinese.txt

ling0412:


frostworx:
wow you're fast! thank you very much again :)

PR #480, opened 2022-06-01
Add DXVK Framerate Limiter option to Game Menu

sonic2kk:
# Overview
DXVK now allows you to [set a framerate limit](https://github.com/doitsujin/dxvk/commit/30a1a29aa612c7a796d3cc1a4b320a24829ed286) per-game using the variable `DXVK_FRAME_RATE`. This PR adds an option to set this variable. A player can enter any FPS limit they wish (as DXVK should accept any value) but there are a few preset options: **30**, **60**,  **75**, **90**, **120**, **144**, **165**, and **240**. These values were chosen as they loosely correspond to common framerate caps/monitor refresh rates.

This option can be found under the DXVK settings on the Game Menu:
DXVK Frameate Limit

This option will cap the framerate at a given value, or if `none` is selected it will let the game decide the framerate. This could be unlimited or whatever the game internally caps the FPS at. Some games, such as Sonic Mania, have a hard cap of 60fps. For this reason I chose to go with `none` instead of `unlimited` as this could be slightly misleading. Having a value of `none` will not have an unlimited FPS cap but will just leave it up to the game to control the framerate.

# Implementation
When a game is launched, the FPS limit value is exported as the environment variable `DXVK_FRAME_RATE={chosen_framerate}`. It has to be set as an environment variable and does not go in the DXVK config file. DXVK will ignore unrecognised FPS values such as `DXVK_FRAME_RATE=gibberish`, so the input box does accept letters which is a little unfortunate as a user could accidentally enter a character and break their FPS limit. Having spaces seems to be fine, both at the start and end of the FPS limit, but it would be nicer I think if any leading and trailing whitespaces were stripped, if only to make the FPS limit box look a bit tidier. Though you could argue the responsibility here is on the user to tinker responsibly :sweat_smile: 

The code for exporting this variable was moved into a new function called `setDxvkVars`, I thought about moving the DXVK HUD configuration there too but decided against it as it has other logic beyond exporting a simple variable. Though right now the only variable being managed for DXVK like this is the framerate limit, breaking it out is cleaner in my opinion and if any future variables are added to be configured we can manage them in this function.

DXVK's frame limiter seems to work just as well as other framerate limiters such as the one in MangoHUD, but the advantage here is that a player doesn't need to install MangoHUD or Gamescope or any other programs to cap their framerate. If a player is using a standard Proton release they are guaranteed to have DXVK. If they are using the WineD3D option (`PROTON_USE_WINED3D=1`) then DXVK options are ignored but in the interest of keeping the environment "clean", the variable is not exported if this option is enabled.

When this option is enabled, it is logged out to the STL log file as `INFO - setDxvkVars - Exporting variable 'DXVK_FRAME_RATE=60' to limit game framerate`.

# Justification
Though this could simply be added as an environment variable in a per-game config file, limiting the framerate is a good workaround for some games that either use too much GPU, that don't have their own framerate cap built in, or that need a framerate cap of 60 to run at normal speeds (such as Sonic Adventure 2).

Limiting the frameate is possible with other tools including ones that STL can configure such as MangoHUD, but this would need to be added as an argument to MangoHUD. Having a dedicated option for limiting the framerate using DXVK removes the need to have MangoHUD installed and also lets you set the framerate limit from the Game Menu using a labelled menu option.

An FPS cap, like other DXVK arguments, could be passed as a custom command to a game, but as framerate limiting is at least a semi-common action IMO it's better to have it explicitly as a setting in the UI. Other options such as DXVK Async are also in the Game Menu, so I think having this option is warranted. The motivation for this feature came from me wanting a nicer way to set an FPS cap so I could be biased in this regard :sweat_smile:

<hr>

I attempted to implement this in the cleanest way possible, by using other parts of the codebase as reference. But as usual, all feedback on this PR is welcome and I am happy to update the wiki if this is merged :slightly_smiling_face:


frostworx:
Wow, thank you very much for another HQ PR, @sonic2kk!
Great to have you on board!
Haven't tested the function, but the code looks excellent!

PR #487, opened 2022-06-07
DXVK HUD Options - Selecting nothing disables DXVK HUD; removed '0' option

sonic2kk:
Updates the DXVK HUD menu to disable the DXVK HUD if no item is selected. Existing behaviour makes the user select the "0" option and if nothing is selected, nothing happens. This PR changes it so that if nothing is selected, the DXVK_HUD variable is automatically set to 0. This is written out to the game's config file as `DXVK_HUD=0` as well.

DXVK HUD options

This does not impact the DXVK Scale value, as it [already has a check](https://github.com/sonic2kk/steamtinkerlaunch/blob/dxvk-hud-disable/steamtinkerlaunch#L14001) for `[ "$DXVK_HUD" != "0" ]`. Not that it would do anything if DXVK HUD is disabled but not setting this is a nice way to keep logging and environment variables clean.

On the implementation of this PR: The call to `updateConfigEntry` twice seems to be necessary, as having it outside of the if/else block seems to result in it not being called when no option is selected. Perhaps my inexperience with Bash is showing so if there's a better way to implement that I'm more than happy to clean up that part of the code :smile: 

**Note**: This menu has some probably-not-ideal behaviours at the moment. Selected variable(s) won't appear in the menu correctly until you close STL and open it again. I'll open a separate issue about this and some of my other ideas for the DXVK HUD options, but it's not related to or introduced by these changes :slightly_smiling_face: 

In my opinion this change is more user friendly, but discussion about this is very much encouraged. I had the idea for this PR when fiddling with the DXVK HUD options myself and finding the behaviour a bit unusual. All feedback as usual is welcome!


frostworx:
Sound and looks reasonable to me. Thanks for the improvement! :)

PR #491, opened 2022-06-07
Fix replay sorcery check

### LINKED ISSUE #489: Replay Sorcery failing with already running
this is because the code here https://github.com/frostworx/steamtinkerlaunch/blob/73c0ea180a332eb744ee64d1fb1ab7364e13d2a4/steamtinkerlaunch#L10806= also finds the kms systemd service `replay-sorcery-kms.service` which is needed for HW encoding 

I'll have a PR for this shortly

zany130:
Make sure it only looks for `replay-sorcery` and not  `replay-sorcery kms` Used `which` to get the path not sure if this is really necessary or even a good idea
fixes https://github.com/frostworx/steamtinkerlaunch/issues/489

First PR ever please let me know if anything can be done better (:

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -10804,7 +10804,7 @@ function useNyrna {
 function useReplay {
 	if [ -n "$RUN_REPLAY" ]; then
 		if [ "$RUN_REPLAY" -eq 1 ]; then
-			if "$PGREP" -f "$REPLAY" >/dev/null; then
+			if "$PGREP" -fx "$(which "REPLAY") >/dev/null; then
 				writelog "SKIP" "${FUNCNAME[0]} - '$REPLAY' already running - skipping"
 				RUN_REPLAY=0
 			else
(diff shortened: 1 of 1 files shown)


frostworx:
Thank you and congratulations for your first PR! :)
Doesn't look bad, but you forgot a closing `"` after `)`
Also, it would be nice if you'd use `command -v` instead of `which`

`which` is not wrong, but `command -v` is bash-builtin and I use it everywhere else in the code.

frostworx:
Perfect! Thank you very much! :)👍

zany130:
@frostworx I think I made an oopsie shouldn't it be `if "$PGREP" -fx "$(command -v "$REPLAY")" >/dev/null; then` not `if "$PGREP" -fx "$(command -v "REPLAY")" >/dev/null; then` so missing the `$` in front of REPLAY noticed it was always passing that condition even if `replay-sorcery` was already running.

EDIT: I originally thought it should be `$"REPLAY"` and it should be `"$REPLAY"` I think (no expert in bash lol)

EDIT2: lesson learned actually test you code when doing a pr 🤦

frostworx:
hah, good find! will push a fix in a sec :)

edit: done

Issue #492, opened 2022-06-07
Custom DXVK Config File Not Loading


sonic2kk:
STL exports `DXVK_CONFIG_FILE` to the correct path and I have verified it finds it correctly if the path exists (and if it's uncommented from the game's configuration file). However it is not loaded with the game. I tested this by adding `dxvk.hud = fps` to the config file.

Interestingly, launching a game without STL and adding `DXVK_CONFIG_FILE="/home/user/.config/steamtinkerlaunch/dxvk/appid.conf" %command%` works fine. The DXVK wiki page [specifically notes](https://github.com/doitsujin/dxvk/wiki/Configuration#configuration-file) that this is an environment variable though.

Perhaps this should be part of the launch command?

frostworx:
hm, very strange. it is exported so it should just work.
I'll add a writelog line, maybe it helps to find the issue.

frostworx:
done :)

sonic2kk:
Thanks, that log line outputs:
```
Tue Jun  7 19:30:19 BST 2022 INFO - prepareProton - exporting variable 'DXVK_CONFIG_FILE' pointing to '/home/user/.config/steamtinkerlaunch/dxvk/584400.conf
```

Not entirely sure why games aren't seeing it since that variable does work if a game is launched without STL.

frostworx:
This "function" is really ancient, and I'm not sure if it really makes much sense it its current state.
might be a good opportunity to re-think it generally. tbh it looks pretty ugly :)

frostworx:
> Thanks, that log line outputs:
> 
> ```
> Tue Jun  7 19:30:19 BST 2022 INFO - prepareProton - exporting variable 'DXVK_CONFIG_FILE' pointing to '/home/user/.config/steamtinkerlaunch/dxvk/584400.conf
> ```
> 
> Not entirely sure why games aren't seeing it since that variable does work if a game is launched without STL.

thanks for testing! strange indeed. need to look deeper into this apparently...

frostworx:
hm strange, dumping the whole env directly before the game launch doesn't reveal anything useful as well.
the var exists and the path is correct.

frostworx:
I'm not sure if this is really a blocker for #493. as already mentioned the `DXVK_CONFIG_FILE` variable is exported fine here.
will have to test a bit more soonish.
Do you have an example dxvk config which makes it very obvious that it was loaded? _(tbh I barely played with it yet generally)_

[...]

PR #493, opened 2022-06-07
Add DXVK and MangoHud config files to Game Files list

sonic2kk:
This PR adds an option to edit the game-specific DXVK and MangoHud config files from the Game Files menu for ease of access. I put these both in one commit as they're both config files and relatively the same action is performed for both.

image

## MangoHud Config File
For the MangoHud config, the path is just added to the Game Files menu if it exists. From my understanding of the code, the file should always exist if MangoHud is installed (it just pulls from the example config in `/usr/share/doc/mangohud` I believe).

## DXVK Config File
For the DXVK config I made a bit of a change. In the game-specific config file, the DXVK Config file path seems to be commented out, so it's not loaded by default. I chose to uncomment this line for a few reasons:
- This will likely override any DXVK config file in the game's EXE folder, but if a user is using STL they'll probably want to use STL's DXVK config file anyway
- If the file is blank or does not exist, nothing happens, so there is no downside from what I can see to having it enabled
- The file is really only for advanced users, so they should be able to resolve anything that is messed up in this file
- If a non-advanced user messes something up with this file, they now have an easy way in the STL UI to find this file and delete it :slightly_smiling_face: 

This leads into my last change, which is creating the DXVK config file if it does not already exist. This is done in `createCustomCfgs` which is hijacking the existing function a little bit I suppose, but I thought it was relevant to create this file here  :sweat_smile: Basically if the file doesn't exist at the expected path, it creates a new one with a warning that this file is generally for advanced options, and points to a link to the upstream DXVK example config.

I wanted to see about creating the DXVK config file using the upstream config as a template much like STL does for MangoHud, but I couldn't think of a clean way to implement this without having to basically store the DXVK config somewhere in STL and load that, which is not exactly ideal and would mean adding it to the Makefile for installation, which is something I didn't want to mess around with too much :smile: 

<hr>

Working on this PR is how I discovered #492, but I thought I would open this PR anyway at least for review and discussion. I'm happy to hold off merging this until there is a resolution to that issue, as the DXVK part of this PR is not very useful yet. Thanks!

P.S. commits were not squashed because I wasn't too sure how to do it anymore (it's been a while), sorry about that.


frostworx:
Thanks a lot again! 😀
Basically this looks already good, but I'm not convinced that creating an "empty" dxvk.conf for every single game is the best solution.

sonic2kk:
Hmm, that's a good point. I checked the MangoHud config files folder and just assumed it was done for MangoHud as well, but upon testing it seems like the MangoHud config file is only created if the option is ticked in the Game Menu.

Perhaps we could copy this behaviour for DXVK and add a checkbox to load a DXVK config file, and if that option is ticked we could create the game's config file? That would help keep the `stemtinkerlaunch/dxvk` folder clean too

frostworx:
Above commit should be the best option to create the `$STLDXVKCFG` file.

sonic2kk:
Looks good to me, thanks for the input! 🎉 I'll take a look at editing the PR later tonight

frostworx:
no hurry, my commit doesn't do anything without your PR :)

sonic2kk:
Okay, finally got around to amending the PR. Basically what I did now was:

- Remove the creation of the DXVK conf file in the `createCustomCfgs` function
- Remove the check for the DXVK conf file existing before adding it to the Game Files array. Otherwise it wouldn't show up as an option for the user to select, and so the commit for creating the file if the user chooses to open it would do nothing since they'd never get to the point of trying to open it :-)
- Moved the adding of the DXVK conf file and MangoHud conf file to above the `createCustomCfgs` as I felt this was a little bit cleaner since the DXVK conf file isn't created in this function anymore. Now the call to `createCustomCfgs` is right above adding the custom game vars paths to the array

Hopefully this is enough, but if I've missed something please let me know :smile:  

EDIT: I believe this is still okay but to draw attention to it anyway, the DXVK conf variable is uncommented in the per-game config file. I think this is okay as it shouldn't be exported or appended anywhere unless the file exists I *think*. However I suppose if it's wanted we *could* have a check to basically uncomment the line in the game config when the DXVK conf file is created, but I'm not sure if that's totally necessary. Thoughts? :slightly_smiling_face:

frostworx:
Looks good so far, thanks.
But:

- I'd prefer the dxvk project page not hard-coded in the language files. Instead, use something like `XXX/blob/master/dxvk.conf`
  and call the var via `strFix "$STLDXVKCFG_WARNING" "$DXVKURL"`
- I'm uncertain if commenting in `STLDXVKCFG` by default AND using it automatically by doing so is the best decision, but I agree it should be immediately a valid option from the beginning and therefore should be commented in by default _(yes, it should be also commented in for already existing configs)_. Probably it is the best idea to simply wrap a new `USE_STLDXVKCFG` opton around it and let the user decide. Thoughts? 😀

sonic2kk:
So as of my last commit I addressed the point of the DXVK config URL. Now the config file has `XXX/blob/master/dxvk.conf` and `strFix` puts in the URL from a variable. I put this variable right above the line to echo this into the DXVK config file, but if it should be in another place (`setDefaultCfgValues`?) I can change it no problem :slightly_smiling_face: 

The last point is a bit of a tricky one. I certainly see the logic in having a separate `USE_STLDXVKCFG` option, perhaps in the Game Menu just like the one for the MangoHUD config. But if a user selects a DXVK config file from the Game Files menu they might expect it to be enabled right away. Here is my idea for a solution:

- By default, there is no DXVK config file anywhere and the `STLDXVKCFG` path in the per-game config is commented out
- We have an checkbox option in the Game Menu for `USE_STLDXVKCFG`, which when enabled and saved, uncomments `STLDXVKCFG` and creates the DXVK config file.
  - This is where it gets a bit tricky. The MangoHUD option for this (`USEMANGOHUDSTLCFG`) does automatically create a MangoHUD config file, so I would say we want to do this for DXVK. Also there is a case where a user might enable this and want to browse to the file without using STL, and may be confused if the file doesn't exist. However it could also be argued that this creates clutter. I would still lean into copying the behaviour for the MangoHUD config option
- Check when STL opens if a DXVK config file for the game already exists in the path, we should enable the `USE_STLDXVKCFG` option and uncomment `STLDXVKCFG` from the Game Menu.
  - The same should also probably go for MangoHUD's config file if it doesn't already 
- When a user selects the DXVK Config File option from the Game Files menu, it should enable the `USE_STLDXVKCFG` option as well, since if a user has just created the file they would probably expect it to be used.
  - Again we could probably copy this behaviour for MangoHUD

Maybe that should go in another PR, but that's the solution I had in mind. Basically I agree with having the `USE_STLDXVKCFG` option and I think we should be smart about making sure its enabled if a config file exists :smile:  

P.S. No rush on feedback for this :-)

frostworx:
Thanks for the bump here _(wow, you've been really busy everywhere while I was away 😀)_
Will need some days to re-dive into this :)

frostworx:
Haven't checked back everything yet and start RL work in a few minutes, but the variable `DXVKURL` is already available via 
`function setDefaultCfgValuesurl` _(→ `url.conf`)_ and therefore doesn't need to be hard-coded

frostworx:
I agree the DXVK config file `STLDXVKCFG` should be created when `USE_STLDXVKCFG` is enabled
and It should also be created when the `STLDXVKCFG` is selected in the editor menu (`USE_STLDXVKCFG` will be enabled automatically simultaneously)

Imho `STLDXVKCFG` should be commented in always, as there is no point anymore to keep it as the only variable which is commented out at all when we introduce `USE_STLDXVKCFG`

We should add a temporary migration function which enables `USE_STLDXVKCFG` automatically when `STLDXVKCFG`
is already commented in _(only when `PROGVERS` is below a specific version)_
The same function should comment in `STLDXVKCFG` when commented out, but not enable `USE_STLDXVKCFG` here.

I think we covered every sane constellation :)
What do you think? :)

sonic2kk:
Oops, sorry for the delayed response (didn't see this in my emails yesterday :sweat_smile:)

If I understand right I think this makes sense, basically always have `STLDXVKCFG` uncommented in the per-game config file and make a temporary function that will "link up" existing config files that have the `STLDXVKCFG` variable enabled. So if someone has explicitly enabled the variable it will automatically enable the `USE_STLDXVKCFG` option in the Game Menu for them. The function will also comment `STLDXVKCFG` in for all configs that have it disabled (since it should always be enabled now) but don't enable it in the Game Menu as these users haven't explicitly tried to enable a DXVK config file.

Assuming I got all that right, I think that's a really solid approach :smile: 

> Haven't checked back everything yet and start RL work in a few minutes, but the variable DXVKURL is already available via
function setDefaultCfgValuesurl (→ url.conf) and therefore doesn't need to be hard-coded

Should be addressed now. ~~but I noticed a slight problem with the PR that I'm looking into addressing (don't remember it being there before). The DXVK Config path seems to always be blank now so I'll take a look into fixing that :)~~ Seems that problem was just me being silly and forgetting some context haha, when `STLDXVKCFG` is commented of course there is no path to show up :)

I also just remembered that this PR and our discussion is going to be held back a bit by #492.

frostworx:
looks like we are in agreement :)
How to proceed? Do you want to make the missing changes and we leave this open and merge it all together,
or should I add it? Both fine for me :)

_(are you sure your force-pushes are correct? It seems a bit strange to look into the changes, so I'm afraid that merging will be a mess. Once someone did something similar (no idea if this was the same) and after the merge all my commits were their commits in the history. was a mess to fix this manually later)_

sonic2kk:
Hmm you're right about the force pushes, there might be something wrong there (I vaguely remember something going wrong when I had to update my branch). I'll have a look into it soon and see if I can squash them but at worst I can close this and make a new branch and PR :) Sorry about that! :sweat_smile: 

As for the other changes, I think it's fine if you make them (and then we merge this/a different PR with fixed commits). I think you could architect it better than I could and if there are any adjustments needed for this PR's changes I can deal with those :)

Thanks for your patience and discussion here, it's much appreciated.

frostworx:
Oh, thanks a lot for looking into the force commits :)
I suggest we wait until your PR is ready, then I'll merge it and add all discussed steps afterwards _(else it might get confusing - at least for me :grin: )_
Thanks mate! 😀

sonic2kk:
Very sorry, I'm not entirely sure how to fix this branch :sweat_smile: and for such a relatively small change it was easier for me to  simply make a new branch. I'll open a new PR for the changes on that branch very shortly, referencing this PR and adding in some of the discussion we had here about how the PR will fit into future changes :smile:

frostworx:
I probably would have done exactly the same :)
Thanks for the extra work :)

PR #500, opened 2022-06-20
Steam Deck: Add ability to choose specific version of innoextract and downgrade innoextract from latest to 1.9-5

damian-anlauf:
The latest version of `innoextract` (currently 1.9-6) will not work on the Steam Deck as it needs a newer version of the `libboost_iostreams.so` library (needs: 1.79.0; currently on Steam Deck: 1.78.0) (mentioned in #272).

This PR adds the ability to choose a specific version of `innoextract` and downloads it from the [official Arch Linux archive](https://archive.archlinux.org/packages/i/innoextract/). 
The version of `innoextract` set to download in this PR is 1.9-5 as this is the last version that works with `libboost_iostreams.so.1.78.0`.

All modifications have been done in the `steamdedeckt` function and no variables have been changed (only added), so it should not affect other parts of the code and thus not affect people who are not on Steam Deck.

P.S. This is my first PR. Please be gentle :smile:.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -19270,6 +19270,11 @@ function steamdedeckt {
 		SDREPO="https://steamdeck-packages.steamos.cloud/archlinux-mirror/extra/os/x86_64/"
 
 		ARCHURL="https://archlinux.org/packages/community/x86_64"
+		ARCHARCHIVEURL="https://archive.archlinux.org/packages"
+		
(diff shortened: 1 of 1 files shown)


frostworx:
Thank you for the PR! 👍
I like the idea to use the archive arch url.
maybe we should use it generally for automatic package downloads on the deck.
I'll bump #272 for a discussion.

PR #503, opened 2022-06-25
Add DXVK and MangoHud config files to Game Files list

sonic2kk:
Fixed version of #493, with what should be identical changes but with fixed commits :smile: 

This PR adds the option to open a per-game DXVK config file (at `$STLCFGDIR/dxvk/appid.conf` by default) and the existing per-game MangoHud config if it exists (existing behaviour is, this file won't be created if `USEMANGOHUDSTLCFG` is not enabled).

DXVK and MangoHud config paths now in Game Files menu

I also changed the generated per-game config file to uncomment the `STLDXVKCFG` variable but this will **only** apply to newly generated config files. This variable already existed but was commented out. The Game Files menu relies on this variable to find the path to the config file. If this variable is commented out (as it would be for all existing game configs) then the path will be blank, so for this iteration of the PR I added a check to make sure the STLDXVKCFG variable is defined. I did not add a check for if the DXVK config file exists, as right now we always create it if the user tries to open it from the Game Files menu.

All feedback is welcome :smile:


frostworx:
Thanks mate!
Will add the other features discussed [here](https://github.com/frostworx/steamtinkerlaunch/pull/493) shortly.

sonic2kk:
No problem, thank you for all the patience with this feature :sweat_smile:

frostworx:
Just pushed [this](https://github.com/frostworx/steamtinkerlaunch/commit/de6debb9758bead0a336b0400764e0d7151b265b).
Should be feature complete. Of course you're welcome to open an issue if something doesn't work as expected or I forgot something

frostworx:
mh, when disabling `USE_STLDXVKCFG`, the function `migrateCfgOption` re-enables it automatically when the `STLDXVKCFG`
was found...

sonic2kk:
Hmm, could it be handled in a similar way to `USEMANGOHUDSTLCFG`? Though I suppose the way the DXVK config works is slightly different, we want to make sure `USE_STLDXVKCFG` is enabled if users enabled it using the "old style" of uncommenting it in the per game config.

I guess the problem is STL can't tell when a user explicitly disabled `USE_STLDXVKCFG` and when a user never had the option to enable in the first place f.e. they migrated from an older version.

I wonder, is it possible to check if the user had `USE_STLDXVKCFG` in their config before? If they didn't (they're coming from an older version) and they had a valid `STLDXVKCFG` variable (they enabled it in their per-game config from an older version), `USE_STLDXVKCFG` should be enabled. If the `USE_STLDXVKCFG` variable already existed they probably don't need it automatically enabled. I'm not sure when variables get written out to the config file and so I'm not sure if this sort of check is feasible.

I was thinking something along the lines of:
```
if USE_STLDXVKCFG is disabled then
  if STLDXVKCFG is found then
    if USE_STLDXVKCFG is in the user's config file then
      # User explicitly disabled USE_STLDXVKCFG
    else
      # User had a DXVK config file enabled from a previous version 
    fi
  fi
fi
```

No idea if that makes sense, I did have a skim through the code to try and see if this might fit anywhere but I'm honestly not sure. This seems like a tricky problem to solve

frostworx:
the only sane way is imho a temporary list generated in `migrateCfgOption`, where seen/updated appids are stored.
something like ~/.config/steamtinkerlaunch/temp/dxvk-${major-steamtinkerlaunch-version}.txt probably is the best choice.

so the `migrateCfgOption` finally will both update new features and remove those lists from a previous update.

frostworx:
[this bump](https://github.com/frostworx/steamtinkerlaunch/commit/96b3b75fb220e514964659246e797e83c6be4b16) should be enough. I doubt that many users used the commented out `STLDXVKCFG` variable anyway.

sonic2kk:
Was a bit busy for a few days sorry about that, but thank you for the work on this :partying_face: And I agree, having a specific DXVK conf was probably not a very common use case, though now it'll be a much easier process for anyone that wants to tinker with DXVK :smile:

Issue #507, opened 2022-06-29
 [FEATURE] add new FSR resolution options


zany130:
see here https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton7-24

~~so the whole fsr thing for proton ge has to be reworked now I guess?~~

EDIT: my bad fsr strength is still there and works how it did before.  It seems there are now new options for forcing a resolution?

sonic2kk:
It seems that way, though doesn't Wine-tkg use the "old" variable method? Just looking by the release notes on this (very old) build: https://github.com/Frogging-Family/wine-tkg-git/releases/tag/7.6.r12.g51472395

> Comes with support for AMD FSR through Proton's fullscreen hack - Enable with WINE_FULLSCREEN_FSR=1 env var, and tweak sharpening amount with WINE_FULLSCREEN_FSR_STRENGTH=2 (2 is default sharpening, higher numbers mean more sharpening)

Going purely on what I've seen (my own experience and ProtonDB), GEProton is more popular than Tkg. Maybe we should have the new-style GE FSR options as a separate menu option, and only apply the GEProton variable if a user is using a version of GEProton >= `GE-Proton-7-23` (I believe this is the newest version with this option, as I don't think 7-22 did [looking at the changelog](https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton7-22), it seems like it relied on a different option that has been superseded by this new `WINE_FULLSCREEN_FSR_MODE`).

If this stays in GE I think it would be great for STL to have this option, but given how features like this in GE can be added and reverted in short spaces of time, maybe we should wait to make sure this will be a more permanent fixture before jumping the gun? Just my thoughts though :smile:

zany130:
it seems these are additional fsr options (which seem  to be not  available on tkg proton yet?) fsr strength is still used

sonic2kk:
Ahhh I see, so FSR strength is still used by `WINE_FULLSCREEN_FSR_MODE` seems like it sets resolution?

> Method 1:
> `WINE_FULLSCREEN_FSR=1 WINE_FULLSCREEN_FSR_MODE=ultra %command%`
> 
> with the options of ultra, quality, balanced, performance and it will auto-add the corresponding FSR resolution in-game for your monitor resolution if supported. You then go in game, set fullscreen mode, and choose that resolution if it is not already automatically set.

[...]

PR #509, opened 2022-06-30
Add option for new FSR options added in GE-Proton7-24 and up

sonic2kk:
Implements #507. Big thanks to @zany130 for suggesting and also helping out with research for this PR :+1: 

[GE-Proton7-24 and above](https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton7-24) have two new Wine FSR variables: `WINE_FULLSCREEN_FSR_MODE` and `WINE_FULLSCREEN_FSR_CUSTOM_MODE`. This PR adds menu options to set both of these options.

FSR Upscaling Mode and Custom Mode comboboxes

## How to use `WINE_FULLSCREEN_FSR_MODE`
This option adds a new fullscreen resolution option which changes a game's list of resolutions based on the user's display resolution. It takes **only one** of four options: `performance`, `balanced`, ` quality`, `ultra`. Based on these options and the user's screen resolution it adds a new resolution to the game based on this. For example, if a user is playing on a 2k screen and selects the `balanced` option, they'll have a fullscreen resolution available in-game of `1506x847`. Usually, this is **selected by default**.

The `WINE_FULLSCREEN_FSR_STRENGTH` is still used and this dictates the strength of the upscaling from the resolution set/added by `WINE_FULLSCREEN_FSR_MODE`.

See also **Page 10** of this OpenGPU FSR 1.0 spec PDF, which lists the recommended resolutions for the four FSR modes: https://github.com/GPUOpen-Effects/FidelityFX-FSR/blob/a21ffb8f6c13233ba336352bdff293894c706575/docs/FidelityFX-FSR-Overview-Integration.pdf

## How to use `WINE_FULLSCREEN_FSR_CUSTOM_MODE`
This option is used for setting specific output resolutions for a game, such as ultrawide displays. To use this option normally and with SteamTinkerLaunch, a user is expected to input a calculated resolution ***to upscale from***, not their display's resolution. It is intended that a user will take their monitor's resolution and calculate the appropriate resolution to upscale from based on whether they want `performance`, `balanced`, `quality`, or `ultra`. GloriousEggroll notes how to calculate the appropriate resolution at the bottom of the release notes for GE-Proton7-24.

In my tests, these resolutions are typically selected by default, but may show up as strange values in the Game's resolution list. For example, Sonic Mania lists it as "Default" and DuckTales Remastered lists it as "Invalid Index!". Mileage may vary when using this option but this happens even when using the environment variable through Steam, so I don't think it's a problem on STL's side.

The dropdown for FSR Custom Modes lists all the preset resolutions as noted by GloriousEggroll in the release notes for GE-Proton7-24. I created an array with these values (in case more or added or a semi-common set of resolutions are found and someone wants to add them in future). I wasn't sure where to put it so if there is a preferred place please let me know :slightly_smiling_face: 

`WINE_FULLSCREEN_FSR_CUSTOM_MODE` will always override `WINE_FULLSCREEN_FSR_MODE`.

## Considerations
I did a fair bit of testing on this to try and understand how these variables work (I documented a lot of it as I went on #507), and even though I volunteered to write this feature, I do have a slight concern. GE-Proton can sometimes have features added and removed depending on feedback, and as this feature is (at the time of writing) less than 24 hours old there is a chance it could be dropped or altered in the next release/hotfix.

Also, as far as I know this is the only release of Proton to support this, I don't think Tkg builds support this. Maybe we should leave it up to the user to apply common sense and only use these options if they are using a build of Wine/Proton that supports this, but if it's preferred I can add a note about this to the descriptions in the language files :smile: 

If this feature is added I'm happy to create/update a wiki page to document more of how this works, as it did take a while for me to wrap my head around it. 

As usual, all feedback is welcome! Also, happy Steam Deck Q3 email day! :-)


frostworx:
It is always impressive to see your perfect PR descriptions. I would require multiple hours for each and nobody would understand what I meant in the end 😀
Thank you very much for another PR.
Changes look good as far as I can tell. I would have put the array into one line, but it is also fine multiline _(maybe even better for readability)_ The position in the code is fine as well :)

You spent lots of time into this so of course I'll merge it - even if GE drops it later again - for now it is a very nice feature!
Thank you here as well to @zany130 for reporting! :)
IMHO there's no need to add any additional hints to the language files, but your offer to update the wiki is very welcome of course.
Maybe you want to add a corresponding note there instead.

sonic2kk:
Thank you for always taking time to review and merge my PRs 😄 I suppose the array is pretty tall but I had readability in mind, though it could of course be changed in future 🙂 

I'll get this info up on the wiki ASAP and mention it in the pinned wiki issue once it's up 🥳

frostworx:
Thank you very much! 🤓

PR #515, opened 2022-07-03
ReShade: Fix detection of DLL conflicts between dxgi.dll and d3d9.dll

### LINKED ISSUE #514: [Problem] SteamTinkerLaunch installs too many ReShade DLLs
- Version of STL: https://github.com/frostworx/steamtinkerlaunch/commit/0706fb2b63931cafea47fb6dbd1de330fc96220a
- Game: L.A. Noire on Steam
- Runner: GE-Proton 7.24
- ReShade options (global): Version 5.2.2
- ReShade options (game config): Checkboxes for "Enable, Install, Update" are all enabled.
- Problem: SteamTinkerLaunch installs both d3d9.dll and dxgi.dll into the game folder. The game instantly quits as a result.

This is what the `ReShade_LANoire.log` says:

```
21:18:03:796 [01252] | INFO  | Initializing crosire's ReShade version '5.2.2.1534' (32-bit) loaded from "Z:\home\username\Games\Steam\steamapps\common\L.A.Noire\dxgi.dll" into "Z:\home\username\Games\Steam\steamapps\common\L.A.Noire\LANoire.exe" ...
21:18:03:796 [01252] | WARN  | Another ReShade instance was already loaded from "Z:\home\username\Games\Steam\steamapps\common\L.A.Noire\d3d9.dll"! Aborting initialization ...
21:18:03:796 [01252] | INFO  | Exiting ...
21:18:03:796 [01252] | INFO  | Uninstalling 0 hook(s) ...
21:18:04:797 [01252] | INFO  | Finished exiting.
```

So I started the game with STL again, and while the game was still processing its launcher, I quickly delete d3d9.dll from the game folder. This allows the game to launch properly with ReShade.

The problem is that the game is loading both dxgi.dll and d3d9.dll because it supports both graphics APIs, and as a result ReShade force-quits the game due to the conflict.

Is there any workaround to prevent STL from constantly putting d3d9.dll into the folder? I tried making a 0-byte `d3d9.dll` which is unwritable, but the game itself crashes if I put an empty dummy there...

Perhaps it's too hard to solve via autodetection in case a game uses multiple DirectX architectures, so perhaps it would be best to just add individual checkboxes like this?

- [ ] Don't override d3d9.dll
- [ ] Don't override dxgi.dll
- [ ] etc...

Arcitec:
Closes #514.

- Efficiently scans the game's exe/dll folder (-maxdepth 1) for the legacy
  "dxgi.log" name and the modern "ReShade_exenamehere.log" name.

- If log files are found, they're passed into grep in "matching filenames"
  mode (-l), which does a quick scan that stops when a match is found.

- We use "wc -l" to count log files with conflicts (RESHADE_CONFLICTS).

- This accurately detects games where conflicts occurred between the
  two DLL files, for both old and new ReShade versions (log names).

- The rest of the code is the same as before. Note that we do the scan
  even if "$NOD3D9" is already set, just to ensure that the DLL gets
  deleted if it ever gets re-added by anything. The scan code is very
  fast, so it's fine to do the scan on every startup.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7676,30 +7676,26 @@ function installReshade {
 
 		INSTDESTDIR="$SHADDESTDIR"
 
-		function maybedrop { # TODO remove?
-		# checking for previous dll conficts between $D3D47 and $RS_D9_DEST
-		if [ -f "$INSTDESTDIR/${RS_DX_DEST//.dll/.log}" ]; then
-			if grep -q "Another ${RESH} instance was already loaded" "$INSTDESTDIR/${RS_DX_DEST//.dll/.log}"; then
-				writelog "INFO" "${FUNCNAME[0]} - Found $RS_DX_DEST conflict with $RS_D9_DEST"
-				if [ -f "$INSTDESTDIR/$RS_D9_DEST" ]; then
-					writelog "INFO" "${FUNCNAME[0]} - Removing $RS_D9_DEST"
-					rm "$INSTDESTDIR/$RS_D9_DEST"
-				else
-					writelog "SKIP" "${FUNCNAME[0]} - $RS_D9_DEST not found"
-				fi
-				
-				if [ -z "$NOD3D9" ]; then
-					writelog "INFO" "${FUNCNAME[0]} - Blocking re-installation of '$RS_D9_DEST' by setting NOD3D9=1 in '$STLGAMECFG'"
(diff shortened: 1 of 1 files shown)


frostworx:
looks good. thanks for the PR.
good to see the commented function actually has some use.
I wasn't sure which, so I commented it once :)

Arcitec:
@frostworx Yep it works well now. Thanks for the guidance. :)

PR #521, opened 2022-07-07
Fall back to xdg-open in general for EDITOR or BROWSER for #519

### LINKED ISSUE #519: xdg-open is possibly handled badly
STL doesn't use `xdg-open` properly.

In Flatpak nothing is set by default such as Editors or Browsers due to sandboxing (over-simplification).

However, Flatpak, using portals, can spawn a Window asking the user what to use (I assume it stating the default app is the portal saying so when the window is spawned, but before this it isn't accessible):
image

Outside of a Flatpak `xdg-open` generally seems to attempt best-guessing (or the DE does anyway) for opening files it doesn't understand from the mime list:
```
xdg-open dsakd.garbagelol 
kf.service.services: KApplicationTrader: mimeType "x-scheme-handler/file" not found
kf.service.services: KApplicationTrader: mimeType "x-scheme-handler/file" not found
```
But then Kate opened anyway, as my default text editor.

It seems odd that STL fails to open files via `xdg-open` when it can't figure out an editor instead of just attempting best guesses, especially seeing as `xdg-open` has numerous exit codes to handle certain scenarios:
```
Exit Codes

An exit code of 0 indicates success while a non-zero exit code indicates failure. The following failure codes can be returned:

1

Error in command line syntax.

2

One of the files passed on the command line did not exist.

3

A required tool could not be found.

4

The action failed. 
```

It appears the issue is happening in `updateEditor`?

Maybe instead of failing if an editor can't be set, fallback to just letting `xdg-open` do a best guess?

This would fix `xdg-open` not working in Flatpak as far as opening files goes, and maybe help handle some general edge cases.

HanPrower:
Attempt to fix #519 more cleanly.

### Reasoning
I would argue that most casual users would expect that trying to open a file or link would fall back to their default application without setting a specific option. This is exactly what `xdg-open` is for.

It shouldn't be a special Flatpak case due to this.

I would expect that if we fork opening files with the trailing `&` it should avoid any "stuck open" issues.

During tests this worked as expected, although in general it appears most uses of opening an external application forked cleanly anyway (i.e. no game hung open after accessing an EDITOR or BROWSER).

Also if something does "hang" open the game, Steam has a big "STOP" button to reap the process anyway. Which should more than suffice for an end user.

### The fix
For the editor I pulled out and tidied most of the recent changes in favour of having a soft selection of `$XDGO` in the actual file opening process.

As for a browser the same is done during `checkHelpUrl`, with the added protection of not trying to kill the `$BROWSER` if `$XDGO` is used, as this seems to want to kill Steam instead.

As this is a soft fallback, nothing is set in the configuration files to avoid potential issues.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -4299,7 +4299,7 @@ function EditorDialog {
 	
 					writelog "INFO" "${FUNCNAME[0]} - Opening Editor '$STLEDITOR' with selected Config Files"
 					mapfile -t -O "${#EdArr[@]}" EdArr <<< "$EDFILES"
-					"$STLEDITOR" "${EdArr[@]}"
+					"$STLEDITOR" "${EdArr[@]}" &
 					unset EdArr
 					# kill browser if it was opened with the editor:
 					if [ -n "$KILLBROWSER" ]; then
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -4070,25 +4070,15 @@ function updateEditor {
 
 	loadCfg "$CFG" X
 	if grep -q "$XDGO" <<< "$STLEDITOR" || [ ! -f "$STLEDITOR" ] ; then
-		if [ "$INFLATPAK" -eq 1 ]; then
-			writelog "INFO" "${FUNCNAME[0]} - Using '$XDGO' directly, because running in flatpak"
-			writelog "WARN" "${FUNCNAME[0]} - This might prevent a started game from stopping!"
-
-			FOUNDEDITOR="$XDGO"
-		else
-			writelog "WARN" "${FUNCNAME[0]} - '$XDGO' selected as editor or configured editor not found - trying to find an installed editor installed"
-			if [ -x "$(command -v "$XDGMIME" 2>/dev/null)" ]; then
-				XDGED="$(command -v "$("$XDGMIME" query default text/plain | cut -d '.' -f1)" 2>/dev/null)"
-				if [ -x "$XDGED" ]; then
-					writelog "INFO" "${FUNCNAME[0]} - $XDGMIME points to '$XDGED', which also exists"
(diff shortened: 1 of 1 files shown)


frostworx:
thanks here as well for the PR.
_(no need to explain xdg-open or big STOP buttons)_

I'm very sure though that "&" was the first thing I tried initially when creating the editor as it is very obvious to try it first.
It might work for your situation, but might not work with others(?). If it was tested carefully, I wouldn't mind adding it of course.

If `xdg-open` would have returned as expected (non-blocking) I would have simply used it as intended of course.
It didn't, and therefore I picked the working xdg-mime solution like already explained.
I don't want steamtinkerlaunch to be responsible for preventing a game from closing successfully - so the big STOP button is no solution.
Won't find the time before holidays to review the bigger part of the commit, so leaving this open for now.

HanPrower:
Please don't respond until after your holiday. I'm happy to wait :)

Fair, and I agree that using STOP would be frustrating and unwanted. However, from my own experience, trying to use the "EDITOR" and nothing happening is more frustrating to a user.

Most systems should handle `xdg-open` some way or another by default.

As for it potentially blocking, I can only assume at the time it was a bug in `xdg-open`? I can't see mention of any other issues with it doing that. I only find people who want to block it so they can set variables in files whilst running scripts, heh.

If it does block I'd argue it's not something you should work around with `xdg-mime` it should be reported upstream to try and get a fix. If a program is used as intended and it doesn't work (especially one that claims it's meant to make everything simpler) it needs fixing in the program, not worked around.

But of course, it's up to you. Even if `xdg-open` has to be fenced in `INFLATPAK` as a fallback again, at least Flatpak should guarantee it'll open with a `portal`

frostworx:
I'd say let's try it and see if someone complains :)
Thanks for the PR! :+1:

PR #522, opened 2022-07-07
Slightly dirty fix for config "none" settings

HanPrower:
### Reasoning
Currently anything in the UI that says "none" is set as such instead of left blank (besides a handful of options that like being "none").

It seems odd to populate numerous options with "none" when they can quite happily be left blank. It adds unnecessary byte bloat to configs (which can be numerous), and whenever checking options everything needs to be checked against both `$NON` and `<blank>`. If things are left blank where expected then options _should_ only need to be checked if they're blank.

### The fix
Is to check if an options in `updateConfigEntry` is "none" is to set it to <blank> instead, unless it's (currently) 3 special `$CFGCAT`s that can be blank.

It's slightly dirty, and hard to read, but I couldn't think how else to do it, besides just an extra if after the `DUMMY` part. That'd be vaguely more legible.

### Caveats
Every time an option that can be really set to "none" is added this bit of code would need to be remembered to be updated.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -8839,7 +8839,10 @@ function updateConfigEntry {
 						CFGVALUE="0"
 					fi	
 			
-					if [ "$CFGVALUE" == "DUMMY" ]; then
+					if [ "$CFGVALUE" == "DUMMY" ] || \
+						{ [ "$CFGVALUE" == "$NON" ] && \
+						{ [ "$CFGCAT" != "DXVK_LOG_LEVEL" ] && [ "$CFGCAT" != "DXVK_LOG_PATH" ] && [ "$CFGCAT" != "STL_PV_SHELL" ] ;};}
+					then
 						CFGVALUE=""
 					fi
 


frostworx:
thanks for the PR, but please test this change heavily, because the configurations returned from yad might break if there are empty values. it was very fragile on initial implementation and I simply want to avoid any future hassle for such a cosmetical change.

HanPrower:
Please don't respond until after your holiday. I'm not desperate for a response ^.^;;

No empty variables make it to yad, so that shouldn't be an issue.

This PR only effects saving the config to file. When the variables are loaded they still get set to "none" as usual.

I had a thought for a possibly cleaner alterative which may be to decouple settings that can be "none" and settings that should be <blank> by changing the blank ones to be read as "unset" instead.

Currently real settings such as "none" for DXVK logging are being mixed with visual indicators of whether a setting has been set or not.

If the two were separated it'd mean there'd be no "special" settings to take in to consideration as I had to currently in the PR. The only issue is there'd have to be a time of migration to get any "none"'s in files changed... :\

frostworx:
Same as with #521:

```
I'd say let's try it and see if someone complains :)
Thanks for the PR! +1
```

frostworx:
the PR apparently breaks at least https://github.com/frostworx/steamtinkerlaunch/issues/534
*reverting soon*

frostworx:
hm, just reverted it, but recognized that the game I used for testing doesn't like MangoHud by default. So maybe a non-issue.
needs more tests again... :(

HanPrower:
Could just add the relevant variable to: https://github.com/frostworx/steamtinkerlaunch/pull/522/files#diff-77c6d4bb1bf7c1988ff5a068c856b6c24e113663f2e61cb3a2c98723c0fafc04R8844

I do genuinely believe variables that can be `none` should be decoupled from the language use of `none`. It's mixing types.

But as I have no time to do it myself I'll just 🤷

**Edit** In regards to your recent comment. Oh xD

frostworx:
thanks for the quick reply.

> I have no time...

dito. I'm currently undecided if I should keep on working on the project generally.

frostworx:
hm, tried it again and it works fine for me. sorry for the noise.
re-adding it in a sec...

PR #556, opened 2022-08-26
Rough Polish update

Faalagorn:
I should probably revise it all, but as always, a rought update for Polish and some minor fixes that cought my eye


frostworx:
thx mate

PR #559, opened 2022-08-27
Rename 'Tags' to 'Collections' in Add Non-Steam Game Menu

sonic2kk:
The Add Non-Steam Game Menu had a few leftover mentions of 'Tags' rather than 'Collections'. This also changes "Exe" to say "Executable Path," which should hopefully make it clearer that this menu supports native Linux games.

image

This PR also updates the Steam Collection Selection submenu, removing the column name for the checkboxes and adds a space to the column heading for the Steam Collection:

image

Went through and updated the translations, avoiding touching any actual translated strings. Let me know if I missed something or accidentally changed something I shouldn't have. All feedback welcome :smile:


frostworx:
nice, thanks!

PR #561, opened 2022-08-27
Set Tags List for Non-Steam Game Tags Dropdown 

### LINKED ISSUE #557: Add Non-Steam Game Menu Doesn't List Tags Correctly
The "Tags" dropdown in the `steamtinkerlaunch ansg gui` menu won't properly separate tags into separate entries, instead it puts them all on one line (which actually gets cut off for me because there are so many). See below:

image

Maybe this is caused by too many tags? Haven't looked into the cause :sweat_smile: The "Tags" menu works correctly though:

image

Hadn't used this menu until earlier today so I never noticed this before.

sonic2kk:
Fixes #557 

This PR sets the Tags dropdown on the Add Non-Steam Game Menu to be a list of Steam tags/categories. Below is a screenshot of how it looks.

image

This change was made by changing how `VALIDTAGS` is set, replacing `tr '\n' ','` with `tr '\n' '!'`, which is done in other places where yad dropdown arguments like this are created.

I have a few concerns with this PR:
1. I'm not sure if I needed to change Line 18563, inside of the `if [[ -v SCATSEL[@] ]]; then`. I can revert this if it shouldn't be set this way
2. I'm not sure what the `sed` checks do after the pipe, or if these need to be changed in any way
3. There was what I presume to be an unwanted category showing up in the categories list, `rtDismissedReviseReviewReminder`. For Steam, `rt` seems to be how they start variables (see how they stored Steam Deck reservation timestamps). The dismissed revise review reminder could refer to when a user clicks "Maybe Later" when asked to write a review, see below. It's unlikely that a user would want to add a game to this category, so I added a loop to remove this category. It's in a list in case there are other unwanted categories we want to remove in future.

    image

    The problem with this approach that I've taken is that it doesn't resolve the bigger issue imo: This item is still returned in the list of tags such as on the submenu where the user can select multiple categories for their game. I'm not sure how better to fix the overarching problem so feedback here would be appreciated!

5. There is a blank entry at the top and bottom of the dropdown. A blank at the top is probably fine imo, but at the bottom might be slightly undesired. Not a huge deal but I might need some advice on how to fix it :slightly_smiling_face: 

I have a pending PR to rename Tags to Collections in this menu, I hope there are no conflicts :smile: 

All feedback is welcome. Thanks!

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v11.0"
+PROGVERS="v11.1.20220828"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
@@ -18478,7 +18478,7 @@ function SteamCatSelect {
 
 	setShowPic
 	unset VALTAGS
-	mapfile -d "\n" -t -O "${#VALTAGS[@]}" VALTAGS <<< "$(getActiveSteamCollections)"
+	mapfile -d "\n" -t -O "${#VALTAGS[@]}" VALTAGS <<< "$(getActiveSteamCollections | grep -v "^rt[A-Z]")"
 	SCATSELOUT="$(while read -r f; do if [[ ! "${SCATSEL[*]}" =~ $f ]]; then 	echo FALSE ; echo "$f"; else echo TRUE ; echo "$f" ;fi ; done <<< "$(printf "%s\n" "${VALTAGS[@]}")" | \
 	"$YAD" --f1-action="$F1ACTION" --image "$SHOWPIC" --image-on-top --window-icon="$STLICON" --center $WINDECO --list --checklist --column=Choose --column=SteamCollection --separator="\n" --print-column="2" \
 	--text="$(spanFont "$GUI_STEAMCATSEL" "H")" --title="$TITLE" --button="$BUT_SEL":0 --button="$BUT_CAN":2 "$GEOM")"
@@ -18520,6 +18520,12 @@ function AutoMarkSCat {
 	}
 }
 
+function filterUnwantedSteamCategories {
+	FILTEREDTAGS=$(echo "$1" | grep -v "^rt[A-Z]")
+
+	echo "$(printf "%s" "${FILTEREDTAGS[@]}" | tr '\n' '!' | sed 's/\!*$//g')"
+}
+
 function addNonSteamGameGui {
 	writelog "INFO" "${FUNCNAME[0]} - Starting the Gui for adding a $NSGA to Steam"
 
@@ -18550,12 +18556,12 @@ function addNonSteamGameGui {
 	NOSTGVR=0
 	unset VALTAGS
 	mapfile -d "\n" -t -O "${#VALTAGS[@]}" VALTAGS <<< "$(getActiveSteamCollections | sed '/^$/d')"
-	VALIDTAGS="$(printf "%s" "${VALTAGS[@]}" | tr '\n' ',' | sed 's/,*$//g')"
+	VALIDTAGS="$(filterUnwantedSteamCategories "$VALTAGS")"
 
 	AutoMarkSCat
 
 	if [[ -v SCATSEL[@] ]]; then
-		NOSTTAGS="$(printf "%s" "${SCATSEL[@]}" | tr '\n' ',' | sed 's/,*$//g')"
+		NOSTTAGS="$(filterUnwantedSteamCategories "${SCATSEL[@]}")"
 	fi 
 	export CURWIKI="$PPW/Add-Non-Steam-Game"
 	TITLE="${PROGNAME}-$NSGA"


frostworx:
thanks for the PR.
1. this should be fine because the initial `NOSTTAGS` variable (being set here) should be only used to draw the tags.
2. the sed removed a trailing ','.  I'd suggest you stdout that line and see if you need to possibly sed a trailing '!' instead
3. interesting. never seen those before. Probably safe enough, easier and more complete to use one generic filterUnwantedTags function which removed all tags starting with `rt`_(?)_
4. the cleanDropDown function should ideally remove those unwanted lines (haven't checked right now). So I guess a trailing '!' in above 2. (partially) causes this

sonic2kk:
> the sed removed a trailing ','. I'd suggest you stdout that line and see if you need to possibly sed a trailing '!' instead
> [...]
> the cleanDropDown function should ideally remove those unwanted lines (haven't checked right now). So I guess a trailing '!' in above 2. (partially) causes this

So the first blank line is probably caused because by default, the tags dropdown is empty. I'm pretty sure I've seen other dropdowns do this so it's probably fine. Plus having a blank option is good for users that don't want to add any tags.

The last blank line is only present when we filter out the `rtDismissedReviseReviewReminder`. Changing the `sed` command to `sed 's/\!*$//g'` (which should remove trailing `!`) fixes the incorrect blank line at the bottom of the categories list, where `rtDismissedReviseReviewReminder` is removed from. You were spot on, thank you! I've pushed that fix now.

<hr>

> interesting. never seen those before. Probably safe enough, easier and more complete to use one generic `filterUnwantedTags` function which removed all tags starting with rt(?)

I hadn't seen them either and it doesn't look to be documented anywhere I could see. I understand idea of removing all tags that start with `rt` but there could be a few complications here:
- A user might have an "RT" category, maybe for games with "raytracing" or something.
- A user might have "RT" as a category name, for example "RTS".
    - Visually, Steam Category names are all uppercase. There is no difference in the Steam UI with a category named "THQ Nordic" and "THQ NORDIC". They all show up as upper-case:
    
     image

Of course there could be a case where a user has an actual category in their library named `rtDismissedReviseReviewReminder`. But in my opinion, this would be _incredibly_ unlikely. Though it would be funny if anyone does and files a bug report about their category not being listed :laughing: I think having a hardcoded list of what we assume to be Steam's internal categories is probably the most straightforward. Plus it could be a bug on Steam's side that this is listed at all.

So is the idea to create that general `filterUnwantedTags` function to filter the tags for both the tags dropdown list *and* for the tags selection submenu where the user can select multiple tags? I think this is a good idea but I just want to confirm :smile:

frostworx:
thanks for the fix.
regarding `rt`:
I agree the user should have the option to also create a collection starting with "rt".
Imho it would be safe enough though, if our `filterUnwantedTags` function _(yeah, exactly - one function for both tags dropdown and submenu)_ would filter every tag starting with lowercase `rt` and having an upper-case letter as next char - so basically a
`grep -v "rt[A-Z]"`
Yeah I know, steam draws all collections uppercase - super annoying, as it is case-sensitive internally... :)

I don't know how many other `rt[A-Z]` internas exist or if others are planned, but with this approach it should be pretty future-proof.
I'm also fine with just filtering hardcoded `rtDismissedReviseReviewReminder`, but chances are higher that it requires future fixes/additions. 

hehe, yeah, creating a `rtDismissedReviseReviewReminder` collection is a good idea 🤓

sonic2kk:
I added the function you suggested and it filters the tags list dropdown. I ran into a slight problem with reusing that function on the submenu with the checkboxes, so for simplicity I just changed `mapfile -d "\n" -t -O "${#VALTAGS[@]}" VALTAGS <<< "$(getActiveSteamCollection)"` to `mapfile -d "\n" -t -O "${#VALTAGS[@]}" VALTAGS <<< "$(getActiveSteamCollections | grep -v "rt[A-Z]")"`.

The reason I ran into problems is that in the new `filterUnwantedSteamCategories` function I wanted to echo back the filtered list with the newlines replaced by `!` and the trailing `!` removed. This removed some duplication for the `NOSTTAGS` filtering, as it can also use the new tag filtering function. Changing the tag filtering function so that it can be used with `getActiveSteamCollections` is probably possible though. I'm happy with whichever way you'd prefer, but I think just adding the grep to `getActiveSteamCollections` is probably simpler.

Thanks :smile: 

EDIT: Just bumped the version to be `v11.0.20220830`, forgot to bump the version in my previous PRs :sweat: Hope this is okay!

frostworx:
thanks for the bump.
Looks good so far and I'd merge it, but I'd suggest to replace `"rt[A-Z]"` with `"^rt[A-Z]"` to only skip collections starting with that string and not also those which contain it
_(I know I suggested `"rt[A-Z]"` myself 😀 )_

edit: oh, and the PR conflicts

sonic2kk:
Oops! That was my bad with the regex, thanks! It should be fixed now :smile: 

The merge conflict was caused by the program version not matching :raised_eyebrow: I just reverted it back, I'm not sure why git was upset about the version being changed but it can be bumped later :-)

frostworx:
looks good! thanks for your work! 👍

frostworx:
fyi: 
`shellcheck` _(highly recommended)_ found some issues with your above PR.
I added fixes for it with [this commit](https://github.com/frostworx/steamtinkerlaunch/commit/beeb9632dac7636133896ddc60655e66fe1eefbd)
should be fine, but I haven't tested them.

Issue #530, opened 2022-08-02
[FEATURE] An option to prepend a command


Nano-Ocelot:
It would be nice to have the ability to place a command in front of the game command in the same way that you can in steam launch options like, `'examplecommand' %command%`. Specifically what I want to do is run the game like this `obs-vkcapture %command%` to start the game with obs-vkcapture capturing the game window and sending it to OBS. Maybe there already is an option like this but I can't seem to find it and I've been reading through the wiki and experimenting with options for hours.

frostworx:
parsing commands prepended in the steam launch option is no option, but I consider adding `obs-vkcapture` support.

Nano-Ocelot:
Excellent. It's a great tool to easily capture a game window for recording and streaming. It works on Wayland and X11 and is probably the most performant of all methods I've ever used. That would be a great feature.

Here's the link to it.
https://github.com/nowrep/obs-vkcapture

frostworx:
Thanks for the link. In fact I already started with the implementation yesterday. :)
I might have soon something to test, but it is not fully functional yet and there are some other issues open which I like to finish before.

Nano-Ocelot:
I'll be a guinea pig if you need any testing. Just let me know when it's ready to test.

frostworx:
The above push contains the initial code for using `obs-vkcapture` directly when launching a game.
The function can be enabled per game with the `USEOBSVKC` option.
I haven't really tested very much yet, but it doesn't seem to do anything useful for now. Will hopefully keep on working on it soon :)
Would be great if you would test (maybe too early in the current state, but maybe you find something out already).
It would be amazing, if you'd even add a little wiki when the function is implemented functionally :)

frostworx:
heh, just tested it _(with Bubble Bobble 4)_ and it just works :)

Maybe a good idea to automatically start obs _(directly with `OBS_USE_EGL=1 obs`)_
when it is not already running. What do you think?

Nano-Ocelot:
That's not a bad idea to just do it automatically. I was just going to add that to my 'game-start-script.' If you're going to do that it might be best to also run it like `OBS_USE_EGL=1 obs --minimize-to-tray`. 

Or maybe it's better to leave that all up to the user? Some people might also want it to start with replay buffer automatically at start. Or go straight to streaming or recording. For instance my personal plan is to start with replay buffer because I'm using it as a more performant version of replaysorcery or nvidia shadowplay.

[...]

PR #562, opened 2022-08-29
update obs-vkcapture to obs-gamecapture

zany130:
https://github.com/frostworx/steamtinkerlaunch/issues/530


frostworx:
what is this PR supposed to do? can't see any useful effect. Is it supposed to be complete.
more information please

sonic2kk:
I believe it's meant to replace `obs-vkcapture` with `obs-gamecapture`, which accomplishes the same thing as `obs-vkcapture` but can detect if a program is OpenGL or Vulkan and from there it can internally select the right backend to record with.

From #530:
> also as <user> also wanted OpenGL capture I tried replacing `obs-vkcapture` with `obs-gamecapture` which automatically selects the correct one depending if the game is OpenGL or Vulkan seems to work here I'll post a PR shortly.

frostworx:
I thought so, but the PR does nothing effectively. That's why I asked if it is supposed to be complete.

edit: I.e: the string "obs-gamecapture" does not occur even once in the PR.
And even if it did, I'd like to know where the program could be found.

zany130:
Wdym?

 I changed the internal variable `OBSVK`to `OBSCAP` as it no longer just Vulkan capture.

And i changed `obs-vkcapture` to `obs-gamecapture` so it can work with OpenGL and Vulkan 

I tested these changes in Witcher enhanced edition using WINED3D (so OpenGl) and DXVK (so Vulkan) obs was able to see and capture the game both times

frostworx:
> And i changed `obs-vkcapture` to `obs-gamecapture` so it can work with OpenGL and Vulkan

where?

and what is `obs-gamecapture` and where is its project page?

sonic2kk:
I believe this line changes it:

https://github.com/frostworx/steamtinkerlaunch/pull/562/files#diff-77c6d4bb1bf7c1988ff5a068c856b6c24e113663f2e61cb3a2c98723c0fafc04R59

And the project page is the same as `obs-vkcapture`, under the usage section: https://github.com/nowrep/obs-vkcapture#Usage

frostworx:
wtf! when I open the https://github.com/frostworx/steamtinkerlaunch/pull/562/commits
I only see 

besides the other correct variable renames.
thanks for clarification. Sorry, but of course, I wouldn't have asked otherwise 🤔

zany130:
Yeah in the first push i forgot to change the actual value of the variable to `obs-gamecapture` I immediately pushed a second commit to fix that and update the lang files

frostworx:
hah, excellent. thanks for the heads up and thanks for a great new feature! *merging* immediately! 😀

zany130:
NP sorry about the confusion anyway something I didn't do in this PR Is remove the check for if we are using wine and if so disable  `obs-gamecapture` 

here https://github.com/frostworx/steamtinkerlaunch/blob/de990bcf54c07780ac275d1d9558c1b45274b8ec/steamtinkerlaunch#L16684

Is this necessary ? I assumed the reason was because wine would use WINED3D (thus OpenGL) instead of DXVK (thus Vulkan) but that's not an issue with `obs-gamecapture` and the user might expect it to work only to realize it was disabled because they are using wine instead of proton

and secondly IDK if changing `OBSVKC` to `OBSCAP` might mess up existing configs with `OBSVK` at least from my testing it doesn't cause any problems

frostworx:
I don't think that the valid renaming of the variable has a deeper impact, because the function is pretty new and probaby only a few game configs of very few people are affected.
Can't remember why I added the wine check, but I guess it had some reason. If someone wants it removed, I wouldn't care as well though :)

frostworx:
Oh, you edited while replying: yes, might have been wine3d related indeed :D

zany130:
yeah sorry I have a bad habit of pressing submit or whatever in mid thought and then I'm like "oh should of clarified this" LOL I try to put EDIT whenever possible but If I always did that you would just see edit everywhere lol, so I just try to edit as fast as I can and hope no one notices 😅

anyway yeah I think that can be removed as a user might want to capture a game that is run with pure wine, before it might have prevented the game from working or maybe just silently failed but, the new `obs-gamecapture` should be able to handle this and automatically switch to OpenGL mode.

frostworx:
hehe, np 😀
[just removed](https://github.com/frostworx/steamtinkerlaunch/commit/286ce650e2cf3a1ceb6ec3749dd9b6dbb48daa39) the if-clause. thanks for suggesting 👍

PR #560, opened 2022-08-27
Fix Non-Steam Proton Games from Launching

sonic2kk:
Non-Steam Proton games wouldn't launch with SteamTinkerLaunch. It seems the cause was a missing `"${GAMESTARTCMD[*]}"` argument when setting `PROTONSTARTCMD`. I discovered this by inspecting the logs and noticing that `GAMESTARTCMD[*]` had the correct value, but `PROTONSTARTCMD` was missing it for some reason. I'm not totally sure how the `setProtonCmd` function works or what functions are called around it, so I'm not sure why this was only broken for Non-Steam Proton games, but this seems to fix Non-Steam Proton games in my tests and doesn't break regular Steam Proton games. In fact, games seem to launch significantly faster now? Could just be placebo but I did compare a build from this branch and a build from master... Who knows :-)

Non-Steam Proton games tested:
- HoloCure
- Sonic P-06

Steam Proton games tested to ensure this change didn't break anything:
- Cookie Clicker (regular ol' Cookie Clicker with no tweaks)
- Old School RuneScape (this has a custom command to launch a native Linux third-party launcher called RuneLite, and also has custom commands like GameMode enabled)
- Fallout: New Vegas (modded to the hilt with MO2, and has a bunch of STL tweaks like DXVK Framerate Limiting, MangoHud and GameMode)

Steam Native games tested to be extra sure that this change didn't break anything:
- Terraria (native Linux game with `steamtinkerlaunch %command%`, no tweaks)
- Celeste (native Linux game with `steamtinkerlaunch %command%`, MangoHud enabled and a custom command line option set to force the game to use Vulkan instead of OpenGL)

I believe I covered as many basis as reasonably necessary to ensure this doesn't break anything, but I welcome all feedback and of course if there's something this change can potentially break let me know.

Thanks!

PR #569, opened 2022-09-01
Only append GAMESTARTCMD to PROTNSTARTCMD if INGCMD is blank

### LINKED ISSUE #568: Sonic Adventure 2 fails to launch with SteamTinkerLaunch from git
Possibly unfortunately related to PR #560.

Sonic Adventure 2 fails to launch with SteamTinkerLaunch from the latest git. I tested v11.0 and it works fine, it's just using a release from git that fails to launch the game. I am sure this is not specific to SA2, but it's the only game I've come across so far that has this issue. This happens on both desktop and Steam Deck (where Steam Deck seems to download the latest Git anyway).

When launching, SA2 has a launcher and then the game starts. Other games with launchers, such as Fallout: New Vegas, work fine for me, though I launch that game via STL through MO2.

Looking at the log, this line looks to be of interest:

```
Thu Sep  1 21:38:30 BST 2022 INFO - launchSteamGame - Final outgoing start command: '/home/emma/.local/share/Steam/compatibilitytools.d/GE-Proton7-30/proton waitforexitandrun /run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe /run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe'
```

In the STARTDEBUG file, these lines are of particular interest as well:
```
Thu Sep  1 21:38:30 BST 2022 - Sonic Adventure 2 (213610) - ORGGCMD '/run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe'
...
Thu Sep  1 21:38:30 BST 2022 - Sonic Adventure 2 (213610) - FINALSTARTCMD '/home/emma/.local/share/Steam/compatibilitytools.d/GE-Proton7-30/proton waitforexitandrun /run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe /run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe'
```

It seems my fix in the PR was not quite right:
```bash
PROTSTARTCMD=("${RUNPROTON//\"/}" "$WFEAR" "${GAMESTARTCMD[*]}")
```

Maybe it would be better to do a check if the `GAMESTARTCMD` path is in `WFEAR` or something along those lines, I'm not too sure. Or maybe that fix is only needed for Non-Steam Proton games, so if we have a way to check for that, we could only apply that line for Non-Steam Proton games. Why this doesn't affect *all* games is strange to me.

Not sure  what the best solution is. If there's a way to know if a game is a Non-Steam game, we should be able to just do something like this as a band-aid fix:

```bash
PROTSTARTCMD=("${RUNPROTON//\"/}" "$WFEAR")
if [ "$NOSTGAM" -eq 0 ]; then
  PROTSTARTCMD=("${RUNPROTON//\"/}" "$WFEAR" "${GAMESTARTCMD[*]}")
fi
```

The overarching solution might be to fix `$WFEAR` sometimes being empty, e.g. for Non-Steam games pretty much exclusively it looks like. But if there's a way to know that we're dealing with a Non-Steam game, the above proposed solution could work for the meantime. 

EDIT: Tried to attach a log and GitHub says it can't process the file. IOU for a log, I'll try again later :-)

sonic2kk:
Fixes #568.

Sonic Adventure 2 was failing to launch as a result of the changes in #560. The launch command ended up having the path to the game's launcher twice:

```
Thu Sep  1 21:38:30 BST 2022 INFO - launchSteamGame - Final outgoing start command: '/home/emma/.local/share/Steam/compatibilitytools.d/GE-Proton7-30/proton waitforexitandrun /run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe /run/media/emma/onigiri/Games/steamapps/common/Sonic Adventure 2/Launcher.exe'
```

To fix this, I changed the logic from my PR #560 to only append `GAMESTARTCMD[*]` if `INGCMD[*]` is blank. For regular Steam games, `INGCMD` seems to always be set. However for Non-Steam games it does not ever seem to be set.

Example log line for Non-Steam Game HoloCure:

```
Thu Sep  1 23:15:53 BST 2022 INFO - launchSteamGame - Original incoming start command: ''
```

I guess however the path for Non-Steam games is sent is not checked by STL. A better fix might be to, somehow, properly parse the INGCMD for Non-Steam Games, but I'm not sure how to do this. I'm not even sure if I'm on the right track for it. What confuses me is why only some games (like SA2) have a duplicated path for their final outgoing command, even though all Steam games I have tested do actually have the correct incoming start command. Maybe it's correctly unset somewhere but not for Sonic Adventure 2, I'm not sure.

As this could be a potentially bigger issue I got a PR up as fast as I could to get some discussion and feedback. It's totally fine if this doesn't get merged as it is a bit of a "hacky" fix. Investigating how to fix the broader problem might take some time though, and this could serve as a fix in the interim. At least that's my logic for rushing to get this PR up. It's also my fault that this broke in the first place, so many apologies about that. I'll try to do better in future.

So essentially, I wanted to get a fix out for this issue to discuss the merits of merging a fix like this, and to maybe dig deeper into the root cause, because I could be way off. It also patches up the bug brought in by my "fix" without having to revert it. I believe on Steam Deck STL will update to the latest Git automatically, so there is a potential for this bug to affect a sizeable amount of users, and I wanted to get it at least temporarily patched ASAP.

If it's preferable to not have these kinds of fixes, or if it's not too hard to fix the root cause of the bug, we can toss this PR aside :slightly_smiling_face:

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -16768,7 +16768,12 @@ function setProtonCmd {
 			
 			if [ -f "${RUNPROTON//\"/}" ]; then
 				writelog "INFO" "${FUNCNAME[0]} - Prepending Proton '$USEPROTON' (='${RUNPROTON//\"/}') to the command line '${GAMESTARTCMD[*]}'"
-				PROTSTARTCMD=("${RUNPROTON//\"/}" "$WFEAR" "${GAMESTARTCMD[*]}")
+
+				PROTSTARTCMD=("${RUNPROTON//\"/}" "$WFEAR")
+				if [ -z "${INGCMD[*]}" ]; then
+					writelog "INFO" "${FUNCNAME[0]} - INGCMD is blank (may be a Non-Steam Game), so appending GAMESTARTCMD to Proton start command"
+					PROTSTARTCMD=("${RUNPROTON//\"/}" "$WFEAR" "${GAMESTARTCMD[*]}")
+				fi
 			else
 				writelog "INFO" "${FUNCNAME[0]} - Still don't have a usable proton executable in RUNPROTON '{RUNPROTON//\"/}')"
 				if [ "$HAVEINPROTON" -eq 1 ]; then


frostworx:
I haven't digged through the new ton of issue mails yet in detail, but I'm sure more tham only #568 is caused by #560.
So "your test results look fine for me. I'm still not sure if this is the perfect place to add it, but I don't find the time to look for a better place, so I'll merge your PR and we'll just wait if someone cries." went realy quickly and someone cries 😀
Might be a better approach to revert #560 and find its correct position instead of patching another `INGCMD` condition around it, which might or might not be the only condition required to get the function working correctly at the current position.

sonic2kk:
That's fine, we can revert #560 and debug. I made this PR because for the meantime I wanted to have some kind of patch, but I agree with your point that this could only be patching one specific instance.

I think reverting the Non-Steam Game fix is best for now.

PR #572, opened 2022-09-03
fixes https://github.com/frostworx/steamtinkerlaunch/issues/567

zany130:
Move down the `OBSCAP` if to below game scope.  

TLDR,
**most** tools need to go after game scope's `--`  since the code is run top-down, by moving the code to append `obs-gamecapture` below the equivalent  game scope code, we can ensure that `obs-gamecapture` will be added after game scopes `--`


frostworx:
yay, nice! thanks for the PR 👍 *merging*

Issue #575, opened 2022-09-04
gamescope with mangoapp + games with backslash in launch command


gu1ll0me:
Pyre game have backslash in launch parameters (`..\`) :

```
/home/user/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=462770 -- /games/steam/steamapps/common/SteamLinuxRuntime_soldier/_v2-entry-point --verb=waitforexitandrun -- /games/steam/steamapps/common/Proton 7.0/proton waitforexitandrun /games/steam/steamapps/common/Pyre/x64/Pyre.exe /c ..\ /PerfDashWarnings=false /KeepAllSaves=false
```

When enabling mangoapp in game menu, STL [adds double-quote for each launch parameters](https://github.com/frostworx/steamtinkerlaunch/blob/master/steamtinkerlaunch#L15518), but the backslash escape the double-quote and the game refuse to launch.

Adding `sed -i "s/\\\/\\\\\\\/" $STLSHM/maprun.sh` after line 15518 solve the problem by doubling the backslash.

sonic2kk:
Nice catch! If you would like you could test this some more with other games, just to ensure it doesn't break anything, and if/when you're confident, you could open a pull request! :smile: (That way you can get direct credit too :wink:)

I had a look at where you said the fix can go and that all looks sane to me. But if you could test with other games, we'd welcome a PR!

frostworx:
Thanks, good find and fix!
I fully agree with @sonic2kk: A few tests if the patch has any negative impacts _(though unlikely)_ and a proper PR would excellent.

gu1ll0me:
I'm not using gamescope+mangoapp on a daily basis because mangoapp doesn't close properly after the game exit, but I will install some more games this week and make a few tests. I will create a PR if there is no problem.

frostworx:
thanks for testing/contributing! take your time!
I don't use mangoapp, but was believing that it is closed _(killed)_ correctly on game exit _(upstream kills it as well iirc)_
does the [Kill Switch](https://github.com/frostworx/steamtinkerlaunch/wiki/Proton-KillSwitch) work at least?

gu1ll0me:
I can't test the kill switch because there is no `KillSwitch` file or folder in `/dev/shm/steamtinkerlaunch/` when I launch a Proton game. To kill `mangoapp` process, I have to use `killall mangoapp` or hit Steam stop button.

frostworx:
[...]

PR #581, opened 2022-09-14
mangoapp - escaping backslash to preserve double-quote

gu1ll0me:


gu1ll0me:
Solve [#575](https://github.com/frostworx/steamtinkerlaunch/issues/575)

frostworx:
thx 👍

PR #600, opened 2022-09-23
Improve existing issue templates

sonic2kk:
Hey, so first off I'm *not really 100% sure* if I'm doing this right, GitHub makes collaborating on this sort of thing a bit challenging :sweat_smile: I wanted to open this PR to show off the ideas I had for this topic. I feel like showing the changes directly is better than simply opening an issue and describing it.

So the current issue template when creating issues is, for some reason ignored by quite a few users. As SteamTinkerLaunch gains popularity and users more and more people will report issues, and to help them help us I think it would be a good idea to make use of the Issue Templates feature that GitHub offers. Valve's [Proton repository](https://github.com/ValveSoftware/Proton) uses this feature and it makes filling out required information a little easier.

When a user tries to open a new issue they'll be prompted to select the kind of issue they want to create: Bug Report, or Feature Request. These have two slightly varied templates, and also have default tags with "bug" and "enhancement" respectively :smile: 

<hr>

Bug Report issues prompt the user for the information they should fill out, with some comments (yes, actual HTML comments which get hidden with markdown!) on what should be filled out for those fields. The hope here is that we help users know what information they should fill out, encourage users to fill out the information by giving them these prompts, and to some extent help discourage low-effort issues that boil down to "stl broken pls fix", with little to no helpful information given.

The Feature Request template is similar to the Bug Report one but only asks for the STL version and distribution, which can give us an idea of why that feature may be desired on that specific distribution. It also does not explicitly ask for a log as this may not be relevant.

<hr>

I opted to keep things simple and just have "Bug Report" and "Feature Request". I did think about breaking it down further and having categories for Flatpak, Steam Deck etc, but I decided against this in the end. The reason being is that I wanted these issue templates to exist to help users fill out good quality issues, but I didn't want the layout of the issue to have too many strict rules. The "flow" of how a user lays out the issue can be left up to them, as long as we have the information - Probably the most often requested piece of information that users leave out is the dang log! I am guilty of this myself, so having it explicitly requested in the issue template can help remind users to attach the log.

Let's have a discussion about if it's worth implementing further issue templates at all, and if so how best to format them. I did take care creating the markdown files but I'm always open to feedback. As I said in the beginning, I'm not even sure if I did this right, so we might not be able to merge this PR. But if we end up with a good resolution feel free to just copy and paste the markdown into the GitHub issue template settings manually.

Thanks!


frostworx:
**YAY, excellent PR! Thank you so much!** 🎉
I wanted to add it long before, but was afraid my bad English could make it more complicated than helpful 😏

sonic2kk:
Woohoo, glad I could help! 😃 

I couldn't really "test" this before opening the PR because github acts strangely when you try to do this on a branch, but I just tested by trying to open a new issue and they showed up! 🎉  There's also an option for a blank issue which I guess is fine in case someone opens an issue that doesn't fit the other categories, but I would hope most users take advantage of the bug report template.

sonic2kk:
Actually, that extra option button seems to be a GitHub feature. On Desktop it looks like this:

image

On the mobile app it's a separate option in the list. So nothing was configured wrong :sweat_smile:

PR #601, opened 2022-09-24
Add 'Valheim' to MO2 games list

sonic2kk:
On March 20th, Valheim was added to the list of MO2 wiki under the list of games with basic support: https://github.com/ModOrganizer2/modorganizer/wiki/Game-Support/_history

This file was discussed recently in an issue I think, about how it should be updated periodically to include new games. I'm not *totally* sure if anything else is needed for this game to be """supported""" so to speak, so please let me know if I have to do any additional work.


frostworx:
Thank you! Of course I'm not happy with this helper file, ideally there would be an external list _(upstream)_.
But until the community takes part to maintain it, it is also no bigger problem.

PR #606, opened 2022-09-24
Add some extra badges to README

sonic2kk:
This is just a slight "touch up" to the Readme to add some extra badges from Shields.io. It adds:
- GitHub version
- Code size
- AUR votes

I did look into adding more, namely downloads, but it seems like that doesn't work.

A fairly inconsequential change but no worries if this is undesirable for any reason. On the other hand if there are any further badges we could add let me know and I can add them :) I did go through Shields quite extensively to see which ones we could add, these were the main ones I could find that worked.

EDIT: Tried to fix a slight formatting problem on the Readme on GitHub about MO2 (looks fine in my markdown previewer but not on GitHub). Caused a merge conflict, it looks like I can resolve it but it says "Commit Merge" and I don't want to press that in case it merges this PR, so I'll resolve it after review :smiley: I manually pasted in the markdown fix, this branch was created a while back so it didn't have this change.


frostworx:
Thank you!
_(hehe, I already wrote a reply, then edited the MO2 README _(it was only meant to be temporarily there anyway)_
and then my reply was gone 🙄)_
I don't know which else badges to exist, but if you find interesting ones it was nice, of course if you'd also like to add them.

PR #607, opened 2022-09-24
Fix getIDFromTitle

### LINKED ISSUE #605: Some Proton games do not return correct AppID when using getIDFromTitle
See also some discussion in #550

`getIDFromTitle` fails for some Proton games (should always fail for native Linux games currently if they've never been ran with Proton before, see #604). For some reason, with some  It seems like this happens when the order of `GAMEID`, `GAMENAME` and `GAMEARCH` in the game meta file is "incorrect" or not what STL is expecting at least.

I tested three games: NieR:Automata, Oblivion and Yakuza 0. Only the last one returned a correct AppID, Automata and Oblivion both returned their `GAMEARCH`.  When I modified the meta config for Automata and Oblivion to have the following order:

```
GAMEID=
GAMENAME=
GAMEARCH=
```

They started to return the correct AppID. It didn't matter where they were in the file, as long as they were together and in this order. This is how Yakuza 0 ordered the variables, so I decided to mirror how that config did it and that's how I discovered the fix.

I wondered how this would work for a game that only has numbers for its name like [21](https://store.steampowered.com/app/938520/Twenty_One/) (store page says "Twenty One", game library says "21"), so I tested it. STL didn't even write the `GAMENAME` for this game (not a big deal since [it doesn't work with Proton](https://github.com/ValveSoftware/Proton/issues/5942)), so I wrote it in myself as just `GAMENAME="21"`. Running `steamtinkerlaunch getid "21"` worked perfectly fine though, again with the order `GAMEID`, `GAMENAME` and `GAMEARCH`.

So it seems like STL is failing here for these games because those 3 variables have to be in a specific order. They can be anywhere in the file it seems like, as long as they are together.

EDIT: Semi-related, it would be nice if we could also search for games based on their `NAME` as well as `GAMENAME`, for example being able to search `DOOM Eternal` instead of `DOOMEternal` and having the ID come back. Maybe this is next part is wishlisty but it would be very cool if this was case insensitive, so `doom eternal` also worked, but not anything too important :smiley:

sonic2kk:
Fixes #605 

Currently, `getIDFromTitle` has some strange behaviour where it can return the wrong value for the game's AppID or miss it entirely! I'm not too well-versed but as far as I can tell, the current code tries to match the AppID from the file.

This PR takes a different approach for this function. It tries to recursively grep each file in `$GEMETA` (which is the `~/.config/steamtinkerlaunch/meta/id/general` folder I believe) and get the `NAME` field from it. If it can't find any file with the `NAME` field, it falls back to `GAMENAME`. This is also good if the user only types in, say, "Fallout New Vegas" and not "Fallout: New Vegas" (with the colon). All of these name checks are **case-insensitive**.

Like before, you can enter a partial name, so the user doesn't have to enter the full name - but if they do, the fallback mentioned above should catch some mistakes they make if they don't get the name exact. As a result of being able to enter partial names, I noticed early on that multiple games could be returned if a user entered something like "Sonic" or "Elder Scrolls". So I made the function return the AppIDs for all matching games

Here is some example output:

image

The `grep` search output looks something like `/path/to/steamtinkerlaunch/meta/title/general/appid.conf:NAME=blahblah`, and each match is on a newline. So we split the string by the newline and turn it into an array of strings by temporarily changing `IFS`. Once we have the array of strings we loop through the array and use `awk` to split the AppID from the *config file name* and actual game name, stripping out the quotes from the game name and using `xargs` to strip any trailing whitespaces (Sonic Adventure 2 had this for me so I wanted to fix it, it looks little nicer cosmetically).

All of this aims to help make the function more reliable :smiley: Feel free to test and make sure I didn't just get lucky with the games I tested. And of course, all feedback is welcome :smile:

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6703,7 +6703,29 @@ function getIDFromTitle {
 	if [ -z "$1" ]; then
 		echo "A Game Title (part of it might be enough) is required as argument"
 	else
-		grep -Rhi -B1 "^GAMENAME=.*.$1" "$GEMETA" | head -n1 | grep -o -E '[0-9]+'
+		# Search for all meta files with a partial match for their `NAME` field
+		# Search on `NAME` primarily, fallback to `GAMENAME`
+		SAVEIFS=$IFS
+		IFS=$'\n'
+		MATCHES=( $(grep -i "^NAME=.*.$1" "$GEMETA" -R) )
+		if [ -z "${MATCHES}" ]; then
+			MATCHES=( $(grep "^GAMENAME=.*.$1" "$GEMETA" -R) )
+		fi
+		IFS=$SAVEIFS
+		
+		# Echo all matches in format "AppID (Game Name)"
+		if [ ! -z "${MATCHES}"  ]; then
+			for i in "${MATCHES[@]}"
+			do
+				MATCH_AID="$( echo "${i##*/}" | awk -F'.conf:' '{print $1}' )"
+				MATCH_NAME="$( echo "${i##*/}" | awk -F'NAME="' '{print $2}' )"
+				MATCH_NAME="$( echo ${MATCH_NAME::-1} | xargs )"
+
+				printf "$MATCH_AID\t\t($MATCH_NAME)\n"
+			done
+		else
+			echo "Could not find AppID for name '$1'."
+		fi
 	fi	
 }
 


frostworx:
Thanks. Looks good for me. This increases the change that the function is actually involved more in the future as well 😀

frostworx:
sorry, I merged this too quick without a proper review.
just noticed, that you introduced multiple glitches.
Would be nice if you'd fixed those when you find the time:

```
$ shellcheck /usr/bin/steamtinkerlaunch 

In /usr/bin/steamtinkerlaunch line 6710:
		MATCHES=( $(grep -i "^NAME=.*.$1" "$GEMETA" -R) )
                          ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In /usr/bin/steamtinkerlaunch line 6711:
		if [ -z "${MATCHES}" ]; then
                         ^--------^ SC2128 (warning): Expanding an array without an index only gives the first element.


In /usr/bin/steamtinkerlaunch line 6712:
			MATCHES=( $(grep "^GAMENAME=.*.$1" "$GEMETA" -R) )
                                  ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In /usr/bin/steamtinkerlaunch line 6717:
		if [ ! -z "${MATCHES}"  ]; then
                     ^-- SC2236 (style): Use -n instead of ! -z.
                           ^--------^ SC2128 (warning): Expanding an array without an index only gives the first element.


In /usr/bin/steamtinkerlaunch line 6722:
				MATCH_NAME="$( echo ${MATCH_NAME::-1} | xargs )"
                                                    ^---------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
				MATCH_NAME="$( echo "${MATCH_NAME::-1}" | xargs )"


In /usr/bin/steamtinkerlaunch line 6724:
				printf "$MATCH_AID\t\t($MATCH_NAME)\n"
                                       ^-----------------------------^ SC2059 (info): Don't use variables in the printf format string. Use printf '..%s..' "$foo".

For more information:
  https://www.shellcheck.net/wiki/SC2128 -- Expanding an array without an ind...
  https://www.shellcheck.net/wiki/SC2207 -- Prefer mapfile or read -a to spli...
  https://www.shellcheck.net/wiki/SC2059 -- Don't use variables in the printf...
```

I suggest running `shellcheck` on your own for better highlighing :)

sonic2kk:
Ah! I just seen this. Yes, my apologies. I'll get a PR up to fix this ASAP (you mentioned shellcheck before and I meant to install it, I completely forgot until now!)

Issue #483, opened 2022-06-05
Uninstaller?


Apacelus:
Is there an uninstaller script somewhere?

frostworx:
Distributions which maintain a package, of course, can simply use their package management, but no, there is no uninstall option in the Makefile currently.

sonic2kk:
Not all Makefiles (very few in my experience) have an `uninstall` block. However, would a manual uninstall process work? I imagine something like this is what would go in a Makefile `uninstall` block but I have no experience with those :sweat_smile: 

```sh
sudo rm -rf /usr/bin/steamtinkerlaunch  # Remove STL program
sudo rm -rf /usr/share/steamtinkerlaunch/*  # Remove STL files (translations etc)
sudo rm -rf /usr/share/doc/steamtinkerlaunch/*  # Remove STL readme
sudo rm -rf /usr/share/applications/steamtinkerlaunch.desktop  # Remove STL .desktop file (shortcut in application launchers e.g. GNOME Dashboard or KDE Kickoff)
sudo rm -rf /usr/share/icons/hicolor/scalable/apps/steamtinkerlaunch.svg  # Removes STL icon
sudo rm -rf /dev/shm/steamtinkerlaunch/*  # Remove remaining STL files, almost certainly unnecessary but can make for a clean uninstall/reinstall scenario :)
```

I got these paths from following the installation steps in the [Makefile](https://github.com/frostworx/steamtinkerlaunch/blob/master/Makefile#L9-L18) and verifying them on my own system, but I recommend checking the exact paths and writing the commands out yourself to make sure you don't mess anything up - Escalated remove commands can be scary! I also didn't test this myself but this should be the gist of uninstalling (I hope).

frostworx:
basically you're right, but I'd prefer to not "hardcode" `sudo`, as the users should know what they are doing and what the makefile does as well. uninstall blocks, **if** they even exist as you already pointed out, often do not work correctly as nobody, including the authors do not really maintain them. The reason is probably that they are usually not required, because un-installing packages system-conform is mainly the job of package management packages.
Installing any program manually IMHO always should be the last resort. It can be dangerous, files might be left behind in the system, updating the program can be complicated and so on. So contributing and maintaining packages to the own distribution
always is the better choice and helps the community as well apparently.

frostworx:
Closing here, but you're invited to reopen any time if you think this needs further discussion.

PR #608, opened 2022-09-24
Add uninstall to Makefile

sonic2kk:
This PR attempts to add some very basic "uninstall" steps to the Makefile to allow for running `sudo make uninstall`. It essentially does the reverse of the install process to accomplish this.

I haven't ever touched a Makefile before this but I've had this idea for a while, and I had a look around online to see if I could find any examples of how to do this "right". Turns out I couldn't find very much at all on this! So my apologies if this is incorrect.

I initially had the idea for this PR all the way back when #483. After the comments there I weighed up in my mind whether or not this would be beneficial and worth investigating. After some research it seemed like this was the correct approach, but in-between various other things I kept this branch around and still weighed up 1) is this the right approach, and 2) is it worth implementing. Given the comments, I understand this PR might have a lower chance of being accepted.

I agree with everything laid out in the response, that installing from a Makefile should be a last resort. I also entirely think that uninstallation is *one of the many reasons we have package managers*, they do this job for us. I also agree that maintenance cost of uninstall blocks is something to take into consideration, which I tried my best to do for this PR.

The approach I took with the uninstall steps was to mirror the install steps (removing files first that are added by `install` last), and to avoid specifying removing *specific* files. This way we just remove the folders. This means we don't have to specify deleting, say, the `misc` folder in `/usr/share/steamtinkerlaunch`, we just remove *everything* under `/usr/share/steamtinkerlaunch`. I also took the consideration of hardcoding sudo on board, the Makefile does not enforce sudo itself though likely most systems will need `sudo make uninstall`.

My reason for going ahead and opening the PR anyway is that even if installing from a Makefile should be a last resort, I still think this would be a nice feature to have. I also think the maintenance cost for this should be relatively low. My hope is that this approach to the uninstall section is more maintainable and should rarely need updating, unless SteamTinkerLaunch writes to a new system folder with install. But if SteamTinkerLaunch ever adds a new folder to `/usr/share/steamtinkerlaunch` or something like that, no changes should have to be made to the uninstall block. From looking around I'm not sure which new system folders STL would ever need to write to, it would probably just add extra items into the `/usr/share` folder.

It might also be useful to have this uninstall block for users that want to "migrate" how they installed SteamTinkerLaunch. For example users on a distro without an STL package and that installed manually, but that want to migrate to Flatpak Steam, or a local install through ProtonUp-Qt.

I have a lot of stuff I do locally with STL just to mess around and learn, and this was an experiment to learn about uninstall blocks. I totally get it if the maintenance cost of this change is still too great or if it's unwanted for any other reason. Since it started as just a little "learning exercise" so to speak, I still think I got something out of working on this. And any feedback on how to add uninstall steps to Makefiles would be great if this is wrong or could be improved.

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,12 @@ install:
 	install -Dm644 "misc/steamtinkerlaunch.desktop" -t "$(PREFIX)/share/applications"
 	install -Dm644 "misc/steamtinkerlaunch.svg" -t "$(PREFIX)/share/icons/hicolor/scalable/apps"
 
+uninstall:
+	rm -rf "${PREFIX}/share/icons/hicolor/scalable/apps/misc/steamtinkerlaunch.svg"
+	rm -rf "${PREFIX}/share/applications/misc/steamtinkerlaunch.desktop"
+	rm -rf "${PREFIX}/share/doc/steamtinkerlaunch"
+
+	rm -rf "${PREFIX}/share/steamtinkerlaunch"
+
+	rm -f "${PREFIX}/bin/steamtinkerlaunch"
+


frostworx:
Thanks mate! I see no reason why this should not be merged.

Issue #604, opened 2022-09-24
SteamTinkerLaunch doesn't add `GAMEID` for native Linux games' meta files


sonic2kk:
Partially related to discussion in #550 

When using SteamTinkerLaunch as a launch option for native Linux games, it seems like the file in `$GEMETA` (`~/.config/steamtinkerlaunch/meta/id/general/<app_id>.conf`) does not get a `GAMEID` written out to it. From investigating the files I have it seems like Proton games always have `GAMEID` but native Linux games do not.

This is namely a problem for using the `getIDFromTitle` function, which is called when running `steamtinkerlaunch getid "Game Name"`. Native games seem to always fail (Portal, Portal 2, Factorio) *unless* they have been ran with Proton before (I ran Half-Life with Portal before and it seems to have this field). Proton games *always* seem to have this field, and usually  `getIDFromTitle` works correctly for these games.

Also, these meta files for native Linux games do not get symlinked to the `title` folder.

I had a look around to see where and how this file is generated to see if I could fix it but I couldn't figure it out (that's where all my PRs came from, I was fiddling around in this part of the code :sweat_smile:)

Below is a log from running native Portal, where I commented out the `400.conf` meta file so STL would re-generate it. From looking at it, it does not even complain that `GAMEID` is missing.

[steamtinkerlaunch.log](https://github.com/frostworx/steamtinkerlaunch/files/9637815/steamtinkerlaunch.log)

frostworx:
Sorry, still drinking my first coffee and I'm a bit lazy at the moment 😀
[via](https://github.com/frostworx/steamtinkerlaunch/issues/605#issuecomment-1256895858):

`getIDFromTitle` is just a fragile little helper one liner which is barely used in the "active" code
_(basically [my answer over here](https://github.com/frostworx/steamtinkerlaunch/pull/603#issuecomment-1256890258) matches here as well 😀)_
You're right, of course it could be improved a lot, but I'm uncertain if it is even worth the effort.
It would be much easier to simply remove the litle function and nost many people would miss it 😀

sonic2kk:
Now that the issue with the function is resolved I might try taking a look at fixing this. STL gets the AppID correctly for native games so it should just be a matter of writing it out to the meta file. I'll have a look at the logs to and figure out why it writes for proton games but not native games.

[...]

PR #609, opened 2022-09-27
Write out GAMEID to meta file for native Linux games

sonic2kk:
Resolves the initial issue for #604 

When I first opened the above issue, the reason was that the old code for getting a game's AppID relied (kinda) on the `GAMEID` field in the game's meta file. #607 changed this, so the original problem was resolved because now native Linux games are found. However the original "problem" I guess you could say still remained, `GAMEID` wasn't written out.

It doesn't seem like this is causing any problems but for parity I thought it would be good to write out the `GAMEID` for native Linux games as well. That's what this PR achieves.

The reason it wasn't being written out before is because before writing to the meta file, STL was only checking if `"$ISGAME" -eq 2`, where `2` is a Proton game. I added a check for `"$ISGAME" -eq 3`, as `3` seems to correspond to native games.

I tested a couple of native games (Portal, Factorio) and they worked. I can't see why this would potentially cause any problems and in my tests it works fine even if it's only an incredibly minor change.

Thanks! :smile:

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -19145,7 +19145,7 @@ function closeSTL {
 	rm "$STLSHM/lola-*.txt" 2>/dev/null
 
 	# only useful for win games or games started outside steam:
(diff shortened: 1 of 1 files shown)


frostworx:
thanks! 😀

PR #610, opened 2022-09-27
Fix shellcheck warnings in #607

sonic2kk:
Fixes all warnings/style messages of `shellcheck steamtinkerlaunch`. The biggest change is that `mapfile` is used to generate the array (removes the need for the `IFS` stuff). `[*]` is also used in the if blocks to expand the `MATCHES` array.

I also noticed a small issue which I fixed in this PR: I forgot to use `grep -i` when checking on `GAMENAME`, which meant checks on that var weren't case-insensitive. That has been fixed.

I did some pretty extensive testing to try and make *absolutely* sure that I didn't inadvertently break anything while trying to fix these warnings. Everything seems to work just as it did before.

I'll be sure to run `shellcheck` in future before submitting PRs.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6705,23 +6705,20 @@ function getIDFromTitle {
 	else
 		# Search for all meta files with a partial match for their `NAME` field
 		# Search on `NAME` primarily, fallback to `GAMENAME`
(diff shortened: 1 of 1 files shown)


frostworx:
thank you so much! ❤

Issue #613, opened 2022-10-02
Unable to run custom command


mmbossoni:
## System Information
- SteamTinkerLaunch version: 11.11
- Distribution: Arch Linux
- Installation Method: aur

## Issue Description
I've set a custom command to run with Hades game, but it isn't working. It used to work before, but from logs it seems it is not finding my proton version or something
I use steamtinkerlaunch by setting it as launch options on steam, not as a compat tool.

## Logs
[ProtonCSV.txt](https://github.com/frostworx/steamtinkerlaunch/files/9693626/ProtonCSV.txt)

[steamtinkerlaunch.log](https://github.com/frostworx/steamtinkerlaunch/files/9693622/steamtinkerlaunch.log)

sonic2kk:
>  I use steamtinkerlaunch by setting it as launch options on steam, not as a compat tool.

Using SteamTinkerLaunch as a launch option is not supported for Proton. You should be using SteamTinkerLaunch as a compatibility tool. The launch option is meant to be used for native Linux games, which to my knowledge Hades is not. 

> from logs it seems it is not finding my proton version or something

I agree with you, check which version of Proton the game is using in the Game Menu, whilst also making sure you are using SteamTinkerLaunch as a compatibility tool. If all looks correct, try deleting `/dev/shm/steamtinkerlaunch` and running your game again.

As for why this was working previously, I believe the launch command has changed. Looking at this part of your log:

```
dom 02 out 2022 18:21:39 -03 INFO - prepareLaunch - CheckCustomLaunch:
dom 02 out 2022 18:21:39 -03 INFO - checkCustomLaunch - USECUSTOMCMD is set to '1' - trying to start custom program '/home/marcelo/HadesTrainer.exe'
dom 02 out 2022 18:21:39 -03 INFO - checkCustomLaunch - FORK_CUSTOMCMD is set to 0 - starting the custom program regularly
dom 02 out 2022 18:21:39 -03 INFO - launchCustomProg - '/home/marcelo/HadesTrainer.exe' not found - searching in gamedir
dom 02 out 2022 18:21:39 -03 INFO - launchCustomProg - '/home/marcelo/HadesTrainer.exe' also not in '/mnt/games/marcelo/steam/steamapps/common/Hades//home/marcelo/HadesTrainer.exe' - checking if absolute path was provided
[...]

PR #614, opened 2022-10-03
Make usage of SteamTinkerLaunch clearer in Readme

sonic2kk:
It was not clear enough that SteamTinkerLaunch should be used as a launch option only for native Linux games. In the past it was probably okay to use the launch option for Proton games but lately there were more users reporting issues. This updates the Readme to make it clearer that the launch option is really only intended for native Linux games, and that the compatibility tool is the intended way of using SteamTinkerLaunch with Proton games.

Also removes the incorrect note that the launch option is unavailable on Flatpak; I asked and it is actually available (flathub/com.valvesoftware.Steam.Utility.steamtinkerlaunch#40).

See also the comments in #613.


frostworx:
Thanks, I'm sure some will still ignore it, but helpful for others for sure. Thanks!

PR #615, opened 2022-10-03
Allow tray icon in Steam Deck Desktop Mode

sonic2kk:
The Steam Deck should be able to use the tray icon in Desktop Mode, so only disable it on Steam Deck Game Mode.

It might not be very likely that a user would use STL in Desktop Mode and need the tray icon but since we have this variable now imo it makes sense to start enabling some STL functionality like this when we're in Desktop Mode. I have a few other branches where I'm experimenting with similar ideas, I'll open some PRs after more testing :smile:

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -4529,8 +4529,8 @@ function openTrayIcon {
 		loadCfg "$STLGAMECFG"
 	fi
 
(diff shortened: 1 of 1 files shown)


frostworx:
makes sense of course as well :) thanks!

PR #621, opened 2022-10-04
Remove unneeded deps files on Steam Deck

sonic2kk:
Removes the cabextract and innoextract archives, and the dot files extracted with their archives. These files are tiny, totalling just over 250kb, but it helps keep the deps folder clean :slightly_smiling_face: 

I opted to remove innoextract and cabextract by name instead of doing an `rm "*.zst"` because the archives might not always end with a file extension. I did think about maybe removing anything that isn't a folder from the STL deps folder but I thought it was best to have the removal of the archives be very explicit, even if it means duplicating their `rm`. That way if we ever did, for any reason, need to keep the archives around, we can do so.

The dotfiles I grouped all under a single `rm` because there's very likely no reason for a dotfile to ever be used in this folder. The ones that exist there are from whichever dependency was last extracted, so its `.MTREE` and things like that. If it's preferred though I can make these explicit too :smiley: 

As usual, all feedback is appreciated!

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v11.12.20220924"
+PROGVERS="v11.12.20221005"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
@@ -20235,6 +20235,8 @@ function steamdedeckt {
 			setYadBin "ai" "sd"
 		fi
 		
+		find "$STLDEPS" -type f \( -name "*.zst" -o -name "*.gz" -o -name ".*" \) -exec rm {} \;
+
 		# update/set compatibility tool to git stl:
 		if [ "$INFLATPAK" -eq 0 ]; then
 			CompatTool "add" "$PREFIX/$PROGCMD"


frostworx:
Thanks for the idea.
`rm "$STLDEPS/."*` is dangerous though.
Think of what could happen when for some weird reason `$STLDEPS` is an empty variable.
Not worth the possible trouble for some cosmetical cleanups

sonic2kk:
Hmm, good point. I guess we could either check for and remove the dotfiles manually, and/or check to make sure `$STLDEPS` is a directory (`[ -d "$STLDEPS" ]`?)

Or if it's not worth the trouble at all, no worries. I could always remove that specific part and keep the cleanup of the archive (should be safer since it only removes them if downloaded).

frostworx:
a find in "$STLDEPS" with the used suffixes as pattern would be the best solution imho- so something like
`find "$STLDEPS" \( -name "*.zst" -o -name "*.gz" \) -exec rm {} \;`
_(untested! :))_

sonic2kk:
Thanks for the code snippet :-) I made a little test folder on my PC and fiddled around to make sure it behaved as expected and it seems to. Also looked up the syntax to understand better how that snippet worked :smile: 

I set `-type -f` to make sure only files were matched, and `-o -name ".*"` to remove the dotfiles, however if it's still preferrable to not remove these files that's totally okay and I'll take this out. Just wasn't sure what the consensus was, since this snippet *should* be safer - `find` should just warn and skip of that directory doesn't exist. We could also add a check to make sure `$STLDEPS` exists before running this command. Whatever works best :slightly_smiling_face: 

EDIT: I forgot to mention, I haven't tested this on my Deck yet! I pushed it to my branch because I want to test it on my Deck. So it's not in a state where I'm fully confident it will work on my Deck yet because I haven't gotten to verify just yet. Will test ASAP though (probably won't be until sometime tomorrow).

frostworx:
looks good for me, thanks for the update! :)
I see no reason why this should not work on the Steam Deck, but will wait with the merge.
Take your time! :)
btw, https://hacktoberfest.com/ is running _(just added a tag to the project, though it didn't help much last year)_

sonic2kk:
Thanks for the patience, just got to test on my Deck and it all works :-) Ready for merging!

(Btw, `steamdedeckt` is an awesome function name 😉)

frostworx:
Hehe, thank you very much for testing! 👍

PR #622, opened 2022-10-05
Add Open Issue action to Tray Icon and Command Line

sonic2kk:
Adds an entry in the Tray Icon menu to open an issue, which uses `xdg-open` to open `$PROJECTPAGE/issues/new/choose` in the user's default browser. This might make it a bit handier for users to find a way to submit an issue. It adds an associated option to the command line as well called `openissue`, with a shorthand `oi`.

I'm not sure if maybe we should have options for specifically submitting a bug report ("Report a Bug") and also having a feature request ("Request a Feature") option either *instead* of or in *addition* to this option. In that case we could also further break down the command where the default just opens the issue tracker but we could pass `bug`/`b` or `featurerequest` or `fr` to open the relevant bug report section.

The issue (no pun intended) with this idea is that it'll rely on a specific URL that points to a specifically named and specifically labelled new issue. It'll have to look something like this for the bug reporting URL: `https://github.com/frostworx/steamtinkerlaunch/issues/new?assignees=&labels=bug&template=bug-report.md&title=` - Should the URLs ever change, or should more templates ever be added, we'd need to maintain this. Having the option to just open the issue tracker is very straightforward and should be clear to most users once they click on it that they have the option of submitting a bug or requesting a feature.

I also had a think about whether this could fit in the UI anywhere as a button. I couldn't really think of anywhere after looking around, I don't feel like it's important enough to make it a button along the bottom, and I'm not sure it fits as a button on the main menu. The tray icon seemed like the obvious choice but not all users might think to look there. Though the same could be said for any UI so maybe this is a non-issue and I'm just thinking too heart :sweat_smile: 

Once again, all feedback is welcome on this :smile: Thanks!


sonic2kk:
Had a branch with some *very* minor changes to `englishUK.txt`, was looking around for more and sort of biding my time until another language file change to fit it in. Just realised *this* PR actually touches language files, so I put those changes in there too. the `englishUK.txt` file is very similar to `english.txt` but someone went to the trouble of making it so I thought it would be nice to help try and ""maintain"" it (even if they're only small changes).

AliceDTRH:
I think it would probably be best for users to pick the type of issue in the browser using Githubs own page for this IMO.

sonic2kk:
Makes sense to me now that I think about it more yup, that's the current approach the PR takes and saves me some effort having to add those extra options I was talking about :smile: Better to just open the issue tracker and let the user pick what fits best.

frostworx:
nice, thanks!

frostworx:
merged, I agree it is complete enough as-is.

PR #645, opened 2022-10-29
DRAFT for https://github.com/sonic2kk/steamtinkerlaunch/issues/643

### LINKED ISSUE #643: [enhancement] Per-Game ReShade Versions
## System Information
- SteamTinkerLaunch version: v11.12.20221028-1
- Distribution: Arch Linux

## Feature Description

Hey there! I just wanted to suggest two enhancements:

- Per-game Reshade versions
- Shader support for native Linux games

 Sometimes, a different Reshade version is needed as a temporary workaround, but the Reshade version can only be set globally, which makes switching back and forth between versions more tedious than it needs to be, IMO.
 
Also, it might not be something available at present, but it'd be great to have shader support for native Linux games! Currently, it doesn't seem like enabling either vkBasalt (which isn't being developed right now, as I understand) or Reshade actually does anything for native games.
 
 Thanks!

zany130:
Bump Reshade version and extract final .dlls into its own folder

TBH I have no clue what I am doing and this is still very early, so any advice or tips would be greatly appreciated


steamtinkerlaunch
```diff
@@ -7400,29 +7400,30 @@ function dlReShade {
 	else
 		DLVERS="$1"
 	fi
-	rm "$RESHADESRCDIR/$RS_64" 2>/dev/null
-	rm "$RESHADESRCDIR/$RS_32" 2>/dev/null
-	rm "$RESHADESRCDIR/${RS_64//.dll/.json}" 2>/dev/null
-	rm "$RESHADESRCDIR/${RS_32//.dll/.json}" 2>/dev/null
+	#rm "$RESHADESRCDIR/$RS_64" 2>/dev/null
```
zany130: Do we need this?

zany130:
So I got extracting reshade into its own folder down 
image

however, I am stuck on getting `overrideReshadeVersion function` to work consistently. For some reason, it is not always run ( added a test write log locally at the beginning of the function and it sometimes doesn't write out so it seems the function isn't constantly being called

sonic2kk:
I had a look at your changes. Maybe it's only being called sometimes because the ReShade download function might only be getting called sometimes? I noticed your overrideReShade function is not called in the ReShade install function, so maybe it isn't being called if ReShade already exists? Just a guess for now :-)

Actually, it's probably in `prepareReshadeFiles` that you would want to make a call as well as, or instead of, the call in `dlReShade`. Not sure but from reading a call in both places might be a good idea. To my understanding of the ReShade install flow, calling `overrideReShade` in `prepareReshadeFiles` should ensure the right version of ReShade is set with an override I think?

I added a review comment of where I think the call could go. It may not be right but it could be a hint at where to add an extra call.

zany130:
yup moved it to `prepareReshadeFiles`, and it works consistently now, it seems the only thing missing now is adding a text box in yad with a few drop-down options ( absolutely have no idea how to do this never worked with yad before 😅)

sonic2kk:
Awesome! Looking great so far!

> it seems the only thing missing now is adding a text box in yad with a few drop-down options

No worries, I'll try and give a couple of hints. (edits to follow...)

The first hint I'll give is that before each test, run `rm -rf /dev/shm/steamtinkerlaunch` - From personal experience this helps make the UI update during testing. No idea why this is needed if I am truly honest, but from experience this gets the UI to update so you can see your changes.

To actually make the textbox, I would start by adding the strings to the language file. I would start with whatever langfile STL defaults to on your system, e.g. `english.txt`. This way you only have to add a couple of strings and if you want to change their wording it'll be very easy. Then at the end once you're happy with the phrasing and stuff, you can copy+paste the strings into the other language files.

Language file strings are conventionally prefixed with `GUI_` for the actual GUI string, such as the text that will appear before the textbox. You could call this `GUI_RESHADEOVERRIDE`. Then the tooltip is conventionally prefixed with `DESC_`, followed by the same name as the corresponding `GUI_` string. So in this case you could call it `DESC_RESHADEOVERRIDE`.

<hr>

For actually creating the dropdown, you can look at other dropdowns and copy them. The relevant function should be `AllSettingsEntriesDummyFunction`. Then you can create your combobox there, I would put it under Shader Options (under the heading `--field="$(spanFont "$GUI_OPTSSHADER" "H")":LBL "SKIP"`).

To actually create your combobox, you can look at other examples, such as how `GUI_HARDARGS` does it. Here is an example of how you could do it (not tested, just an example):
```
# Spacing here is just cosmetic for the code afaik
--field="     $GUI_RESHADEOVERRIDE!$DESC_RESHADEOVERRIDE ('RSVERSOVRD')":CBE "$(cleanDropDown "${RSVERS/#-/ -}" "$RESHADEVERSIONS")" `#CAT_Misc` `#MENU_GAME` \
```
Explanation:
- This will create a **C**ombo**B**ox **E**ntry (CBE) that has the `GUI` text of whatever `GUI_RESHADEOVERRIDE` is. the `DESC_` specifies the tooltip that appears on hover. These are separated with an exclamation mark
- The `('RSVERSOVRD')` is just an extra string that appears on hover for the tooltip. This means your tooltip will show up like `This is my cool tooltip, a cooltip if you will ('RSVERSOVRD')`
- `:CBE` tells Yad the type of field this is, in this case a combobox
- `cleanDropDown` is function created by Frostworx for more easily creating dropdowns from string arguments
  - The first argument to this, `"${RSVERS/#-/ -}"`, is the default value for the combox. This is set to the `$RSVERS`, meaning it'll match the current global ReShade version. I'm not sure if this is exactly the behaviour you will want, but it is just an example :-) You might want to have `RSVERSOVRD` equal the global ReShade version by default, and in your override function, check if they are not equal with `if ! [ "$RSVERSOVRD" = "$RSVERS" ]; then`. Just a thought, maybe there's a better way to go about this. But I think having the version that shows up in the ReShade Override box match the global reshade version by default makes for better UX.
  - The second argument, `"$RESHADEVERSIONS"` is a string with exclamation marks to separate the values. In this case, I imagine it to be a preset list of ReShade values. You could set this somewhere else in the code like `RESHADEVERSIONS="5.4.2!5.0.0!4.2.0!3.5.3"` and they'll show up in the dropdown box. You don't need to include the default version in this list, it'll get populated automatically.
    - Having these versions stored in a string means you can basically do the same thing in the Global Menu

It could just be that I'm sleepy, but I don't remember how STL sets the value for the variables from these menu items. Maybe it knows based on the `('RSVERSOVRD')` line.

Hope that gives you a good starting point. Please feel free to ask if anything here is unclear or if you need any more guidance :-)

sonic2kk:
Another thing we might need to do is write out the overridden ReShade version to the game's config file, if we aren't doing that already. It shouldn't be stored globally, but in the game config. But first we should focus on getting the UI element there and getting it working with a selected version, then we can store it (I believe it should be added in `setDefaultCfgValuesdefault_template` with something like `if [ -z "$RSVERSOVRD" ]						; then	RSVERSOVRD="whatever"; fi`, where the default value matches the global ReShade version)

Also, merge conflict is probably the versioning change. Don't worry about resolving it till the end :-)

zany130:
still planning on doing this btw just been very busy. Hopefully I should finish this tomorrow if not this weekend

sonic2kk:
No worries, take your time :-) There's no rush on my side (I'd like this in the next release but that's still over a month away at least)

zany130:
Sorry for the delay, but finally found some time to work on this. Your hints were really useful thanks @sonic2kk. As for this:

> * The second argument, `"$RESHADEVERSIONS"` is a string with exclamation marks to separate the values. In this case, I imagine it to be a preset list of ReShade values. You could set this somewhere else in the code like `RESHADEVERSIONS="5.4.2!5.0.0!4.2.0!3.5.3"` and they'll show up in the dropdown box. You don't need to include the default version in this list, it'll get populated automatically.

I think specialk version select dropdown auto populates with GitHub releases couldn't we do something similar here? Trying to find the relevant code now

EDIT [here](https://github.com/sonic2kk/steamtinkerlaunch/blob/b1f5e6e494c7ba4e0131c529d78b84d5e5749dc1/steamtinkerlaunch#L4874) and [here](https://github.com/sonic2kk/steamtinkerlaunch/blob/b1f5e6e494c7ba4e0131c529d78b84d5e5749dc1/steamtinkerlaunch#L8394)

sonic2kk:
Awesome, thanks for the work!

Great catch with the SpecialK versions, from glancing at the code yup a similar approach could be taken - assuming it still correctly fetches the versions, just glanced at the code. If it doesn't we can find another way, but fetching the versions is a great idea.

Haven't reviewed the new stuff in-depth yet but I will soon, and once you're ready I can test the branch :-)

zany130:
pretty much done except for two things

1. Dropdown isn't showing hardcoded reshade versions to download just shows up as empty (though a version can manually be typed in)
2. for some reason `RSVERSOVRD` is getting saved as `RSVERS` need to look into that (prob somewhere I set `RSVERSOVRD = RSVERS` where it shouldn't be

EDIT: oh and of course reusing the specialk code to auto-generate the drop-down but I'm leaving that for last, first I want to get it working in its current form and also solve all the conflicts 

EDIT2: fixed issue 2 will upload shortly

PR #649, opened 2022-11-08
Reshade version override

### LINKED ISSUE #643: [enhancement] Per-Game ReShade Versions

zany130:
https://github.com/sonic2kk/steamtinkerlaunch/pull/645
Sorry didn't mean to close the previous PR  accidentally merged your changes into my branch, which not only closed the PR but also messed some things up. So I deleted the branch and created a new one
This branch should be up-to-date and correct. The only things left to do here is 

1. implementing the same auto-populating check box as specialk (right now, it still shows as with no values to select from)
2. possibly remove the parts I commented out, as they are no longer needed?
I already did some preliminary testing, and everything seems to work.

Again sorry about the mistake I think everything is as it should be


sonic2kk:
No worries about the other PR, it's all good :-) 

> implementing the same auto-populating check box as specialk (right now, it still shows as with no values to select from)

They might be blank because it's looking for `$RESHADEVERSIONS` too early, it might not exist when that dialog is created. You might want to create a separate function and call it right under the `createDLSpekList` call in `MainMenu`. You could call this function `createDLReShadeList` and then later on put your logic for generating the ReShade versions into that. Then you can set your `$RESHADEVERSIONS` string and Yad should see it.

The rest of your `overrideReShadeVersion` function should be able to stay in-tact.

I'll add a few more comments in the PR (mostly the same as what I added in the last PR) but it's looking pretty good so far! :smile:

steamtinkerlaunch
```diff
@@ -2591,7 +2591,7 @@ function setDefaultCfgValues {
 		if [ -z "$ROBERTAARGS" ]						; then	ROBERTAARGS="--wait-before-run"; fi
 		if [ -z "$LUXTORPEDACMD" ]						; then	LUXTORPEDACMD="$STEAMCOMPATOOLS/luxtorpeda/luxtorpeda"; fi
 		if [ -z "$LUXTORPEDAARGS" ]						; then	LUXTORPEDAARGS="wait-before-run"; fi
-		if [ -z "$RSVERS" ]								; then	RSVERS="5.3.0"; fi
+		if [ -z "$RSVERS" ]								; then	RSVERS="5.4.2"; fi
```
sonic2kk, line 2603: This isn't anything for you to change right now, but just a note that once you get the version fetching working, you'll probably want to bump this version to whatever the latest from GitHub is. You can set `RSVERS=whatever` in the function you end up calling in the `MainMenu`, and then call `updateConfigEntry "RSVERS" "<latest ReShade Vers here>" "$STLDEFGLOBALCFG"` (where the syntax is `"var to update" "var value" "config file to update"`).

Not a big deal right now and we can leave this default at whatever since you'll be able to bump it before the user sees it anyway - maybe check if the value is valid before overwriting the `RSVERS` in case we can't access GitHub or whatever at that moment (we can cross that bridge later, not sure what values you might get back from the GitHub check so I'm not sure how we'd validate it just yet).

steamtinkerlaunch
```diff
@@ -7433,37 +7437,51 @@ function dlReShade {
...
 		DLVERS="$1"
 	fi
-	rm "$RESHADESRCDIR/$RS_64" 2>/dev/null
-	rm "$RESHADESRCDIR/$RS_32" 2>/dev/null
-	rm "$RESHADESRCDIR/${RS_64//.dll/.json}" 2>/dev/null
-	rm "$RESHADESRCDIR/${RS_32//.dll/.json}" 2>/dev/null
+	#rm "$RESHADESRCDIR/$RS_64" 2>/dev/null
+	#rm "$RESHADESRCDIR/$RS_32" 2>/dev/null
+	#rm "$RESHADESRCDIR/${RS_64//.dll/.json}" 2>/dev/null
+	#rm "$RESHADESRCDIR/${RS_32//.dll/.json}" 2>/dev/null
```
sonic2kk: Copied from review on #645:

If we're re-downloading we might want to remove any files that already exist. That is, say we're re-downloading because of a missing `$RS_64`. If we try to do that and the `.exe` or whatever already exists, that _might_ pose an issue (not 100% sure).

We could pass another optional argument to `dlReShade` to specify if we want to redownload. Usage could be something like `dlReShade "" "1"`, and then the ReShade function could have something like this at the beginning right under the check for `$1`:

```
if [ -z "$2" ]; then
    writelog "INFO" "${FUNCNAME[0]} - ReShade redownload was requested"
    RESHADEREFETCH=1
else
    RESHADEREFETCH=0
fi
```

Passing an empty string for the first argument should still pass the check for `[ -z "$1" ]`.

Then we can check `if [ "$RESHADEREFETCH" -eq 1 ]` and then run those commented `rm` commands.

<hr>

The files might just get overwritten anyway when we redownload and extract again. But if they don't we might want to do this, in case extraction fails *or* in case the old files don't get overwritten. If overwriting files works, this won't be an issue :-)

steamtinkerlaunch
```diff
@@ -7433,37 +7437,51 @@ function dlReShade {
...
 
 	dlCheck "$RSSETUP" "$DLDST" "X" "Downloading $RSSU"
-	if [ ! -f "$DLDSTV" ]; then
-		writelog "INFO" "${FUNCNAME[0]} - Creating backup '${RSSU}_${DLVERS}.exe'"
-		cp "$DLDST" "$DLDSTV"
-	fi
+	#if [ ! -f "$DLDSTV" ]; then
+	#	writelog "INFO" "${FUNCNAME[0]} - Creating backup '${RSSU}_${DLVERS}.exe'"
+	#	cp "$DLDST" "$DLDSTV"
+	#fi
```
sonic2kk: No rush to do it *right* now or anything if you want to keep testing but as you asked which commented lines can be removed, I believe these can be :-) Or if you would prefer leave them that's fine, I just don't think it's needed (I'll also need to do some testing myself before merging, but since we create new directories now this shouldn't be needed).

steamtinkerlaunch
```diff
@@ -7433,37 +7437,51 @@ function dlReShade {
...
 		writelog "INFO" "${FUNCNAME[0]} - Downloaded and extracted ${RESH}-v${DLVERS} file '$DLDST'"
 	fi
 }
 
+function overrideReShadeVersion
+{
+	RESHADEVERSIONS="5.4.2!5.0.0!4.2.0!3.5.3" # List some hardcoded reshade versions to selct for the override #TODO replace with a auto poplating list based on the last few releses
+
+	if ! [ "$RSVERSOVRD" -eq "0" ]; then
+		writelog "INFO" "ReShade version - Set to '$RSVERSOVRD'"
```
sonic2kk: Can't multi-select lines to review apparently, but each log message should start with `${FUNCNAME[0]} - ` - That way in the logs they show up like:

```
Tue  8 Nov 20:53:29 GMT 2022 INFO - overrideReShadeVersion - ReShade version - Set to '5.4.2'
```

This just makes it easier to trace when debugging. This applies to all log lines but GitHub won't let me select multiple lines that aren't continuous, and I don't want to spam you with review comments on each line with the same message :-)

zany130:
oh, sorry didn't see you added comments on the old one.

Also looks like I still forgot something, as it's not installing the overridden reshade into the game so that needs to be fixed

sonic2kk:
No worries, maybe something in `installReshade` needs updating (maybe `INSTDESTDIR` needs updated?)

steamtinkerlaunch
```diff
@@ -7475,8 +7493,8 @@ function prepareReshadeFiles {
 		dld3d47 "32"
 		dld3d47 "64"
 
-		if [ ! -f "$RESHADESRCDIR/$RS_64" ] || [ ! -f "$RESHADESRCDIR/$RS_32" ]; then
-			writelog "404" "${FUNCNAME[0]} - '$RESHADESRCDIR/$RS_64' and/or '$RS_32' missing - downloading"
+		if [ ! -f "$RESHADESRCDIR/$RSVERS/$RS_64" ] || [ ! -f "$RESHADESRCDIR/$RSVERS/$RS_32" ]; then
+			writelog "404" "${FUNCNAME[0]} - '$RESHADESRCDIR/$RSVERS/$RS_64' and/or '$RS_32' missing - downloading"
```
sonic2kk, line 7575: Do we want to also check for missing ReShade JSON files for Vulkan, and have that prompt a re-download of ReShade? I know you didn't write this part originally, it's just something that caught my eye :-) If we're re-downloading on missing ReShade 32/64bit DLLs, do we want to also redownload on missing JSON files for Vulkan.

zany130:
nah I don't think I changed anything there (at least, looking at some old files I have on my computer, I didn't) I was missing some changes to `installRSdll` though it doesn't seem to be 100% consistently working, it seems the global `RSVERS` is overwriting it

sonic2kk:
Was that an issue we had before related to where the ReShade function was being called? https://github.com/sonic2kk/steamtinkerlaunch/pull/645#issuecomment-1296034113

zany130:
hmm don't think so it seems my function is always getting called it just seems when yad refreshes it sets it back to the global version ( it is also successfully being saved to the game config file so idk why yad is doing that) 

I  did have this issue before when I set the default value of `RSVERSOVRD` to be equal to `RSVERS` but I changed the default to 0 so that should no longer be an issue....

EDIT: only thing I can think of that can be causing this is this 
```
$GUI_RESHADEOVERRIDE!$DESC_RESHADEOVERRIDE ('RSVERSOVRD')":CBE "$(cleanDropDown "${RSVERS/#-/ -}" "$RESHADEVERSIONS")" `#CAT_Shader` `#SUB_ReShade` `#MENU_GAME` \`
```

EDIT2: yeah just reread you explanation of the yad stuff and that's exactly whats happening 😅

Will finish this sometime this week (Tomorrow probably) but I'm thinking i can leave the yad default blank so it doesn't update everytime yaw is called?

Then i can just set the default value in the default Cfg function to be 'RSVERS' and check for that in my function

sonic2kk:
> Will finish this sometime this week (Tomorrow probably) but I'm thinking i can leave the yad default blank so it doesn't update everytime yaw is called?
>
> Then i can just set the default value in the default Cfg function to be 'RSVERS' and check for that in my function

No rush :-)

You could potentially leave it blank but you'll want the values for the different ReShade versions in the dropdown. Maybe the default should match the global ReShade and then in your override function you can check if the versions are different?

zany130:
The problem is (at least it seems might be wrong) whatever value I set after the cleanDropDown gets set to `RSVERSOVRD` (and then to `RSVERS`) every time yad is called.

I noticed this because I can type in say 5.4.1 hit save and when the menu comes back ti will be set back to the global (5.4.2). Its saving in the config file correctly as 5.4.1 but in yad its getting overridden to 5.4.2 so I think its because of that.

so what I am thinking is I can leave the yad default blank and then in my versions function  _I think_ I can just add `"$RSVERS` to the list of values , then in `setDefaultCfgValuesdefault_template` I can set `RSVERSOVRD  = `"$RSVERS"`

so that way, if no version has been set in the config, it will default to the global version otherwise it will use the version set in the game config. Then the dropdown will have a option to see it back to the global version

sonic2kk:
> I noticed this because I can type in say 5.4.1 hit save and when the menu comes back ti will be set back to the global (5.4.2). Its saving in the config file correctly as 5.4.1 but in yad its getting overridden to 5.4.2 so I think its because of that.

Oooohhh this may be because the first value in the `RESHADEVERSIONS` string in `createDLReShadeList` is `5.4.2`. You could try changing that to something else and seeing if it defaults to that :-)

I would try that before delving into keeping the value blank :-)

zany130:
Actually, yup you are right its `RESHADEVERSIONS` that's overriding the `RSVERSOVRD` not the default value set in the yad command.

I removed the default value part from the yad command and it still overrides it what ever is the first version in `RESHADEVERSIONS`

steamtinkerlaunch
```diff
@@ -7433,7 +7433,7 @@ function installd3d47dll {
 }
 
 function createDLReShadeList {
-	RESHADEVERSIONS="5.4.2!5.0.0!4.2.0!3.5.3" # List some hardcoded reshade versions to selct for the override 
+	RESHADEVERSIONS="5.0.0!4.9.1!3.4.1" # List some hardcoded reshade versions to selct for the override
```
sonic2kk: You could set the first item in this list to be `${RSVERS}` :-)

zany130:
So no clue why `createDLReShadeList` is overriding whatever the user enters into the text box. Everything works as expected ( and it will keep different versions of specialk in there own folder) however, if you just enter the value and hit save and play, it will also keep this value until the next time you open the game menu, at which point if you hit save it will get set to the first value in `createDLReShadeList`

as for auto-populating the list, that looks a lot more complicated than I originally thought. It SHOULD be technically possible by using the tagged releases here https://github.com/crosire/reshade/tags

other than those two things, I think this is pretty much complete and once I figure out how to prevent the override, it can be merged. 

What do you think @sonic2kk ?

sonic2kk:
First of all, I totally messed up and didn't understand how GitHub PR reviews worked. I was adding comments and thought they'd show up right away, but I had to hit a "submit review" button. Wow I'm dumb... I'm so sorry for that. It must've seemed like I was silent on your work! I was leaving all that feedback incorrectly...

I submitted the review now I think, I noticed I couldn't view it in private browsing so I hunted around and I think I got it now :sweat_smile: 

<hr>

It's getting a bit late here, but I'll take a look at the code and see why it's not setting the version properly. If I don't come up with anything tonight I'll take a closer look tomorrow.

When you enter the value and it keeps the override version of ReShade, does it use that version of ReShade? Is there even a way to tell that? :sweat_smile: 

<hr>

For fetching the versions, we can come to that last.

<hr>

Really sorry again for my inexperience here with reviewing PRs :sweat: But I agree once we have some discussion on the feedback I left and once we get the ReShade version issue sorted out, we can take a look at the tag fetching for getting the versions and getting this merged. We can just fetch the last 5 or so, maybe we can use similar logic to the `getLatestGitHubExeVer`. I actually wrote a function recently that fetches the latest tag for a project, I tried to adapt it to get the last N number of tags:

```bash
#!/usr/bin/env bash

GHURL="https://github.com"
WGET="wget"

function getTagsList {
    PROJURL="$1"
    N="$2"

    RELEASESURL="${PROJURL}/releases"
    EXPANDEDASSETSURL="${RELEASESURL}/expanded_assets"
    TAGSURL="${PROJURL}/tags"

    TAGSGREP="${RELEASESURL#"$GHURL"}/tag"

    LATESTTAGS="$("$WGET" -q "${TAGSURL}" -O - 2> >(grep -v "SSL_INIT") | grep -m "$N" "$TAGSGREP" | grep -oE "${TAGSGREP}[^\"]+")"
    LATESTTAGS="${LATESTTAGS##*/}"
    LATESTTAGS="${LATESTTAGS//$'\n'/!}"

    echo "$LATESTVER"
}

echo "Latest STL ver: $( getLatestGitHubReleaseVer "https://github.com/sonic2kk/steamtinkerlaunch" "5" )"
```

I did a brief test and that returns duplicates (probably will just need some extra logic in the grepping to ignore duplicates or something) so it's not perfect but something along these lines might work. If you'd like it's something I could implement after merging :slightly_smiling_face:

sonic2kk:
Some late-night Bash scripting later, I made a function that fetches the last N tags from a given repo. Check it out:

```bash
#!/usr/bin/env bash

GHURL="https://github.com"
WGET="wget"

function fetchGitHubTags {
    PROJURL="$1"
    N="$2"

    RELEASESURL="${PROJURL}/releases"
    TAGSURL="${PROJURL}/tags"

    TAGSGREP="${RELEASESURL#"$GHURL"}/tag"

    mapfile -t BASETAGS < <("$WGET" -q "${TAGSURL}" -O - 2> >(grep -v "SSL_INIT") | grep -oE "${TAGSGREP}[^\"]+" | sort -urV | grep -m "$N" "$TAGSGREP")
    for TAG in "${BASETAGS[@]}"; do
        basename "$TAG"
    done
}

RESHADETAGS="$( fetchGitHubTags "https://github.com/crosire/reshade" "5" )"
RESHADETAGS="${RESHADETAGS//$'\n'/!}"
RESHADETAGS="${RESHADETAGS//v/}"

echo "${RESHADETAGS}"
```

Tested with ReShade and STL and it works :-) 

You should be able to just drop the function into STL and use it. The echoes are just there to give you an idea of what the output is. Usage is `fetchGitHubTags "<URL>" "num of tags to get"`.

The function is really limited by how many tags come back on one GitHub page, which I think is 10. But that's fine to be honest.

To save on calls, you might want to also wrap a call to that function in an `if [ -z "$RESHADEOVERRIDEVERSIONS" ]` or something :-) That way it's only called once while STL is open. I dunno if it's worth saving the actual list of override values, I think it's fine to just fetch it once per execution. There is a chance that a user could spam it, but it's probably not a big deal.

*(if I'm feeling brave, maybe in future I'll expand a function like that to go back in pages and get all the tags that way, but that is hurting my already sleepy brain too much :sweat_smile:)*

steamtinkerlaunch
zany130: that actually ended up being my solution to the override problem I set the first value to `$RSVERSOVRD`  that way it will overwrite itself with itself :) 

then I removed the default value from the yad menu command (as that was also overriding and since I am always initializing `RSVERSOVRD` to be the same as `RSVERS` it will by default auto-populate `RSVERS` 

I'll push this change in a bit so you can review it when you have time

steamtinkerlaunch
sonic2kk: Thinking about this again, I don't think it's needed. When we download and extract with `unzip` I'm almost certain it does just overwrite the existing files in the directory. Could you test this? You can check the file modified time and see if it changed, or if they have a time when they get downloaded, modify a file yourself so it says "just now" and then you'll see the old time when you extract it again.

And if you'd like just to be certain, when redownloading you could first remove the ReShade version directory, then re-create it before downloading :-)

steamtinkerlaunch
zany130: Think I caught them all sorry about that

steamtinkerlaunch
sonic2kk: Actually thinking about this again I'm not sure this is a good idea. I'm not sure that we *want* to always bump the ReShade version? It introduces a few concerns in my mind that for some reason I didn't consider before:
- It puts a lot of reliance on the version fetching to be correct. I suppose we could validate it with Regex and use `-n` checks to make sure it's not blank but bleh
- There could be potential compatibility issues

You're probably more familiar with ReShade than me, do you think always bumping to the latest version automatically is a good idea?

To be honest, I think bumping the version here is out of scope of this PR so I won't ask you to do it. I'd like your input on whether or not you think it's a good idea though. Once you reply you can mark this resolved, you don't have to implement this :-)

steamtinkerlaunch
```diff
@@ -7427,43 +7432,59 @@ function installd3d47dll {
...
+	fi
+}
+
 # prepare reshade files if not found:
 function prepareReshadeFiles {
-	if [ "$DOWNLOAD_RESHADE" -eq 1 ]; then
-		writelog "INFO" "${FUNCNAME[0]} - DOWNLOAD_RESHADE enabled"
+overrideReShadeVersion
+if [ "$DOWNLOAD_RESHADE" -eq 1 ]; then
+	writelog "INFO" "${FUNCNAME[0]} - DOWNLOAD_RESHADE enabled"
```
sonic2kk: Really nitpicky and not sure if GitHub is just messing up the formatting but does this need indented?

steamtinkerlaunch
sonic2kk: Looks good to me! Wish I could thumb up some parts of the code or something :sweat_smile: Looking awesome!

steamtinkerlaunch
zany130: oh, you meant to add code to auto-bump this... that would be pretty easy if there were releases on the reshade github we could then use this https://docs.github.com/en/rest/releases/releases#get-the-latest-release but I don't know about tags

EDIT: not against auto bumping the reshade version. reshade users are pretty much expected to run the latest version. HOWEVER very rarely these updates break Linux compatibility....

sonic2kk: > that would be pretty easy if there were releases on the reshade github

The output of the function I posted could be edited to get the first tag (`echo "${RESHADETAGS}" | head -n1` would probably work), but I agree probably easy :-)

> not against auto bumping the reshade version. reshade users are pretty much expected to run the latest version. HOWEVER very rarely these updates break Linux compatibility....

Aaaaahhhh this is what I was wondering. This is out of scope for the PR so I won't push any of this on you, but I think if this is added in future it'll be a global menu option called "Auto latest ReShade" which will take priority over the global ReShade but **not** over the overridden ReShade. Then it'll fetch and get the latest version each time I guess.

Not needed for this PR but thank you for the insight all the same! :smile:

steamtinkerlaunch
zany130: I'm not sure if you can install shade for Vulkan games on Linux. if you look frostworx didn't even add an install case for Vulkan games. the JSON is extracted but never really even used so not sure if really nessicary I could add it though since it shouldn't be to hard to do

steamtinkerlaunch
```diff
@@ -7427,43 +7432,59 @@ function installd3d47dll {
...
 	else
-		"$UNZIP" -q "$DLDST" -d "$RESHADESRCDIR" 2>/dev/null
+		"$UNZIP" -q "$DLDST" -d "$RESHADESRCDIR/$RSVERS" 2>/dev/null
 		writelog "INFO" "${FUNCNAME[0]} - Downloaded and extracted ${RESH}-v${DLVERS} file '$DLDST'"
 	fi
 }
 
+function overrideReShadeVersion
+{
+	if ! [ "$RSVERSOVRD" -eq "$RSVERS" ]; then
```
sonic2kk: I think I actually suggested this syntax so sorry if this is wrong, but does this check work? If they're both strings you might have to do a check like `if [[ "$RSVERSOVRD" = "RSVERS" ]]; then` (note the double square brackets -- Bash is fussy about this for some reason...)

If it works though, we can leave it as is :-)

steamtinkerlaunch
zany130: huh shouldn't be indenting that much it should just be a small indent to show it goes in the if ill take a look at that

steamtinkerlaunch
zany130: i think you are right now that I think about it I never did see that condition pass ... meaning RSVERSOVRD is equal to RSVERS

sonic2kk:
I had a couple more thoughts just quickly before I sign off for the night:

It might be a good idea to guard this behind a checkbox as well, that way a user can more easily toggle this on and off. It might give them more peace of mind and it's easier to remember to turn off the checkbox than to set the ReShade version back to the global version. We could then do a check in the same place where you have the check for the override and global ReShade versions being the same, so something like `if [[ "$RSOVRDVERS" = "$RSVERS" ]] && [ "$RSOVRD" -eq 1 ]; then`

It hopefully shouldn't be too much to implement. If you need help with it I can try post an example tomorrow. Should be the same idea as the dropdown though: Add the strings, the UI element (you could look for some examples in the code that end in `:CB`) and add the variable, also saving the variable to the game config.

<hr>

The other thing I thought of is, we may need some better validation for checking the ReShade version, but I'm not even sure this exists for the regular ReShade version so likely this is out of scope for this PR and something that should be worked on separately :-)

<hr>

Sorry I just thought of this now so late on. If you aren't up for looking into adding any more functionality that's totally fine and it can be added at a later date :-)

zany130:
no worries, I have been toying with the idea of a toggle since the beginning going back and forth the main reason for not doing it if I'm being honest, is laziness and being afraid of more code :sweat_smile:  but your right it should be more or less the same as what I already did so I think I can manage...

steamtinkerlaunch
sonic2kk: Yeah this logic might need a little bit of tweaking. If the two versions are equal we probably won't want to do `RSVERS="$RSVERSOVRD"`.

Maybe it could look like this?

```bash
# Assumes we have a checkbox like in my comment - No rush, just letting you know the context of how I drafted this snippet :-)
if [ "$RSOVRD" -eq 1 ]; then  # This is meant to be the checkbox var
    if [[ ! "$RSOVRDVERS" = "$RSVERS" ]]; then
        writelog "INFO" "${FUNCNAME[0]} - Overriding global ReShade version '$RSVERS' with '$RSVERSOVRD'"
        RSVERS="$RSOVRDVERS"
    else
        writelog "SKIP" "${FUNCNAME[0]} - ReShade Override version and ReShade global version match - Not overriding"
    fi
else
    writelog "SKIP" "${FUNCNAME[0]} - ReShade override is disabled - Skipping"
fi
```

steamtinkerlaunch
sonic2kk: On GitHub it's showing up as not indented at all on those lines but it could be a GitHub glitch :-)

image

steamtinkerlaunch
sonic2kk: > the JSON is extracted but never really even used so not sure if really nessicary I could add it though since it shouldn't be to hard to do

This is interesting. I haven't tested though I own a couple of games and may test, because we could add a note about this to the wiki. I could see why it might not get used though, I think DXVK just passes through to system Vulkan.

I think it would be nice to have though, in case it ever works in future :-)

sonic2kk:
> the main reason for not doing it if I'm being honest, is laziness and being afraid of more code

Totally fair :smile: Good luck! I'll try write up an example if you need it tomorrow.

(Also, review has been more thorough than I thought. I appreciate all of your time! It's looking really good)

steamtinkerlaunch
zany130: > Aaaaahhhh this is what I was wondering. This is out of scope for the PR so I won't push any of this on you, but I think if this is added in future it'll be a global menu option called "Auto latest ReShade" which will take priority over the global ReShade but **not** over the overridden ReShade. Then it'll fetch and get the latest version each time I guess.

ahh ok i like that idea!! I'm getting used to adding option to yad so don't mind a third one lol though when I wake up tomorrow think i earned a break for now :sweat_smile:  (lucky i have no work this week hence the influx of commits)

steamtinkerlaunch
zany130: added skel support for Vulkan only thing missing is installing to the game directory

steamtinkerlaunch
sonic2kk: Thanks, resolved now :-)

sonic2kk:
Wow you made a ton of progress! This is looking virtually complete to me, there are only a few outstanding things left before I think this is ready to merge:
- Does the ReShade override check work? It looks like it should work to me, just wondering 
- Does the ReShade override checkbox value and the ReShade override version itself get written out to the game's config file? I think it should, I think all config entries get fetched and written out (pretty sure it does that based on the value in `('brackets')`)
- ReShade version string fetching - you may just need to do something like this (using the `fetchGitHubTags` function from earlier:

    ```bash
    RSVERSONLINE="$( fetchGitHubTags "$RESHADEPROJURL" "3" )"  # May have to define `$RESHADEPROJURL` in `setDefaultCfgValuesurl` and then write it out to the global cfg in `saveCfgurl` - It's just cleaner to have a definitive URL instead of hardcoding it 
    RESHADEVERSIONS="$RSOVRVERS!$RSVERS!"
   ```
- My own testing once you're happy enough for me to test on my machine (I could test the values but I don't want to test until you're ready)
- Commit squashing + any commented code you want to remove, at the very very end :-)

This is looking like a really comprehensive implementation!

zany130:
> Does the ReShade override check work? It looks like it should work to me, just wondering

yes just tested though some more testing is welcomed ofc 

>Does the ReShade override checkbox value and the ReShade override version itself get written out to the game's config file? I think it should, I think all config entries get fetched and written out (pretty sure it does that based on the value in ('brackets'))

yes but for some reason its not saving with the description so its saving like this
```
RSOVRD="1"
RSOVRVERS="5.4.2_Addon"
```
Only thing missing to implement now is the autobump for the global reshade version

The main thing that needs testing is the `dlReShade` function  I think it might not be handling redownloading corrupted reshade files well....

btw I added the `-o` flag to unzip  in that function as it should allow unzip to override without confirmation from the user



Oh ad the auto populating list isn't working correctly its getting all the downloaded versions and combining them 
image

sonic2kk:
> yes but for some reason its not saving with the description so its saving like this

Maybe take those values out of the config file and put them back in, and see if it populates with the string correctly? Sounds like a strange problem, my initial thinking was maybe the langfiles but I doubt it looking at the code again.

> Only thing missing to implement now is the autobump for the global reshade version

Unless you *want* to add it, this is something I can add another time :-) You've already done a ton here

> The main thing that needs testing is the dlReShade function I think it might not be handling redownloading corrupted reshade files well....

In `prepareReshadeFiles` I would log and inspect the values of `"$RESHADESRCDIR/$RSVERS/$RS_64` to see if they're valid. If it's not downloading with a missing file/missing files maybe it's because the paths aren't correct. A complete and total guess though :sweat_smile: 

> btw I added the `-o` flag to unzip in that function as it should allow unzip to override without confirmation from the user

Awesome catch!

<hr>

One small observation: In that screenshot it looks like the override dropdown values are not correct. Maybe it's being returned as a string on newlines? They should be returned as exclamation mark separated strings, something like this on line 7467:

```bash
RSVERSONLINE="$( fetchGitHubTags "https://github.com/crosire/reshade" "5" )"
RSVERSONLINE="${RSVERSONLINE//$'\n'/!}"
RSVERSONLINE="${RSVERSONLINE//v/}"
```

The second line replaces the newlines with exclamation marks, and the last line removes the `v` from the version - That way we only pass `5.4.2` instead of `v5.4.2`. I should've made that clearer last night, sorry.

<hr>

So overall looking great! The checkbox and dropdown are looking perfect (save for the small values observation above) :smile:

zany130:
> Unless you _want_ to add it, this is something I can add another time :-) You've already done a ton here

Already this far  in might as well see it through :sweat_smile: 
just pushed some skel work for this, however I'm not sure how to go about updating `RSVERS` in the global cfg

>In prepareReshadeFiles I would log and inspect the values of "$RESHADESRCDIR/$RSVERS/$RS_64 to see if they're valid. If it's not downloading with a missing file/missing files maybe it's because the paths aren't correct. A complete and total guess though

yeah no it works fine if I delete one of the files (including the .json now) I just though for some reason there was a check to see if the files were 0 bytes (there is a check for the reshade exe but not the extracted files I think)

>One small observation: In that screenshot it looks like the override dropdown values are not correct. Maybe it's being returned as a string on newlines? They should be returned as exclamation mark separated strings, something like this on line 7467:
RSVERSONLINE="$( fetchGitHubTags "https://github.com/crosire/reshade" "5" )"
RSVERSONLINE="${RSVERSONLINE//$'\n'/!}"
RSVERSONLINE="${RSVERSONLINE//v/}"
The second line replaces the newlines with exclamation marks, and the last line removes the v from the version - That way we only pass 5.4.2 instead of v5.4.2. I should've made that clearer last night, sorry.

yup that did the trick

sonic2kk:
> however I'm not sure how to go about updating RSVERS in the global cfg

Once you have an updated value for `RSVERS` you can update it like this:

```bash
touch "$FUPDATE"  # This is needed for... some reason >_>
updateConfigEntry "RSVERS" "$RSVERS" "$STLDEFGLOBALCFG"  # Usage is "config variable" "value to update it with" "config file path"
```

Maybe that will work as is once you have a bumped `RSVERS` value :-)

> yeah no it works fine if I delete one of the files (including the .json now) I just though for some reason there was a check to see if the files were 0 bytes (there is a check for the reshade exe but not the extracted files I think)

Ahhh okay I suppose that's a good sanity check. I think you can do that with `if [ ! -s "/path/to/file" ]; echo "file is 0 bytes; else echo "file is not zero bytes": fi` :-) Pretty sure I did that for a sanity check on the game art file to see if it needs redownloaded

> yup that did the trick

Awesome :smile:

sonic2kk:
I'm actually running into the exact same problem of not having the descriptions writing out to the config file on my `hmm-support` branch. It seems like the logic to write it out to the config file isn't being called, and the values are only ending up in the config file when they're written out with `updateConfigEntry`? Maybe it's the same issue in this PR, I haven't delved too much into it yet.

<hr>

That aside, looking at your latest commit it's all looking really close to being done - If not already done! I probably won't get any time to test this tonight but over the weekend I'll try and get some testing in. From glancing I don't anticipate any problems though.

I'll see if I can figure out the cause of the config descriptions problem. It's not a big deal and something I'd probably be happy to merge this PR without fixing, but it is something I'll try to fix before the next release.

zany130:
apparently, the issue is when you already have an existing config file, it just adds the new values at the bottom without the description. This happens any time a new config is added to an existing config file it seems as if deleting the config file beforehand fixes it
Also, the auto bump reshade version is not working for some reason its going into the function but it's not passing the check if there is a newer version (for testing I set `RSVERS` to `5.4.1` in my global config)

I know I made an opsie and put `=` instead of `<` which leads into my problem... i think it's because I'm storing `RSVERS` and `RSVERSLATEST` as strings....

steamtinkerlaunch
```diff
@@ -7427,43 +7467,78 @@ function installd3d47dll {
 	fi
 }
 
+function autoBumpReShade {
+	RSVERSLATEST="$( fetchGitHubTags "$RESHADEPROJURL" "1" )"
+	if [ "$AUTOBUMPRESHADE" -eq 1 ] && [ "$RSVERS" = RSVERSLATEST ]; then
```
sonic2kk: Ahh I see what you meant in your comment. Should be an easy fix :smile: 

If it's storing `$AUTOBUMPRESHADE` then you'll want to compare with `[[ "$AUTOBUMPRESHADE" = "1" ]]` (note the double square brackets - Bash is fussy about this...)

For the second check, this will fail unless you use double square brackets. So it needs to be `[[ "$RSVERS" = RSVERSLATEST ]]`

This should hopefully resolve the problems :-)

sonic2kk:
Left a review comment about `autoBumpReshade` (had similar problems in the past myself too - Bash is certainly an interesting beast sometimes...)

> apparently, the issue is when you already have an existing config file, it just adds the new values at the bottom without the description.

Ah! I suppose that's fine then. Though I thought STL had logic to update the global config, maybe it isn't working (hopefully it wasn't something that *I* broke...)

It's not a huge deal, but it's something I'll try to find time to investigate!

sonic2kk:
Ah fishsticks, the HedgeModManager merge created langfile conflicts. No rush on resolving those though, we can do that at the end :-)

frostworx:
nice to see some active contributions 👍

sonic2kk:
@zany130 I resolved the conflicts, that was my bad and I should've waited on you before merging the HMM branch, so I went ahead and cleaned up my mess.

This PR is looking pretty sweet. Is autoBumpReshade the last thing left to do? I believe all of my review comments (except for the autoBumpReshade hint) are resolved now (the last one I forgot to mark resolved as the file overwrite, since you added the `-o` option that should resolve it).

Of course there is still no rush, and you can just let me know when you feel it's ready for merging :smiley:

zany130:
thanks for helping with the conflicts yeah the autobump for the resade version is the only thing left. Haven't had much time yesterday or today to take a look at it, but I hope I can finish it tomorrow. I wrote a response to your review on where I was stuck with the auto bump reshade function.

everything else seems to be working nicely and I think is ready for merging

sonic2kk:
If there's a response to that review it isn't showing up for me, I hunted around but tbh the review workflow is still very new to me :sweat_smile: if you could leave it as a regular comment I can take a look and try help out however I can :-)

zany130:
ahh sorry about that what follows is post i mentioned

I think its the second part that giving me issues (really should be `[ "$RSVERS" \< RSVERSLATEST ]`)  the problem is I think `RSVERSLATEST` is being stored as `"v5.4.2"` 
I tried ` RSVERSLATEST="${RSVERSONLINE//v/}"` but that just sets it to `" "`
btw this is how I have it offline right now 

```
function autoBumpReShade {
	echo "HELLLLLLLOOOOOO"
	RSVERSLATEST="$( fetchGitHubTags "$RESHADEPROJURL" "1" )"
	RSVERSLATEST="${RSVERSONLINE//v/}"
	echo " latest reshade version is '$RSVERSLATEST'"
	echo " current reshade version is '$RSVERS'"
	if [ "$AUTOBUMPRESHADE" -eq 1 ] && [ "$RSVERS" \< RSVERSLATEST ]; then
		echo "updating reshade"
		writelog "INFO" "${FUNCNAME[0]} - Found newer version of '$RESH' - Updating '$RSVERS' to '$RSVERSLATEST"
		touch "$FUPDATE" 
		updateConfigEntry "RSVERS" "$RSVERSLATEST" "$STLDEFGLOBALCFG"
	else
		echo "not updating reshade"
		writelog "SKIP" "${FUNCNAME[0]} - '$RSVERS' is the latest version of '$RESH' - not updating"
	fi
```

sonic2kk:
> really should be `[ "$RSVERS" \< RSVERSLATEST ]`

This might work already with `[[ "$RSVERS" > "$RSVERSLATEST" ]]`

> I tried `RSVERSLATEST="${RSVERSONLINE//v/}"` but that just sets it to `" "`

Ahh, it's because of this:

```bash
RSVERSLATEST="${RSVERSONLINE//v/}"
```

should be this:

```bash
RSVERSLATEST="${RSVERSLATEST//v/}"
```

The variable `RSVERSONLINE` doesn't exist anywhere, so it's blank by default. Hope that helps :-)

<hr>

A fixed solution might look like:

```bash
function autoBumpReShade {
	echo "HELLLLLLLOOOOOO"
	RSVERSLATEST="$( fetchGitHubTags "$RESHADEPROJURL" "1" )"
	RSVERSLATEST="${RSVERSONLINE//v/}"
	echo " latest reshade version is '$RSVERSLATEST'"
	echo " current reshade version is '$RSVERS'"
	if [ "$AUTOBUMPRESHADE" -eq 1 ] && [[ "$RSVERS" < "$RSVERSLATEST" ]]; then  # Note the double brackets when comparing strings
		echo "updating reshade"
		writelog "INFO" "${FUNCNAME[0]} - Found newer version of '$RESH' - Updating '$RSVERS' to '$RSVERSLATEST"
		touch "$FUPDATE" 
		updateConfigEntry "RSVERS" "$RSVERSLATEST" "$STLDEFGLOBALCFG"
	else
		echo "not updating reshade"
		writelog "SKIP" "${FUNCNAME[0]} - '$RSVERS' is the latest version of '$RESH' - not updating"
	fi
    # any extra code can go here...
fi
```

zany130:
> Ahh, it's because of this:
> 
> ```shell
> RSVERSLATEST="${RSVERSONLINE//v/}"
> ```
> 
> should be this:
> 
> ```shell
> RSVERSLATEST="${RSVERSLATEST//v/}"
> ```
> 
> The variable `RSVERSONLINE` doesn't exist anywhere, so it's blank by default. Hope that helps :-)

image

yup, to be honest, I wrote this function late at night a few days ago and haven't even looked at it since

works now :sweat_smile: 

going to push it in a bit I'm not sure about the placement of the call to `autoBumpReShade` currently, it is in the main menu maybe I should put it in the global menu? or in prepare launch, as that's the only way it will get updated if the user hits save and play? as it is right now if the user doesn't go back to the main menu it won't run the function

sonic2kk:
> yup, to be honest, I wrote this function late at night a few days ago and haven't even looked at it since

I've been here many times (even recently), it happens :smile: 

> going to push it in a bit I'm not sure about the placement of the call to autoBumpReShade currently, it is in the main menu maybe I should put it in the global menu? or in prepare launch, as that's the only way it will get updated if the user hits save and play? as it is right now if the user doesn't go back to the main menu it won't run the function

Ah good thinking. I think a call to this in `prepareLaunch` is probably a good idea - A call to `autoBumpReShade` in the main menu and `prepareLaunch` should be good. That way we cover the bases you mention: When a user presses "Save" and goes back to the main menu it'll auto bump, and if they press "save and play" it will a auto bump (or when they launch SteamTinkerLaunch again, it will try to autobump the ReShade version).

<hr>

Awesome work here!

sonic2kk (approved):

sonic2kk:
The new function and function call placement looks good to me. I think GitHub has an option to squash commits but if you'd like you can squash them locally and force-push so that there's just one single commit for this :smiley: After that I'm good to merge this if you're ready

zany130:
I think that did it? not sure. Sorry still learning GitHub :sweat:

sonic2kk:
I don't think it did it but it's no problem, I'll use GitHub's squash feature (it should work, I just know some projects shy away from it).

sonic2kk:
Yup looks like that merged in fine - Thanks for all your work here! :heart:

zany130:
sorry small bug I just notice @sonic2kk  `echo "## $AUTOBUMPRESHADE" ` needs to be `echo "## $DESC_AUTOBUMPRESHADE"` on line 3027

sonic2kk:
Fixed it :-)

Changelog has also been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog#changes-in-latest-master (feel free to edit if you want to make any changes :smiley:)

sonic2kk:
Oh something that totally slipped my mind, @zany130 If you would like you can update the ReShade wiki page: https://github.com/sonic2kk/steamtinkerlaunch/wiki/ReShade

Feel free to update as much or as little as you want on this page, or if you don't want to, I can do it another time. There is no rush on this though :-)

Issue #598, opened 2022-09-23
⚠ MO2 - Mod Organizer 2 overhaul discussion 


frostworx:
This issue is meant to be a generic brainstorming issue for a possible MO2 overhaul.
Feel free to contribute any ideas you have to improve it. Of course PRs are very appreciated as well as usual.

Starting with this proposal:
If dotnet is no longer required at all for MO2 _(can someone confirm it is no longer required for modding?)_,
it might make sense to simplify the whole implementation by using one global "install directory" (=MO2 Program Files dir)
and f.e. symlink it into the corresponding compatdata without any further installation.
Imho this would help massively to solve almost all problems people have with the confusing "gamemode" and "standalone modes".

sonic2kk:
DotNet was removed a while ago, right? I modded Oblivion and New Vegas around June-July time and didn't really have any problems (using Proton 7.0, forget what I used for Oblivion but I used 7.0-4 for New Vegas). So in my own experience, DotNet shouldn't be required anymore. Other users have reported modding Oblivion with success too on the issue discussing the recent MO2 download patch.

Maybe any work and testing for this could be done on a separate branch? That can get... Tedious, but to preserve existing mod installs and get feedback having a separate branch once any kind of work starts on this might be nice.

Just my thoughts though, I've only modded 3 games that use similar engines and modding processes. It's very possible there are problems in other areas I haven't encountered. Though I will say the recent link to the Rockerbacon(?) ModOrganizer 2 project mentions Wine having "built-in support" which I assume means that MO2 should be fully compatible with Wine. But this project seems to mainly focus on Bethesda games, which is the extent of my experience too.

I would guess that MO2 might not require DotNet but *some* of the other GUI applications that you can launch via MO2 for the likes of Skyrim character editing **_might_** need DotNet. I couldn't say for sure though.

<hr>

Just realised the tone of my reply might not get it across (sorry!) but I think being able to clear up any confusion here is a great idea :)

frostworx:
yeah we recently discussed this already, but I haven't found the corresponding issue anymore to refer it here (haven't searched for long and assumed you'd reply here anyway again. thanks!)
You would not beliebe how much time I spent with testing and writing all that MO2 stuff (without actually using it...).
It once worked fine but apparently it doesn't anymore...
It would be really "nice" if people interested in MO2 would take part in a discussion here and contribute what they know.
[...]

PR #687, opened 2022-12-11
[WIP] Python port

Lightfire228:
I don't expect this to be merged.  This is just here for reference


src_python/v0.1.1/config/env.py
```diff
@@ -0,0 +1,31 @@
+import os
+
+from pathlib import Path
+
+
+def read_env_var(name: str, default=None):
+    return os.environ.get(name, default)
+
+# I can't decipher how STL bash detects steam deck, so I'm using this instead
```
sonic2kk, line 9: It checks if the file `/etc/os-release` exists and if `steamdeck` is present in that file. Some distributions (i.e. Fedora) use `lsb-release`, but Arch-based distros including SteamOS will always use `os-release` (at least afaik right now), so we use this.

Not sure if there's a performant `grep` alternative in Python aside from checking a file manually. This is part of what I meant about Bash being mostly suitable for STL :-)

sonic2kk:
Looking spiffy! Indeed, this probably be merged but it's no problem to have an open draft PR. Definitely neat to see how SteamTinkerLaunch could look if it was written in Python.

Just for reference, one of the things writing this in Python might benefit from is being able to use the `[steam](https://pypi.org/project/steam/)` and `[vdf](https://pypi.org/project/vdf/)` packages.

However, before you get too far into this, there may be various limitations here with the Steam Linux Runtime sandbox and running an actual Python program as a Steam compatibility tool. See [this comment](https://github.com/sonic2kk/steamtinkerlaunch/issues/598#issuecomment-1345205273) from Frostworx, the original creator of SteamTinkerLaunch. There may be various things to consider here. I would highly recommend reading the last few comments on that issue discussing why SteamTinkerLaunch uses Bash.

A Python implementation would be, honestly, an entirely different program. The Flatpak would need to be changed entirely, Steam Deck support would probably not work very well (it's not feasible manage Python versions and dependencies from source on Steam Deck), and all package maintainers would have to migrate.

So it's a cool experiment for sure, but it is not practical for many reasons. It may also take a very long time to port some functionality such as Vortex and MO2 support. A lot of SteamTinkerLaunch is working with files and running commands, which Bash is great for. Python can do it but not as cleanly as Bash can.

That's just my two-cents here. Be sure to have fun though! That's what programming is all about :-)

sonic2kk:
Also cc @frostworx, in case you're interested in what's going on here :-)

Lightfire228:
@sonic2kk 
> Not sure if there's a performant grep alternative in Python aside from checking a file manually.

You keep mentioning performance.  Since I'm not too familiar with the codebase I don't yet know the "functional" requirements (I'm translating STL function-for-function into python atm), but my assumption was that since this is just an "automation" tool for various setups and configurations, it doesn't *need* to be super performant.

Am I wrong about this?

---

Also if performance, and the Steam Linux Runtime, are a concern, then perhaps a rewrite in C/C++ or Rust would be preferable (since they run so much more "bare metal")?

I've not got much non-memory managed experience, but I've been looking at rust a little.  (I much prefer rust over C/C++)

Or perhaps a split language approach.  Write the scripty things in python, and the demanding things in rust

sonic2kk:
> it doesn't need to be super performant.

It doesn't need to me, but what I meant was that `grep` *can* be pretty dang fast.

>  then perhaps a rewrite in C/C++ or Rust would be preferable

It's moreso the SLR here but a C++ implementation is more what Frostworx had in mind iirc. But Rust would probably work as well :-)

> Or perhaps a split language approach. Write the scripty things in python, and the demanding things in rust

That could certainly be an approach! And if it was going to happen, I would probably do it more as Bash for most things and Rust for others. However, for the foreseeable I am set on sticking with Bash. One of the problems with a rewrite is that the dependency structure changes entirely. Bash just relies on system libraries, whereas languages like Python have a standard library, and Rust has a package manager. Plus, these languages *themselves* are dependencies - Python being a particularly large dependency. Most Linux systems will come with 95% of the SteamTinkerLaunch dependencies out of the box and for users that use package managers (really just the AUR and Pacstall for now with STL), the hard dependencies are very light.

As you mention, STL is a Bash wrapper. It is a wrapper for *commands*, and Bash is one of the most popular and standard shells across distributions. What makes using Bash so convenient here is that SteamTinkerLaunch is just a wrapper for running shell commands. Adding Python or Rust adds an additional layer here that isn't strictly necessary, it's moving one layer up. Something to keep in mind is that STL is not only a GUI application but equally a command line program. This is more than possible with other languages _including_ Python but moving one layer up to turn what is, at its core, a shell utility program into a shell utility program implemented in Python seems... overkill to me.

Probably the main advantage of using Python or another language is to have a better frontend, but much of STL's code is intertwined with Yad for several reasons. Something that was brought up in another issue was, instead of changing the way STL is implemented, contributing upstream to Yad to improve it would be a better approach.

So I have no intent to port or merge ports of STL in other languages, I'm happy for now to stick with Bash and I hope I outlined my reasons for it well. Another user very recently offered to attempt to [rewrite STL using wxPython](https://github.com/sonic2kk/steamtinkerlaunch/issues/598#issuecomment-1344675486) and the consensus there as well was to stick with Bash and focus development efforts on improving Yad as well as the existing STL code, instead of spending time porting to another language unnecessarily.

sonic2kk:
Not sure if there's any interest left in this, so *closing* :-)

Feel free to re-open in future.

PR #702, opened 2022-12-29
HMM: Option to use '--force' to always install dotnet48 for game

sonic2kk:
Opened PR because this will have conflicts and I want to review further using the PR interface.


sonic2kk:
The `--force` option does not seem to help very much, Proton 5.0 seems to be the main way to successfully install mainly on Steam Deck, but it can also work without Proton 5.0. This branch has some nice log file naming cleanups though. I'll need to test to make sure it doesn't break a clean install (which it shouldn't) and then it can be merged.

sonic2kk:
The `--force` option needs to be added after the version. If no version is provided, `--force` cannot be used.

- Correct usage: `steamtinkerlaunch hmm start nightly --force`
- Incorrect usage: `steamtinkerlaunch hmm start --force`

This option will remain undocumented as it isn't really useful for anything other than debugging. However this PR adds some nice logging fixes and changes, so I want to merge it.

For whatever reason, dotnet48 is really fickle about installing correctly sometimes. This feature was originally envisioned to try and fix it, but Proton 5.0 seems to be the most surefire way to fix it. Falling back to Proton 5.0 is not really feasible as it may not be installed and a version of Proton could come out which also correctly installs dotnet48. Proton 5.0 also fails to launch HMM, and takes *dramatically* longer to install. The Proton 5.0 "fix", in my mind, is flimsy at best, so I am hesitant to update the installation process any further. Contributions on fixing it from any very kind user reading this in the future would be welcomed.

It is a shame that it has to be so convoluted, but once installed at least the other parts of the support seem to work well (automatic updating is pretty neat).

Doing one final test of a full install process with this PR to ensure that it still works and does not always use `--force` when it shouldn't (some parts of STL do use `--force` for Winetricks but as long as behaviour matches master when `--force` is not used, it is no big deal in my mind).

PR #705, opened 2023-01-01
D8VK Support

### LINKED ISSUE #653: [Feature Request] D8VK support
## System Information
- SteamTinkerLaunch version: v11.12.20221112-7
- Distribution: Arch Linux

## Feature Description
An idea I had the other day was to implement support for [D8VK](https://github.com/AlpyneDreams/d8vk), a fork of DXVK that translates Direct3D 8 calls to Vulkan. I am not sure how many games use this and would benefit from it (maybe some Sonic Adventure mods and really old games?), but I'd like to add a way to enable this per-game and toggle it on and off. I'm not sure specifically if these are games that D8VK helps support with, but [their wiki](https://github.com/AlpyneDreams/d8vk/wiki) has a list of games. 

The enabling part isn't a big deal, probably we can just have a checkbox and when checked and saved per-game, it'll fetch the latest Actions build just like we do for HedgeModManager using nightly.link - D8VK doesn't have any releases to download from. We can store these in a SteamTinkerLaunch downloads folder, and only redownload when the version we have locally is different from the online version.

Looking at [D8VK's setup script](https://github.com/AlpyneDreams/d8vk/blob/master/setup_d3d8.sh), it seems like it automatically backs up the two existing DLLs that it overwrites (`d3d8.dll` and `d3d9.dll`). It also offers a `--symlink` option, though I'm not sure if we want to use that since it means if STL is ever uninstalled, the DLLs will stop working since their original files will be deleted with STL.

The D8VK script also offers an uninstall option, which restores any existing backed up DLLs.

Implementation could be as straightforward as:

```bash
function installD8VK {
    fetchLatestD8VK  # Would handle the version checking and downloading, as well as checking if D8VK already exists in STLCFGDIR
    writelog "INFO" "${FUNCNAME[0]} - Install D8VK for '$AID' to prefix '$GPFX'"
    WINEPREFIX="$GPFX"
    "$D8VKSETUP" "install"
    writelog "INFO" "${FUNCNAME[0]} - Successfully installed D8VK"
}

function uninstallD8VK {
    writelog "INFO" "${FUNCNAME[0]} - Uninstalling D8VK for '$AID' to prefix '$GPFX'"
    WINEPREFIX="$GPFX"
    "$D8VKSETUP" "install"
    writelog "INFO" "${FUNCNAME[0]} - Successfully uninstalled D8VK"
}
```

Very loose draft here, there would need to be more checks for whether the prefix exists at that directory before installing D8VK and other details like that.

I am not sure when I will get a chance to look at this, contributions would be welcome!

sonic2kk:
Implements #653.

Just UI options and settings for now - No downloading logic yet.

- [ ] Download latest `USED8VK` when enabled
- [ ] Automatically fetch the latest version if `autoBumpD8VK` is enabled
- [ ] Uninstall from prefix if `USED8VK` is disabled
- [ ] Use Notifier
- [ ] Update langfiles


sonic2kk:
Note to self: One game that supposedly benefits from D8VK is TES 3: Morrowind. Will have to test.

sonic2kk:
(OOps, didn't mean to close)

D8VK seems to have "official" releases now: https://github.com/AlpyneDreams/d8vk/releases

It seems there is only one real release though. Unsure yet if we should use an official release or download the latest git. I am somewhat leaning towards just sticking with the latest git. I figure that's what a lot of people that use something "niche" like D8VK would use too. I suppose it could be a checkbox too, though since there is only one version of D8VK out at the moment and I am not sure how frequently they release, I'm not sure if it's worth it.

We will see. This feature is someone on the backburner for now as I have other things I'm working on for STL for now :-)

sonic2kk:
Seems like there is interest in getting this added to Winetricks, though STL could probably have the ability to download from git and keep D8VK more up-to-date. Not sure, we will see. This is still on the backburner :-)

PR #722, opened 2023-01-14
Major GameScope Additions

sonic2kk:
Adds a *lot* of new options to the GameScope menu, so many that it's now split up by headings and has a scrollable dialog.

Mainly adds the new HDR and VR options available in newer GameScope builds, as well as a few potentially useful debug options. A couple of new general options have also been added, such as a feature I now find very useful, `--force-grab-cursor`, which locks the cursor to the GameScope window.

A couple of options for "embedded mode" have also been added. This is when GameScope is ran from a TTY instead of nested as a compositor inside of the Desktop Environment's compositor e.g., KWin. This is how Game Mode on Steam Deck works, but it is also useful to do this for performance reasons on the desktop if you're running games at very high resolutions and framerates. Therefore adding these options could be beneficial for users.

Ever since initial support was added and even in the ~10 months ago that I added a bunch more options to GameScope, it has gotten really big and has a lot of neat options. A refactor of the menu was going to have to happen at some point in the future anyway.

image

There are a couple of pending changes I have to make before this is ready for merging:
- [x] Add checkboxes to enable/disable some features like custom cursor and custom SteamVR icon
- [x] Fix paths for cursor image/SteamVR icon not working when there's a path with spaces (if possible)
- [x] Test "backwards compatibility" with previous GameScope menu more thoroughly (only did a brief test and it seemed to work fine)
- [x] Get an updated wiki page ready, as these are significant changes.
- [x] Update langfiles

Should hopefully be merged soon.


sonic2kk:
Checkboxes added for custom cursor image, SteamVR icon and custom stats path. Untested for now.

sonic2kk:
Backwards compatibility works fine, tried enabling and disabling various options and switching between this branch and master.

sonic2kk:
Langfiles updated.

The only real pending work now is for the filepaths, which seem to break whenever there is a path with a space. Will investigate.

Insertion of GameScope launch options also needs some refinement for the newer options like `--default-touch-mode`, seems  they are sometimes inserted incorrectly

sonic2kk:
Fixed the issue with some of the new GameScope arguments being incorrectly appended.

From further investigation, it seems like the file pickers aren't working when you go out and back into the GameScope menu, regardless of the paths having spaces. More work is needed on this at least (I could live with the path spacing problem if push came to shove).

~~Also a tooltip is missing for the custom cursor checkbox, `GUI_GSENABLECUSTCUR` has no corresponding `DESC_GSENABLECUSTCUR` (though one is probably not needed, could probably just re-use this text as a tooltip).~~ Done.

~~Finally, some minor code cleanup is needed. There is some commented code that can be removed now.~~ Done.

Once I can address these points, the code should be ready for merge.

sonic2kk:
Hardcoding the path to a file makes the file picker widgets work as expected, f.e.:

```bash
--field="$GUI_GSCURSOR!$DESC_GSCURSOR ('GSCURSOR')":FL "/home/gaben/Documents/hl3_cursor.png" \
```

Will correctly point to that file.

So the issue is likely with however parameters like `GSCURSOR` are being passed to the widget. For some reason it isn't able to correctly read it, even in a file path without spaces.

I did a test and outputted what `GSCURSOR` is being set to on menu load, and it looks valid (unless the path has spaces):

```bash
# Custom cursor image (--cursor) -- File picker
if [ -z "$GSCURSOR" ]; then
	if ! grep -qw "\-\-cursor" <<< "$GAMESCOPE_ARGS"; then
		GSCURSOR=""
		GSENABLECUSTCUR="0"
	else
		GSCURSOR="$(tr ' ' '\n' <<< "$GAMESCOPE_ARGS" | grep -A1 "\-\-cursor" | tail -n1 | sed -r -e "s/\s/\\\ /g")" 
		GSENABLECUSTCUR="1"

		echo "Loaded cursor path is '$GSCURSOR'"
	fi
fi
```

The paths look correct here on the GameScope menu load.

Even hardcoding the `GSCURSOR` value in the conditional above doesn't get the file picker to see the cursor path correctly. It still shows up as `(none)`.

In the Main Menu under the gamescope arguments, the path is correctly passed as far as I can tell. I have only tested a couple of files but the file paths I did test are valid, and yet the file picker widget won't display the selection correctly.

sonic2kk:
Fixed one issue with the icons not loading, though they still won't load if the path has spaces.

I feel like that's going to be quite a problem regardless, because the gamescope arguments textbox in the Game Menu doesn't preserve backslashes or double quotes if you enter them and save them. So I'm not sure how a user is meant to enter a path with spaces.

sonic2kk:
If I cannot find a fix in the next few days I may just add a temporary note to the tooltip about the paths and merge this, I don't want to hold back the GameScope improvements for *too* long.

I did a quick test with a "regular" yad script and it kept the quotes, so I think this is not any kind of Yad limitation but an issue that I will have to track down separately :-)

sonic2kk:
Figured out the problem, working on a (slightly hacky) solution.

The problem boils down to the `tr ' ' '\n'` logic for splitting the flags, which has issues when a flag has a path argument with a space.

sonic2kk:
Okay, all cursor paths should work now.

They should get parsed correctly when the GameScope config is saved and loaded, as well as when passed as an option directly in the GameScope args line.

I fixed the quotes not displaying on the paths in the Game Menu GameScope args box by using single quotes, so now the args will look like `--cursor '/home/gaben/half life 3/cursor.png'`.

If a path is entered manually in the GameScope using double quotes, it could cause problems. However, I don't see this as a particularly common case and it may be more effort than it is worth to fix. So for now I will leave it.

Shellcheck also has no further complaints so I think this is ready for a merge :-) Docs will be updated overtime.

PR #724, opened 2023-01-18
Bump Dependencies on Steam Deck

### LINKED ISSUE #719: Dependencies on Steam Deck need to be Manually Updated
## System Information
- SteamTinkerLaunch version: v12.0
- Distribution: SteamOS 
- Installation Method: Manual 

## Issue Description

Dependencies on Steam Deck do not automatically update. This is not a major problem as a user can simply reinstall SteamTinkerLaunch and the issue should be resolved.

Dependency versions are usually not a problem, but some like `innoextract` need to be bumped when SteamOS gets rebased on a newer Arch snapshot. It seems like based on #716 that dependencies do not automatically update, which is something I never even considered before. Not sure how it slipped my mind.

I am not sure of the best way to handle updating these dependencies to be honest. For now, the fix path is to simply reinstall SteamTinkerLaunch clean by clearing out `~/stl` and reinstalling. Very straightforward (especially with ProtonUp-Qt), but not documented until now I guess and also not ideal.

## Logs
N/A.

sonic2kk:
Fixes #719 

Initial logic for bumping dependencies to the version SteamTinkerLaunch expects (e.g. bumping from 1.9-5 to 1.9-7). Entirely untested for now.

There is currently no logic for "bumping down", which a user may find useful if they switch between SteamOS versions.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v12.12.20230201-1 (steamdeck-autodeps-update)"
+PROGVERS="v12.12.20230201-2 (steamdeck-autodeps-update)"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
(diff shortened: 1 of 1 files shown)


sonic2kk:
I have been busy and in my free time I have been using my Steam Deck to play some games, so I haven't had time to test this on my Steam Deck yet. I'll try to find some time this weekend to test it out :-)

sonic2kk:
Finally have some time (woo January over! :sweat_smile:), taking a look at this again.

The version check was changed to a `!=` check which should accommodate "bumping down" versions. It may not really be used very much but if a user rolls back to an older commit of STL specifically to downgrade a dependency (or a stable, once this change is in a stable release) then it should not require manual intervention to downgrade a dependency.

I want to release a new minor version of STL soon (maybe v12.12? :wink:) with a handful of minor fixes and tweaks, and this is one change I would like it to have. Having this in a stable would be nice and allow a good "rollback" point in the event of dependency versions needing bumped for one version of SteamOS but not another.

PR #727, opened 2023-01-27
Add DXVK_HDR checkbox

sonic2kk:
DXVK 2.1 now has support for HDR, which can be enabled with the `DXVK_HDR` environment variable (see [patch notes](https://github.com/doitsujin/dxvk/releases/tag/v2.1)). As the patch notes state, this is currently only compatible with GameScope as no major DE supports HDR yet, and also requires GPU drivers with the required Mesa patches (which some distributions like Nobara ship with).

SteamTinkerLaunch's GameScope menu has support for enabling HDR now though, so this change should go hand-in-hand with it. Users with a HDR10 display, a capable GPU, GameScope and a distribution with the relevant patches should be able to configure their HDR settings with SteamTinkerLaunch now :-)

This option works identically to other, similar DXVK environment variables, like DXVK Async.


sonic2kk:
Also, it's been over a year since my first contribution to SteamTinkerLaunch :partying_face: (#404, for #403)

PR #728, opened 2023-01-28
Update 'list' Command

sonic2kk:
Previously the list command only listed out a bunch of AppIDs, now in addition to listing the AppIDs it can also list the game name, game path, and all three.

This PR also does some minor updates to the `getGameDir` function. There is an internal change to allow only listing the path without listing the game name and AppID, an option to expose this to the user could be implemented fairly easily and will probably be done outside of this PR. The other update is a bugfix which caused soundtracks to be returned as not found, as `getGameDir` only checks `steamapps/common` in each library folder, whereas soundtracks are stored at `steamapps/music`.

There is an outstanding issue with getting the name from the App Manifest, which was implemented outside of this PR. xargs, which is used to strip trailing whitespace from the game names fetched from the App Manifest, doesn't like when game names have 'unmatched' single quotes in them, e.g. "Shantae and the Pirate's Curse". This will need to be addressed, but could be addressed in a separate PR as some other places which use the function to get a value from the App Manifest will need to be tested.

The help page needs updated to reflect the changes to the `list` function.


sonic2kk:
The `getgamedir` command was updated to allow passing an `only` option, which will only display the path to the game installation directory and will not display the `Game Name (AppID)` part.

PR #749, opened 2023-02-18
Fix ping logic, with valid shellcheck

### LINKED ISSUE #748: Ping logic update
Recent ping update changes the logic. instead it should be

```
! (ping -q -c1 archlinux.org &>/dev/null || ping -q -c1 google.com &>/dev/null)
```
we need to but the "not" symbol outside the "or" block to ensure we are logically requiring (logically equivalent to) both archlinux and google not being able to be pinged. See this simple test as a demonstration:

```
(1)(deck@steamdeck horizon_scripts)$ ! false || true
(deck@steamdeck horizon_scripts)$ echo $?
0
(deck@steamdeck horizon_scripts)$ ! true || false
(1)(deck@steamdeck horizon_scripts)$ echo $?
1
(deck@steamdeck horizon_scripts)$ ! (false || true)
(1)(deck@steamdeck horizon_scripts)$ echo $?
1
(deck@steamdeck horizon_scripts)$ ! (true || false)
(1)(deck@steamdeck horizon_scripts)$ echo $?
1

(deck@steamdeck horizon_scripts)$ 
```

This option passes shellcheck.

_Originally posted by @trentondyck in https://github.com/sonic2kk/steamtinkerlaunch/issues/743#issuecomment-1435697680_

trentondyck:
Fixes https://github.com/sonic2kk/steamtinkerlaunch/issues/748

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -22952,7 +22952,7 @@ function steamdedeckt {
 			notiShow "$NOTY_STEAMDECK_INSTALL" "X"
 			echo "$NOTY_STEAMDECK_INSTALL on Steam Deck"
 
-			if ! ping -q -c1 archlinux.org &>/dev/null || ping -q -c1 google.com &>/dev/null; then
+			if ! (ping -q -c1 archlinux.org &>/dev/null || ping -q -c1 google.com &>/dev/null); then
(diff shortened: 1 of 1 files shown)


sonic2kk:
Heh, beat me to it :-)

Thanks! Merging

PR #756, opened 2023-03-04
Modding: Update .desktop file generation

### LINKED ISSUE #637: Vortex & MO2 Not Handling NXM Links
## System Information
- SteamTinkerLaunch version: Latest GIT 
- Distribution: SteamOS 
- Installation Method: ProtonUp-Qt

## Issue Description

This is a combination of one issue, that the wiki says STL solves, but for me, it does not. But to be clear, the bug inherently lies in the fact that Vortex is running through a Wine layer.

I had to re-read the Wiki a few times to make sure I wasn't crazy, so this MIGHT be a bug with STL also.

> "create and update the nxm-protocol mime type for directly opening NXM with Vortex (renaming that from Mod-Organizer 2 when found)"

When Vortex launched for me the first time, there was an error and it closed. I opened it again, and Vortex asked me if I wanted to "repair" the installation, and I did. It asked me if I wanted to set Vortex to open NXM links, and I did. However, when clicking on "Download with Manager" on Nexus, Firefox prompted me to pick an application and nothing more. Nothing Vortex or STL related was selectable. 

I have been helping a friend, trying to get Vortex up and running so they can use the Collections feature, but even for them clicking the "Download with Manager" button did nothing. They are also using a Steam Deck with ProtonUp-Qt, and never reported errors for their install, but I will focus on mine for this report. I just wanted to make sure to note this, as this is why I started testing for it.

As for solving the issue: I presume that STL is running a script (like the other vortex and mo2 installers) to enable the NXM link manually. Assuming that the above bug report is valid, but the issue is not in relation to this script and the script works, it's more of a Feature Request to get this added to the GUI so that a user can have the script re-apply the NXM link. However, I am unaware if this is an issue with the script, or that it just needs to be run again, so I will Bug Report it just in case.

Also worth noting: In one MO2 testing instance for me (before STL), providing my API key manually worked flawlessly. For Vortex, we're not given an API key option as it's a tool by Nexus, for Nexus. They prompt users to log in to check things like Premium, endorse files, etc. Unsure if they store the API key locally, but they provide no way for the user to input this via the GUI. If they do store the API locally, it might be feasible to add an option to the STL GUI for a user to copy and paste their Nexus API key for it to be inserted manually by STL into Vortex. That is a passing thought though, I have not dug in far enough to see if they store it at all, or purely just use the login method dynamically. 

Side note: on triple checking my information and Launching Vortex again, it looks like it KEEPS asking me to repair the installation. Unsure if this is because of the Wine Layer and if I can Ignore it, or if I need to repair it. "Repairing" the installation might be why the NXM link no longer works.

image

## Logs
The only .log I have in the "dev/shm/steamtinkerlaunch/steamtinkerlaunch" path is winres.log. Says this twice:

> Sun Oct 23 05:00:50 PM PDT 2022 INFO - pollWinRes - Using GEOM '--geometry=1280x800+0+0' from '/home/deck/.config/steamtinkerlaunch/guicfgs/1280x800/22320/SteamTinkerLaunch-PrepareVortexGameStatus.conf'

sonic2kk:
This may improve/fix link handling, mainly for MO2 and Vortex, following discussion in #637. May also help HMM support. It makes the following changes to the `.desktop` files for each:
- Set `NoDisplay=false`
- Add `Hidden=false`

`.desktop` files may need to be re-created. We could probably add a command to re-generate them, something `steamtinkerlaunch mo2 <something>` (not sure what the command name should be just yet).

Looking at the code it seems there may be some logic to check if the files are up-to-date already, but adding the option would do no harm imo :-)

If all goes to plan this will go into the next release.

TODO:
- [ ] Test on Linux Desktop (from me and hopefully other users)
- [x] Test on Steam Deck (user tested, seems to work, will wait for more feedback though)
- [x] Set icon(?) if `.desktop` entry appears in application list
- [x] Update help screen


sonic2kk:
As suspected this creates a XDG menu entry. Probably for MO2 and Vortex we could just ~~start them normally with these files, e.g. if no URL is passed then we could call their respective start functions.~~ on second thought, I will instead log and show a notifier that this is not how you're meant to use MO2 and Vortex. I don't want to attempt to retroactively fit this kind of support it and then attempt to debug issues that will inevitably rise. It may be confusing at first but we can document this on the wiki, and the notifier should help too.

~~For HMM it isn't so straightforward. We could just log and show a notifier saying that this one should not be used.~~

sonic2kk:
Pushed a change that shows an error dialog if `steamtinkerlaunch <mod_tool> u` is called without a URL argument (i.e. if it's called from the `.desktop` file entry without a URL passed). This should inform users that this is not valid usage.

More testing is needed here to ensure that this PR is ready, I want to ensure there are no regressions. But this seems like it will at least massively help with the MO2/Vortex link downloading support, so I am hoping this will require minimal extra work.

~~The icon thing is going to be fun to figure out for Vortex/MO2 given that so their mime stuff is shared...~~ No it's not I can't read, should be straightforward if they come with icons.

sonic2kk:
Icon has been set to the STL icon as a compromise, as Vortex at least doesn't seem to come with an icon. Using the STL icon will also hopefully make it identifiable in the browser application menu thing, while also discouraging use from the application menu.

As well as this, the category has been changed from "Games" to "Utilities". Hopefully this also makes it clearer that these files are not to be used.

sonic2kk:
Might also be an idea to add/update the comments for these .desktop files to further emphasise that they should not be used.

sonic2kk:
Merge conflicts...

Issue #467, opened 2022-05-21
Add Steam-Metadata-Editor support


frostworx:
Recently I found the very useful tool [Steam-Metadata-Editor](https://github.com/tralph3/Steam-Metadata-Editor)
I think it would be nice feature to have direct support for it in steamtinkerlaunch.

frostworx:
corresponding [upstream issue](https://github.com/tralph3/Steam-Metadata-Editor/issues/18)

frostworx:
I will mostly stop working on steamtinkerlaunch soon and only occasionally add one or another feature.
This project is still interesting, so *leaving the issue open*

sonic2kk:
~~There has been no change on this upstream, but~~ this would still be pretty nifty to have, so leaving this issue open :smile: 

Upstream is still interested too, no ETA (though not planned for next release). Leaving this open until this feature is implemented :-)

sonic2kk:
I am not sure how I wanted to integrate this at the time. The maintainer of the project is really friendly, and I encourage anyone reading to check it out, but I think now there is no need for any integration :-)

Closing.

PR #757, opened 2023-03-05
Implement Artwork Selection

sonic2kk:
Adds a GUI and command to set the library artwork for a given Steam game based on a passed AppID. It allows for setting the Hero (banner), logo, boxart and tenfoot (recently played banner). It does this by taking some **absolute** file paths, and putting those files into the Steam Grid directory (`~/.local/share/Steam/userdata/<user_id>/config/grid`). By default it will `--copy` the files, but it can also `--link` or `--move` them.

If the artwork already exists in the grid folder, it will remove the existing files.

There is no validation on the AppID, as this would be a slow process (and there is currently no way to validate Non-Steam Game AppIDs as they don't come back for owned/installed AppIDs), so if it doesn't exist it will simply create

There is still some work to do on this, but this is the initial implementation for now. In future there could be potential integration with Steam-Metadata-Editor (#467).

image

TODO:
- [x] GUI Dropdown for selecting how to set art (copy/link/move)
- [x] Update langfiles
- [ ] Create wiki page for the F1 action


sonic2kk:
Fetching game artwork online may go in a separate PR, not sure yet.

sonic2kk:
We already have UI integration with SteamGridDB. We can just rely on this if a user wants to use that, it accomplishes basically the same thing except this PR is for local game grid art.

Seems ready to merge pending the wiki page.

sonic2kk:
Wiki page created: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Custom-Game-Artwork

PR #758, opened 2023-03-06
Steam Deck: Remove wget dependency

sonic2kk:
Seems like `wget` is now installed on Steam Deck, at least with relatively modern versions of SteamOS (3.3.X and above), and so is not installed by STL since it comes with the system (the check if the command exists before fetching already passed because it's part of SteamOS base).

Need to test this to ensure it works, hence the PR and not a direct push for such a minor change.


sonic2kk:
Tested on my Steam Deck with a completely fresh install of STL using this branch installed via ProtonUp-Qt, had no issues. I also confirmed that the same version of `wget` requested by SteamTinkerLaunch is available on SteamOS.

Just needs a version bump and a slim to ensure I removed all wget references and this is ready to merge.

sonic2kk:
`SDREPO` seems unused, according to shellcheck and verified with a Ctrl+F. Commented out for now in case it could be used in future for any other potential dependencies.

PR #771, opened 2023-03-30
MangoHud: Add option to pass --dlsym

sonic2kk:
This was one of the preset options in the config, however the option isn't meant to be passed as part of MANGOHUD_CONFIG, so this adds it as a separate menu item. When enabled, this passes `/path/to/mangohud --dlsym` to a game executable. **This option is only for OpenGL games which do not work out-of-the-box with MangoHud**.

Noticed this when I tried to play Torchlight II and couldn't get MangoHud to work.

Langfile text is not finalised yet (writing this at 1am :sleeping:) and I may also change the MangoHud "arguments" textbox text to more closely reflect that this is the Mangohud config textbox, as this was not clear to me. Should also update the presets in the dropdowns to add in some more useful preset configs (`--dlsym` was removed as this is not valid usage).

Tested this with Torchlight II and it worked, as this was the example which didn't work and prompted me to work on this (interestingly, native OpenGL performs better than Proton + DXVK as well as Zink).

Given how this hasn't been reported by now, and this option has been in place for I think almost a year, I guess that shows how often SteamTinkerLaunch + MangoHud is used with native games :sweat_smile: 

TODO
- [x] Update Mangohud argument dropdown presets
- [x] Update Mangohud argument text 
- [x] Update langfiles


sonic2kk:
The description text for the MangoHud config arguments was updated to reflect that they're for the MANGOHUD_CONFIG variable.

Other strings were not touched as the text in english.txt is not finalised just yet.

<hr>

MangoHud presets were updated to include some presets I thought were useful, based on the table on the [MangoHud readme](https://github.com/flightlessmango/MangoHud#environment-variables-mangohud_config-and-mangohud_configfile).

sonic2kk:
Updated langfiles for new MangoHud checkbox

Issue #772, opened 2023-03-30
Directx 11/3d acceleration error on steam deck


jmg3195:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: SteamOS Holo (OS version 3.4.6 stable channel)
- Installation Method: ProtonUp-Qt

## Issue Description
So I'm trying to launch a game trainer on the steam deck with STL. This process hasn't ever failed until the last 2 days. What's weird is both of the games I've had issues with work fine without STL. So I'm not sure what's causing the error or how to fix it. I spent the last 2 days combing through the internet for help and nothing I've tried has worked. This is the error the 2 games have shown (this error is specific to launching the actual game .exe):

"Failed to initialize player
Details:
Failed to initialize graphics.
Make sure you have DirectX 11 installed, have up to date drivers for your graphics card and have not disabled 3D acceleration in display settings. InitializeEngineGraphics failed"

## Logs
Attached is a screenshot of the error and 2 logs. 

The file titled "steamtinkerlaunchlog.txt" is the log located at:
/dev/shm/steamtinkerlaunch/steamtinkerlaunch.log

The file titled "steamtinkerlaunch.log.for.specific.game.txt" is the log located at:
~/.config/steamtinkerlaunch/logs/steamtinkerlaunch/title/${GameTitle}.log
![DirectX 11 3D acceleration error](https://user-images.githubusercontent.com/51762894/228916325-6bacb711-cf5c-4b1c-884c-433b78c6fddd.jpg)
[steamtinkerlaunchlog.txt](https://github.com/sonic2kk/steamtinkerlaunch/files/11114256/steamtinkerlaunchlog.txt)
[steamtinkerlaunch.log.for.specific.game.txt](https://github.com/sonic2kk/steamtinkerlaunch/files/11114257/steamtinkerlaunch.log.for.specific.game.txt)

jmg3195:
I just realized I only attached the logs for 1 of the 2 games. I would imagine that won't be a problem as the error on both games was exactly the same. If you do want the logs for the 2nd game I'm also happy to provide those :)

sonic2kk:
I don't think this is a SteamTinkerLaunch bug, though I'm not sure why it only happens with STL. I recall a similar issue reported before with #532 and #564, but these weren't STL bugs.

Does this error only appear when launching the game with STL even with the custom command disabled?

[...]

PR #773, opened 2023-03-30
Allow Steam Linux Runtime for Non-Steam Games

sonic2kk:
Allows the Steam Linux Runtime to be used for Non-Steam Games. May fix some instances where a Non-Steam Proton Game needs the Steam Linux Runtime to run.

Needs tested against native Linux Non-Steam games, as well as more thorough testing for Proton games. It could be useful for Non-Steam native games, but most that I have seen these days are AppImages. Still, older DRM free ones (i.e. Undertale) may still benefit here.

May help #772


sonic2kk:
This will not be merged until further testing is done at the very least, and until there is more discussion in #772 to clarify if this is needed to fix that instance.

Even if it is not needed I will test and merge this if there are no other glaring fires from merging it, but I will wait for the feedback on that issue out of a personal curiosity.

sonic2kk:
Was confirmed to fix a game not working, will test some more and merge soon.

sonic2kk:
Forgot about this, thought it was already merged too... Will resolve conflicts and merge now.

sonic2kk:
Apparently something here causes a game crash if no SLR is enabled.

sonic2kk:
Maybe not, tested again and didn't have any crashes with the same game. Other games worked fine too.

Issue #781, opened 2023-04-11
steamtinker takes several minutes to load or open a new page


saufall:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: nobara 37 newest update
- Installation Method: protonU

P-Qt

## Issue Description
Freshly installed stl from protonUP-Qt, set up a game in steam thats previously configured to run with proton to run under steam tinker launch did not clean proton prefix. it takes a long time (5-10 minutes) to load the game. After the first launch it gets a little better but it still takes 2-3 minutes for the launcher to pop up an 1-2 minutes to load windows.
#issues/317 seems related but I purged my stl settings multiple times and it still takes forever to load. it is especially serious when i quit the STL while it is waiting  to load a window . It immediately takes 10 minutes or so to load the STL at **all.
## Logs
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/11204149/steamtinkerlaunch.log)
[Skyrim Special Edition.log](https://github.com/sonic2kk/steamtinkerlaunch/files/11204163/Skyrim.Special.Edition.log)

sonic2kk:
Not sure how this could be caused or resolved. The issue you linked (#317) to does indeed seem to be related, but I don't have this issue myself, on my PC (root install), laptop or Steam Deck (both ProtonUp-Qt).

@zany130 Do you recall anything about the issue you opened a while ago? Maybe you could give some advice, though I appreciate that the issue in question was opened a ***very*** long time ago :-)

<hr>

I noticed it seems to get stuck initially when fetching Steam Deck compatibility information. This should only be fetched once but it may account for one single source of a slowdown. It might be worth testing STL while disconnected from any networks, and seeing if the issue persists. It is also possible that it just takes a long time to get to the point where it tries to load the main menu.

Apart from that, I can see in the log timestamps that it does take several minutes to load. But there is nothing I can see from this right now that suggests why it should be taking so long.

<hr>

Out of curiosity, are you using any slower drives (i.e. HDDs) on your machine that may be causing a bottleneck? All of the machines I have tested SteamTinkerLaunch on all use SSDs. My main PC has one HDD but I have no libraryfolders on it, so STL would not even be looking at it.

zany130:
Actually forgot about that issue so had to go look it up. I think that was the first issue i ever made here 😆


Anyway yeah i have had slow starts of STL every now and then.  But they always go away on there own.

[...]

Issue #634, opened 2022-10-21
Track Valve's Steam Deck compatibility rating


sonic2kk:
## System Information
- SteamTinkerLaunch version: STL git
- Distribution: Arch Linux

## Feature Description
It would be cool if Steam Deck compatibility was listed in various relevant places around the UI, such as in the wait requester and in the main menu, probably around where the ProtonDB information is shown. Since various pieces of information are hidden on Steam Deck we'll want to make sure that this is definitely shown, as its most useful on Steam Deck.

There can actually be a difference between the ProtonDB rating and the Steam Deck rating, which is the motivation for showing this. A game can work flawlessly on Steam Deck but not work on desktop Linux - Two recent examples of this are [Uncharted: Legacy of Thieves Collection](https://www.protondb.com/app/1659420) (needs CPU/GPU environment variables) and [Persona 5 Royal](https://www.protondb.com/app/1687950) (seems like there's a Wine call that can't be made on Linux desktop but can on Deck). On the flipside a game may be marked as verified or playable but have a poor ProtonDB rating.

I'm not sure if there's a good place to pull this from. ProtonDB is able to list it though its API doesn't provide it. I'm not sure where it gets it from. Unless I missed it I don't think SteamDB provides it and I'm not sure if the Steam Web API provides it. We couldn't use the web API anyway because it needs a token and we can't ship a token. And having a user provide their own would make this feature more effort to use than it would be worth.

I'll take a look and see if there's anywhere to get this info. If anyone knows where to possibly get this info, please leave a comment. It would be nice if we could have the compatibility criteria for each game too, so we could potentially list that too. If I can find somewhere we can safely pull this data from, anyone is free to contribute a pull request for this :smiley:

sonic2kk:
Seems like there's a Steam endpoint that can be used? Not sure what the potential restrictions around this are e.g. are we allowed to actually call this, but this is the GET endpoint: `https://store.steampowered.com/saleaction/ajaxgetdeckappcompatibilityreport?nAppID=<app_id>`.

This returns information (with placeholder strings and not actual strings) about the game's compatibility, e.g. `#SteamDeckVerified_TestResult_DefaultControllerConfigFullyFunctional`. Unsupported messages also have specific strings corresponding to things like misconfigured anti-cheat (_Halo: The Master Chief Collection_), unsupported anti-cheat (_FIFA 23_), and poor performance on Deck (_LEGO Incredibles_).

[...]

PR #782, opened 2023-04-11
Global Menu: Add toggle for Steam Deck compatibility information

sonic2kk:
May help #781.

Adds a checkbox to the global menu which controls whether or not to fetch Steam Deck compatibility information. This seems to be an outlier, as other options (ProtonDB rating, game info, etc) already have toggles here. Guess it slipped my mind when implementing it (#634, 6c64dc2833243ee640c1ab269e5eaa00a9c4059e).

Needs some further testing, and other options may also need options. Should investigate that too.

Also, I don't think this feature has a wiki page. Should add one.


sonic2kk:
Documented here: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Game-Data#steam-deck-compatibility

sonic2kk:
Made a change in this PR too which doesn't display the Steam Deck compatibility rating on the wait requester if we don't have it. The Main Menu already followed this too, so it's a little bit of consistency.

Also improved some of the logging too following the toggle added here. If we have `DLSTEAMDECKCOMPATINFO` off, it won't warn about `jq` potentially being missing when we don't have any Steam Deck compat info. We would expect to not have any compat info if this option is off.

Issue #660, opened 2022-11-14
Lutris and GE wine builds


zany130:
## System Information
- SteamTinkerLaunch version: v11.12.20221114-10
- Distribution: Garuda Linux

## Feature Description
[lutris](https://github.com/lutris/wine) and [GE](https://github.com/GloriousEggroll/wine-ge-custom) wine builds could be useful

sonic2kk:
Thanks :-) I said it on the other issue too but for completeness, I am fully in favour of adding these features. Sadly I just don't have time to work on it right now, but in future I am hoping to! It may not make it into v12 though

sonic2kk:
Something to note here is that some parts of SteamTinkerLaunch including Steam expect a specific directory structure including a Proton script, which afaik the GE-Wine and Lutris Wine builds do not have (since they are Wine builds). I also suspect they have slightly different directory structures, including GE-Wine which may have a different structure to GE-Proton.

I think the "solution" here is to do two things:
1. Document that using Wine builds with Steam is unsupported
2. Update the code to check for a Wine build and try to run with that. Some tools e.g. MO2, Vortex, HMM will expect the Proton script and they are set up to use it, so these tools should probably log a warning when a user tries to select a Wine version to run it with
3. Log a warning if a user tries to run a Steam game with a Wine build
4. Investigate writing these out to a separate CSV i.e. `WIneCSV.txt`, or maybe we can just include them in the regular CSV. The issue here is that some tools e.g. MO2 or HMM, and the Proton mismatch checker, may need updated if we write them all out to the same CSV. There is code which looks for `GE` in the `ProtonCSV.txt` file in order to resolve version mismatches. This may need to be changed, maybe setting it to `GE-Proton` is enough (though I wonder how this would impact older GE-Proton versions...)

The merit in having support for the Wine builds is mainly for custom commands, or at least that's how i see it. I think it would be handy to be able to run a custom command with a GE-Wine build for example. And also, in general, the ability to download these tools would be handy too.

A potential implementation would be to allow downloading these versions, writing them out to a separate CSV file and then showing them in a separate dropdown under "Wine Options". This would override the system Wine and of course the regular Proton version. I also like this approach because it means Wine options won't be listed under "Proton version".

This will unfortunately almost certainly miss v12, but it is something I would like to see added at some point!

sonic2kk:
[...]

PR #783, opened 2023-04-12
Wine: Add Lutris Wine and Wine-GE as available Wine versions

sonic2kk:
Implements #660.

Adds Lutris Wine and Wine-GE builds to the Download Custom Wine menu. Appears to work in that it downloads and extracts Lutris Wine and Wine-GE, but game compatibility will of course vary. Cookie Clicker got to the language menu and then crashed after selecting a language, for example (with SLR on).

The Steam Linux Runtime is enabled/disabled the same way as for native/Proton games for consistency.

This option is probably most useful for Non-Steam Games running through SteamTinkerLaunch, where Proton isn't really meant to be used in the first place.

This menu could use a further overhaul. I have noted some ways it could be improved, I may open an issue for it too.

TODO:
- [ ] Update wiki


sonic2kk:
Currently this should only need some light further testing. I will probably merge in a few days if no issues come up.

sonic2kk:
Should be ready to be merged. Lutris Wine and Wine-GE use the same archive format (Wine-GE uses the same build system afaik) and they seem to work in my tests. Compatibility can vary, and it may be worth allowing Wine versions for One-Time Run as well, but these are further enhancements that can be made in future.

PR #786, opened 2023-04-13
One-Time Run: Allow Command-Line Usage

sonic2kk:
The goal of this PR is to overhaul the One-Time Run option to use a single function for executing One-Time commands. This would allow for a more streamlined GUI function, as well as command-line usage.

Currently in this PR command line usage is as follows:

```sh
steamtinkerlaunch otr 970570 --exe="/path/to/exe" --proton="GE-Proton7-53" --useexedir --args="-funnyarg"
```

In light testing, this appears to be working the same as on master, but I need to test all options thoroughly to ensure no regressions. 

The command line function is called from the GUI and the arguments passed to it are based on what is selected on the UI. For example: `commandlineOneTimeRun "$1" --exe="$OTEXE" --proton="$OTPROTON" --args="$OTARGS" --workingdir="$OTCUSTWORKDIR"` - This allows for better re-use and matches some other parts of the codebase, such as the game art code.

The code enabling this currently is also a little messy, especially for the GUI function where we have a bunch of checks for which argument to pass. I wonder if there's a way to do this with some kind of argument building function. I will test this once further testing is done and once I know the splitting of the logic is working fine.

This PR is currently marked as draft as this needs ***extensive*** testing to ensure all options still work, and that no regressions are present. One-Time Run should function identically to before.

<hr>

I would also, at some point, like to implement the following improvements to One-Time Run, but they may be out-of-scope for this PR:
- Native Linux executable support/detection (with option to force Proton like we have on the Game Menu)
- Support for using Wine versions (maybe as a separate checkbox/dropdown)
- Steam Linux Runtime support

<hr>

Executables currently tested with this PR:
- ROBOTICS;NOTES ELITE Improvement Patch by Committee of Zero
- Sonic P-06
- HoloCure

<hr>

This PR also ended up fixing some bugs:
- Fixed One-Time Run arguments and custom working directory order being incorrect 
- Fixed Save function not saving some arguments

<hr>

TODO:
- [x] Investigate improving logic
- [x] Extensive further testing of all options (especially "save")
- [x] Update help screen
- [x] Update langfiles
- [ ] Update wiki


sonic2kk:
Appears to work in light testing.

Will do some practical testing with a Committee of Zero game patch.

sonic2kk:
~~ROBOTICS;NOTES ELITE Committee of Zero patch installer did run and transfer game content, but the game crashes when trying to boot. This does not happen when using master.~~

After a successful install from master, then uninstall, all subsequent re-installs (from master or this PR) appear to fail. Testing a fresh install of ROBOTICS;NOTES ELITE with this PR to see if that works. It could be that uninstalling just does not work for some reason.

<hr>

Since the parameters needed to run this command are working, that knocks a few of the options needed to be tested off the list.
- Executable
- Proton version
- Exe dir as working dir

Still to be tested are:
- Arguments (expecting issues with this one tbh)
- Custom working directory
- Save option

sonic2kk:
Seems like the issue was just uninstalling. Re-installing the entire game into a fresh folder and then using this PR to install the patch appears to work.

On master, with a fresh ROBOTICS;NOTES ELITE install, the patch works. Once the patched is uninstalled and re-installed, the game fails to boot. This happens on v12.12, master, and this PR, so I think this is just general behaviour and not something caused by this PR.

One-Time Arguments need tested primarily next, as well as the custom working directory. I will also test the save option, though I am not so sure how it will work. I am mostly concerned with the first two.

sonic2kk:
Save is not bound to the right value from the GUI

<hr>

It seems a few of the values for One-Time Run were not correct. The Save value and custom working directory do not appear to have been set correctly...

I suppose this PR will address that too. Nice!

sonic2kk:
Order of args is now correct, and the save function has been further updated to work with the custom working dir and exe dir options. Neat!

<hr>

These appear to work from the GUI. The only remaining option to test from the GUI is the args. I think they will work now, but I am not sure.

Since the GUI uses the command line function internally, this should mean the command line code works too. But I will also give it a quick test once the GUI is confirmed to be working.

<hr>

Once all functionality has been confirmed working, I will move onto code cleanup.

sonic2kk:
Would be pretty great to have the option of running based on the saved parameters. These are stored in the config file, we should try to read them if something like `--auto` is passed (I originally considered `--saved`, but this may be too close to `--save`).

<hr>

Seems like this will require quite a refactor.

sonic2kk:
Was busy with other projects, back to take a look at this.

sonic2kk:
Appears to still work fine, just installed STEINS;GATE Steam patch. It had an issue where it couldn't run some specific file but that happens on master and 12.12 too, so it probably isn't an STL bug.

sonic2kk:
Tested, help screen updated, this appears to work fine in testing.

If issues crop up I'll fix them or if users report issues I can investigate.

Wiki will be updated after merging.

Issue #788, opened 2023-04-14
One-Time Run Improvements


sonic2kk:
## System Information
- SteamTinkerLaunch version: git
- Distribution: Arch Linux

## Feature Description
There are various improvements I want to implement for the One-Time Run dialog. As there are several, I feel it would be useful to list them here.

When features from this list are implemented, I will add their PR hash into the subheading. I am currently working on an overhaul of how One-Time Run is written in #786 to allow for the first item in this list, command-line usage, so further changes should probably wait until that is merged, if at all (hopefully it will be though :wink:)

UPDATE 11/06/23 - Out of nowhere, One-Time Run now seems to require `STEAM_COMPAT_CLIENT_INSTALL_PATH="$SROOT"` to be set. MO2, Vortex, and custom commands still appear to work fine. This was fixed as part of #833.

### Command-line usage for One-Time Run (#786) 

Allowing full One-Time Run usage from the command line would be pretty sweet. It *is* currently possible to use `steamtinkerlaunch otr` but it isn't documented on the help page. More than that, it just opens the GUI. I want the ability to fully run something from the command line with STL.

This would be pretty handy for automation as well. An option to execute the One-Time Run command based on saved values would also be pretty useful I think as well, if enough are saved. That way automated scripts could be implemented, and it could be useful for #625, if/when that gets implemented.

### Defaults Button (#833)
Reset any saved parameters for the One-Time Run GUI by writing out blank values to the game config. There is currently no straightforward way to "clear" values from the frontend, this is especially a pain for the working directory dropdown. This option should also clear any "saved" values.

This should be fairly straightforward to add.

### Native Linux program support (#801)
One-Time Run only supports running with Proton at the moment. It should be possible to check when we are given a native Linux binary, and to ignore the Proton version. Language files will probably need updated slightly to reflect this, off the top of my head it would be nice of the Proton dropdown mentioned explicitly that this will be ignored for native Linux programs.

Custom commands already have some logic to differentiate between a native Linux binary and a Windows binary, so we should be able to reference that implementation for this feature in the One-Time Run logic.

### Force Proton for Custom Programs (#801)
Once the above native Linux binary support is implemented, there may sometimes be issues. The logic for differentiating Windows and Linux programs isn't perfect, as shown by #710. It seems Windows binaries built on Linux (apparently a rarity, as only one user has reported the problem :sweat_smile:) report different information and there is no solid way of telling them apart

[...]

PR #801, opened 2023-05-13
One-Time Run: Add support for Native Linux executables

sonic2kk:
Part of work for #788.

This should include scripts, AppImages, native binaries, etc.

The logic used to detect native Linux executables is the same as for custom commands, which means that it's not fool-proof. For custom commands, we have an option to force Proton with custom executables, should the check be incorrect. Currently one time run does not have support for this, but it is planned to be added before this feature is merged.

The other remaining issue is that we still require a STEAM_COMPAT_DATA_PATH, as the setOneTimeRunVars function still looks for this, and we rely on having this for Proton games. Instead, we need to change the logic to return a warning instead of an error, and then only check for this path if we're using a Wine executable.

Currently this PR has been tested to ensure GUI and command-line functionality still works for Proton games, as well as testing native Linux games from the GUI and commandline:
- Committee of Zero STEINS;GATE Improvement Patch installer (Proton)
- Touhou Fumo Racing (Proton)
- Shellcheck (Native)
- SuperTuxKart (Native)
- OpenRGB (AppImage)

<hr>

There is also still currently no option to use the Steam Linux Runtime for One-Time Run. This would require some additional work as the logic for detecting the SLR is not very "portable", and so should be split out somehow. Another reason is that some of the new One-Time Run logic could be improved or simplified (there is a lot of nesting which annoys me), adding another conditional to check for and append the SLR might get messy.

So the SLR option will come in probably two separate PRs: One to split the SLR logic and make it more portable, and another to add SLR functionality to One-Time Run.

<hr>

TODO:
- [x] Allow One-Time Run menu for native Linux games (currently can't use One-Time Run for games which don't have a `STEAM_COMPAT_DATA_PATH`)
- [x] Add option to force Proton with an executable, overriding STL's detection (just like we have for custom commands)
- [x] Update langfiles to note that OTR Proton will be ignored for native Linux executables, and to note that native Linux executables can be used 
- [x] Sync langfiles


sonic2kk:
A command line option, `--forceproton`, has been added, but this is not configurable from the One-Time Run GUI. This still needs to be added, along with the ability to save this option.

Adding to the GUI might require a refactor of how we pass arguments. It should be feasible, but will just take a little bit of work (and a refactor of that code would be nice anyway).

sonic2kk:
Good progress so far:
- Refactor of how we pass arguments from the GUI to `commandlineOneTimeRun` has been completed, removing the nasty conditional logic that was in place. Now we build an array of the flags we want to pass, and pass that expanded array to `commandlineOneTimeRun`. This makes it much more flexible to add commandline options as we can just add one if statement, instead of having a whole bunch of repeated and nested `if`s.
- Add checkbox to One-Time Run GUI to force Proton even if SteamTinkerLaunch detects the executable as a native Linux executable. This is basically the same behaviour as we have for custom commands in the Game Menu.
- Update **`english.txt` langfile only for now** to reflect the new code changes (passed executable can be Windows or Linux, Proton version is ignored for Windows games, add label and description for new force Proton checkbox).

The remaining work is to allow One-Time Run for native Linux games running with STL, and then to update the remaining langfiles. Allowing One-Time Run when using the GUI with a native Linux game might be a bit tricky, as the current logic expects `STEAM_COMPAT_DATA_PATH` to be defined (and that won't be defined for native games). The Winetricks and Winecfg buttons also expect this, creating further problems. We'll need some way to maybe show a pop-up or something in these cases to make it clear to users that Proton options are not available when using native Linux for OTR (there's no prefix defined to run Wine inside of, and we should not assume the user has Wine on their system).

sonic2kk:
One-Time Run can now be used with native Linux games. When `ISGAME` -eq 3 we can ignore `STEAM_COMPAT_DATA_PATH` being undefined. Of course this means Winecfg and Winetricks will be unavailable, but we can document this on the wiki. A UI overhaul with a description could also add a section that notes that Proton options will be ignored when using native Linux custom commands or OTR when using STL with native Linux games.

Now this PR just requires further testing, synced langfiles and a version bump.

sonic2kk:
Langfiles are updated. PR needs some further testing and then a version bump, then ready to merge.

sonic2kk:
Dammit, `ISGAME` is always `3` when running OTR outside of Steam. I guess this check won't work...

The reason this doesn't work is that `ISGAME` is set to `3` is STL is being ran from the script.

sonic2kk:
Removed the `ISGAME` check, and instead we simply warn if `STEAM_COMPAT_DATA_DIR` is not defined. Now we only enforce having a `STEAM_COMPAT_DATA_DIR` if `$ISWINDOWSEXE -eq 1` (if we detect a Windows executable or if the user forces Proton).

I re-tested with the games from the OP with this change and things appear to still work, including with the mix of options (STEINS;GATE Improvement Patch is a good test candidate because it requires `--isexedir`).

This PR should be ready to merge now.

sonic2kk:
Forgot to apply these changes to the `OneTimeRunGui` function... Gotten too used to the commandline OTR :sweat_smile:

sonic2kk:
To recap the changes for anyone reading this PR (including future me):
- Add some logic to tell when we're using native Linux executables. The Proton script's `run` option may work for most executables, but replacing it with an explicit executable launch is cleaner and allows for Linux executables to run even if we don't have a Proton version.
- Since we now have logic to determine if we're running a Windows binary or not (the same logic used by custom commands), we add an option to force Proton in case we get this check wrong. Some Windows executables that are built on Linux return a different output from `file`, but this appears to be a rare case (only one issue report here), so we stick to our logic and then if we get it wrong we give the user an option to force Proton.
- A handful of minor changes (such as ensuring the Wineprefix we select exists in `setOneTimeRunVars`, sometimes Proton prefixes for Windows releases of native games were detected even after the prefix was deleted)
- Finally, we add support for using One-Time Run with native Linux games. Since native games don't have a Proton prefix (unless the Windows release was previously used and the prefix was not cleaned up), One-Time Run would fail because it was expecting a `STEAM_COMPAT_DATA_PATH` variable (used to define where a game's compatdata is). The purpose of One-Time Run is mainly to run an executable in a game's prefix, but not allowing it for native games is a bit annoying to me, so I decided to allow it with the caveat that only Linux executables (which were not supported until this PR) can be used. One-Time Run is generally useful for giving a command/UI to run executables from (Windows *and* Linux, i.e. AppImages) without having to do it from a separate UI. It is just, in my opinion, a nice little convenience.

PR #802, opened 2023-05-14
Vortex: Gate Pre-Release Installation Behind Checkbox

sonic2kk:
## Overview
Currently SteamTinkerLaunch always defaults to downloading the latest Vortex release, even if that is a pre-release. This may not be desirable for any number of reasons, including stability, and that I believe it defaults users to the "Beta" release channel (which is not a good idea under Wine).

There was a checkbox existing already on the Global Menu to use Vortex pre-releases, but this checkbox does nothing. In fact, the tooltip was not even properly displayed as the variable used to set the description was incorrect (missing an underscore in `DESC_USEVORTEXPRERELEASE`).

This PR adds functionality to this checkbox. We check if this checkbox is enabled and if it is, we pass a new optional parameter to `getLatestGitHubExeVer`. This parameter controls the URL that we send to `wget` in order to `grep` for the latest tag.

## Implementation
When we're fetching the latest regardless of whether it's stable or pre-release, we grep the `https://github.com/orguser/project/releases/tags` URL and get the first link on this page which matches a set tags regex, which in the case of MO2, Vortex, and HMM, points to the latest release.

This PR changes the logic so that if we want the latest stable, we `wget` the url `https://github.com/orguser/project/releases/latest`, which will always point to the ***latest stable tagged release***. For Vortex, currently `1.8.1` is the latest stable and `1.8.2` is the latest pre-release. The `releases/latest` page will point to `1.8.1`, but the `releases/tags` URL will have a list of *all* release tags in descending order, *including* pre-release tags. The regex we use to get the first matching URL that has a release version works for the `releases/tags` and `releases/latest` page, so we only need to change the value of `TAGSURL` based on whether `EXCLUDEPRERELEASES` (the checkbox value) is true (`1`) or not. The rest of `getLatestGitHubExeVer` should remain unchanged. And since this is an optional parameter, there should be no breaking changes to this function as a result of this PR.

This will not retroactively impact Vortex installations, it will only apply to fresh installations.

## Adjacent Work
This PR is the beginning of some work to try and improve stability under Vortex, given the recent breakage under Wine requiring newer GE installations. There are two pieces of future work that will add to the usefulness of this PR:
- Investigate removing .NET installation logic, as Vortex 1.8 should be able to install .NET itself now(?)
- Add option to disable Vortex auto updates by setting `setVortSet "settings.update.channel=\"\\\"$NEXUPDATECHANNEL\\\"\""`, which should allow users to stay on the current stable release that they install. If this option is disabled, Vortex should default to pre-release if the checkbox option from this PR is enabled, and otherwise revert to stable.

<hr>

TODO:
- [x] Test actual Vortex installation, only tested downloading the exe for now with `steamtinkerlaunch vortex download`

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230509-2"
+PROGVERS="v14.0.20230514-1 (vtx-stable-pre-logic)"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Vortex installation worked, installed 1.8.1 by default and 1.8.2 after enabling Pre-Releases and re-installing.

sonic2kk:
In future, a complimentary change to go along with this work could be to add an option to set a Vortex version to look for, and then fall back to stable or pre-release depending on what the default would've been. It should be possible to modify `getLatestGitHubExeVer` to get a specific tag, or if not or too complex, a new function could be created.

PR #803, opened 2023-05-14
Vortex: Remove dotnet48 installation; use Winetricks for dotnetdesktop6

sonic2kk:
Vortex 1.8.0 and above no longer require dotnet48 to install, however Vortex still requires .NET 6 or above. The manual .NET 6 installation works, but Winetricks should be able to install .NET 6 for us, so we should replace the manual code. This was not the case when this logic was written, as Winetricks could not install dotnet6 at that time.

The relevant Winetrick we want is `dotnetdesktop6`. I am not sure if this is available in the latest stable Winetricks, but as all users should know, an up-to-date Winetricks should always be used when tinkering. A note will be added to the Vortex wiki around this.

Currently Vortex will work with this PR, but it'll tell you that you need .NET 6 or above. It does offer a fix, and the fix *will* correctly install .NET 6, but I would still rather automate this step as it should be relatively cheap. Also, the .NET 6 install logic was in the `startVortex` section. I guess this was because it was needed *after* Vortex was installed and set up, but now it can probably go in the `installVortex` section.

This should also significantly reduce the footprint that Vortex has, down from about 3.5gb to 1.2gb on my PC. Installation also seems to be around a minute or so faster (dotnetdesktop6 installs *very* quickly compared to dotnet48).

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230516-2 (vtx-dotnet-freedom)"
+PROGVERS="v14.0.20230516-2"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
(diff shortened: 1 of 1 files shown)


sonic2kk:
This appears to work for Vortex 1.8.1 and Vortex 1.8.2 Pre-Release, so this should be good to merge.

Issue #625, opened 2022-10-06
Feature Request: More Custom Command Control


sonic2kk:
## System Information
- SteamTinkerLaunch version: v11.12.20220924
- Distribution: Arch Linux

### Background
I had an idea today while working through #624. That feature request was to add support for a specific mod manager to make setup easier. As it turns out, that mod manager pretty much only needs a winetrick or few. But because of how mod managers work, each game's prefix would need that winetrick. This got me thinking about ways to drastically improve and change how custom commands are handled with STL.

### Problem
Some custom commands such as those related to modding, require tweaks and winetricks to get working. A user is very much capable of doing this with STL with its Winetricks menu, but for standalone executable programs that need to be selected on a per-game basis, this can get really tiresome. Having to go in and choose the winetrick each time, plus a user might forget if it's been a while!

There is also the case where a user might want a custom program installed to a specific prefix, either with or without Winetricks, unrelated to the current game's prefix. Say Cheat Engine for example - A user might want to download and install Cheat Engine once, into one prefix, and set that as a custom program for each game. Well as far as I know currently they **can't choose a custom prefix for custom commands**, and they'll have to remember where the Cheat Engine executable is for each game and set it (or check the config files with a text editor), and so on. You get the idea, this isn't ideal (and also not exclusive to Cheat Engine - Just a relevant example like HMM!)

The problem basically boils down to a lot of manual effort on the user's behalf when they want to set up a custom program for use with STL. In fairness, they could write a script that runs the application in the specific prefix, but that's not super user friendly :-) Having STL do some of the heavy lifting here would really help imo.

### Enter: Overhauled Custom Commands
This feature request would solve the concerns I and others had around #618, and would also help solve #624. The idea I had to resolve of this is basically as follows:

#### Custom Command Management Menu
This would be a new menu where a user can manage a list of custom commands/executables that they create and configure. Initially it would be blank, but there would be some buttons along the bottom: "Back" "Edit", "Delete", "Duplicate" and "Add". When a user clicks add, they'll get a dialog asking them to enter information about the custom executable, similar to the "Add Non-Steam Game" dialog. This would add a custom command with a set configuration that can be accessed globally for each game.

For example they could add a CheatEngine custom command with a bunch of parameters, then for each game they can go into the Game Menu and see a dropdown where they can access a custom command for every game. That way, per game, they only have to select "Cheat Engine". They can also choose where the Cheat Engine prefix is stored and configure other options described below.

[...]

Issue #806, opened 2023-05-19
Vortex broken - but easier


ajventer:
So the latest update to vortex broke it again - but it also makes it easier to fix and removes the need for work-arounds.

As of the latest update: two main things changed. 
1) It now requires windows 10, this breaks the installer but should be an easy fix
2) The DotNet requirement is now DotNet6. This is actually good news because DotNet6 doesn't require winetricks or work-arounds, it just installs cleanly under wine and is directly usable.
I've tested it, and you don't even need to preinstall it, Vortex will install it automatically on a clean prefix.

So this means - it should no longer be mandatory to run Vortex in it's own distinct prefix with symlinks etc. It could conceivably just be installed in game prefixes like we do with ModOrganizer. 
Even if you opt to stick to the current model, it should simplify the installer as you no longer need to use winetricks to install the very broken dotnet48 into it.

I tested it by simply downloading the installer and installing it using the "custom setup file" in Lutris, this created it's own prefix but for testing that's fine anyway, and hte only missing functionality is browser integration.

So, with a little luck, once STI fixes this one - it's fixed for a long time, and should not require nearly as much handholding. At this stage Vortex is actually very important because Wabbajack is still unusable on the steam deck (or linux in general) and Nexus collections is currently the only real alternative out there.

ajventer:
Update: did some more testing, it may not be quite as easy  as I thought, since out of the box it doesn't run in a proton prefix. I tried installing it into one using protontricks, the installer seems to run but the app crashes on load without any useful information.
I'm not sure what Lutris did differently so I can't reproduce it, but thought it worth mentioning.

sonic2kk:
Is this something that broke with Vortex 1.8.3? Vortex 1.8.0, 1.8.1 and 1.8.2 were fixed in STL git. The DotNet6 change was also already implemented in STL-git (#803).

> The DotNet requirement is now DotNet6. This is actually good news because DotNet6 doesn't require winetricks

Vortex can install DotNet6 itself, but STL-git will actually install `dotnetdesktop6` during installation so that Vortex doesn't have to. It's one less step for a user who may not fully trust the .NET 6 install process from Vortex (ex: me :-) ) 

> It now requires windows 10, this breaks the installer but should be an easy fix

The default version for Proton 8 at least should already be Windows 10. My Vortex prefix defaulted to Windows 10 with GE-Proton8-2.

[...]

PR #807, opened 2023-05-20
Vortex: Add option to use Steam Linux Runtime

sonic2kk:
Vortex runs with Proton, but doesn't run inside of the SLR. This may be causing compatibility issues on Steam Deck.

This PR adds a new option to the Vortex options in the Global Menu to enable the Steam Linux Runtime with Vortex. The changes required to implement this were non-trivial, but a combination of Proton and native games were tested before and after launching Vortex from the Global Menu (to ensure SLRCMD from the Vortex launch was not being re-used for the game launch). Everything in testing appears to still work fine.

The `setSLRReap` function was changed significantly to enable this functionality. It can now take an option to force Proton (instead of falling back to ISGAME=3 which causes native Linux SLR to be selected). It can also take a custom Proton version to use, so we don't use RUNPROTON by default. This means we can pass the Vortex version to getSLRReap and fetch the correct Steam Linux Runtime specified in its toolmanifest.

This implementation has the issue of very old compatibility tools which pass the SLR to use from the launch command will probably not work, as Vortex cannot detect this. However, recent Vortex versions require an up-to-date Wine anyway, so I doubt this is a problem.

The changes made to setSLRReap here should make it more straightforward to integrate with One-Time Run, MO2, and any other third-party application. As a result, these changes will actually end up being a prerequisite to some of the work in #788. It could also help with a future custom command overhaul where custom-commands could have a dedicated Proton version separate from the game (#625).

<hr>

Currently in my testing this has not caused any regressions with Vortex *launches* on my Linux Desktop (Arch Linux KDE Plasma Wayland), but mod compatibility etc remains to be seen.

<hr>

TODO:
- [x] Further testing
- [ ] Update langfiles


sonic2kk:
There was one instance during development where Vortex warned me that the installation was corrupt and that I should reinstall. The warning has not come back and could have come up because of some changes/incorrect logic in how the SLR was being/not being applied.

This should be investigated before merging, to see if it is a real issue and reproducible or if it is just a case of Wine being weird.

sonic2kk:
Merge conflict was just langfile from recent PRs, and STL version as usual.

sonic2kk:
According to https://github.com/sonic2kk/steamtinkerlaunch/issues/806#issuecomment-1565237885, Sniper runtime fixes Vortex (should be used by Proton 8 builds). Will do a little bit of further testing and probably merge this soon.

sonic2kk:
Forgot to mention it in the commit, but 70af528ce56417b2154025f66a57a1e9c5dd7dca **enables the SLR with Vortex by default**. If it can't find the given SLR it should fall back to running without it, like we do for game launches, and it appears to fix more issues than it doesn't, so I figure it's safe to enable by default.

sonic2kk:
Shellcheck looks good, version bumped, ready to merge.

Issue #809, opened 2023-05-23
[Feature request] Vortex downgrade


short-circuit:
## System Information
- SteamTinkerLaunch version: 12.12
- Distribution: Archlinux
## Feature Description
Give the possibility to install an older version of Vortex as newer install tend to break. This can be manually downloaded and then installed through the STL terminal command specifying the path to the install file e.g.:

```sh
steamtinkerlaunch vortex install PATH_TO_INSTALLER
```

Or if possible just by specifying version number, STL should install the requested version such as:
```sh
steamtinkerlaunch vortex install 1.7.8
```
Please correct me if this is already possible. Currently it is locking me out of using Vortex since the latest install fails with any provide proton version. Older Vortex installed without problem.

sonic2kk:
Thanks for the feature request!

> Please correct me if this is already possible. Currently it is locking me out of using Vortex since the latest install fails with any provide proton version. Older Vortex installed without problem.

Fresh installs of Vortex 1.8.0 and above work for me on Arch Linux with GE-Proton8-2 using STL-git. This has also been confirmed to work on Linux Desktop by another user, the outlier appears to be SteamOS (though #807 may resolve it, pending testing). If you're still having issues you could try STL from latest master, as it changes a little bit of the install process (mainly removes `dotnet48` installation). The latest version I have tested is Vortex 1.8.3, and I confirmed that a fresh installed worked just now with GE-Proton8-2. But this is besides the point of the feature request, which tl;dr I agree with but have a couple of reservations.

<hr>

I did consider the possibility of giving users the option to install older Vortex versions (mentioned in https://github.com/sonic2kk/steamtinkerlaunch/pull/802#issuecomment-1547003246 :wink:), particularly given that STL now has support for disabling Vortex auto-updates. The issue is that when specifying an older Vortex version, automatic updates would *have* to be disabled, otherwise Vortex has a chance of automatically updating. I'm not sure if this is a good idea or not, given that older Vortex versions will eventually have a cut-off according to Vortex (it shows a warning when you disable automatic updates), and a user might install an older version and have it also not work. However, a user can simply have enough sense to know that they should disable automatic updates before specifying a custom version.

[...]

PR #810, opened 2023-05-23
Vortex: Add option to specify version

sonic2kk:
Partially implements #809.

## Overview
This PR adds an option to specify a Vortex version to download, either from the commandline with something like `steamtinkerlaunch vortex start v1.7.0`, or from the GUI by enabling the option in the Global Menu and selecting a valid version. On the UI, the default version is "none" (`$NON`) which we account for and flag as an invalid value, so we fall back to the latest (either stable or beta, depending on what the user has selected).

If the user passes an invalid value, we fall back to getting the latest version, which as above can be either the latest stable or the latest beta depending on what the user has selected.

## Implementation
Example usage:

```
# Will just start Vortex normally with one of the following:
# latest stable, latest pre-release (if enabled), custom Vortex version (if set from Global Menu)
steamtinkerlaunch vortex start

# Will use the version passed here **regardless of menu preference**
steamtinkerlaunch vortex start v1.7.0
```

The logic in `getLatestGitHubExeVer` is just fetching the latest GitHub tag, and then feeding that in to build a URL for the download URL for that tag. Since this URL could be any (valid) tag, this part was split out into a separate function, `getGitHubExeVer`, that builds a URL based on a given tag. `getLatestGitHubExeVer` then passes its tag to this function.

When we pass a tag to Vortex to get, we pass our tag to `getGitHubExeVer` directly via a new function, `getVortVer`, which gets called if `USEVORTEXCUSTOMVER` is enabled (the checkbox on the UI) and `VORTEXCUSTOMVER` has a sane value (is not `$NON` and is set). If `getVortVer` cannot find a valid setup exe, we fall back to `getLatestGitHubExeVer`. The tag value can be set from the Vortex options on the Global Menu.

From the commandline, we check if we're passing any `$2` argument to Vortex that could potentially start Vortex, such as `download`, `install`, etc. If we are, we explicitly enable `USEVORTEXCUSTOMVER` (so the above check mentioned will pass) and then set `VORTEXCUSTOMVER` to the value passed in. The rest of the logic from here is handled by `getVortVer`, which will take the value of `VORTEXCUSTOMVER` as if it were passed from the UI.

## Related Work
#809 mentions passing a custom executable from the commandline. This would be a good idea imo as well, but will go in a separate PR.

Also, an option from the commandline to set no automatic Vortex updates would be useful imo as related work to this, but would also go in a separate PR and would only apply to the start command with something like `steamtinkerlaunch vortex start --no-auto-updates`. This flag would be permanent and could apply across updates. Applying this would set the preference in Vortex as well as in STL, which is something to be considered (we need to store it in the global config so the checkbox on the UI reflects the setting).

## Caveats
There are a few implementation caveats which were designed this way explicitly:
- Vortex versions passed from the commandline **will always override any preference for a given version**, as well as **overriding the preference to do this at all**. In other words, if a version tag is passed from the commandline, **we will always attempt to use this**
- This will **not reinstall Vortex**. To change Vortex version, the user is expected to re-install. I don't think there's any way for us to do this without implementing some kind of prefix removal, and users could lose data this way, so it is up to the user to understand the risks of meddling with Wine prefixes.
- If users are installing a Vortex version **older than the latest stable**, they should enable the option on the Global Menu to disable automatic Vortex updates. This will **not** be disabled automatically.

<hr>

TODO:
- [x] Some further testing (ideally from OP of linked issue)
- [x] Update help screen
- [x] Update wiki
- [x] Update langfiles


sonic2kk:
OP in #809 confirmed that this worked, noting that Winetricks needed for be updated beforehand with `sudo winetricks --self-update`. This will be noted on the wiki also.

sonic2kk:
Updated help screen and langfiles. Will do one more test pass, then update the wiki and merge.

sonic2kk:
[Wiki has been updated](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Vortex).

PR #811, opened 2023-05-25
Commandline: Add Winetricks option for MO2 and Vortex

sonic2kk:
Adds an option to run Winetricks in the MO2 and Vortex prefix. This can be useful if workarounds are needed for mods or newer versions of these tools to work (such as using an older version of Vortex in a custom prefix).

It also adds an option to output the prefix path for MO2 and Vortex, as convenience.

Some logging was also improved and fixed up around MO2 and Vortex's commandline Winecfg commands.

TODO
- [x] Update help screen


sonic2kk:
No langfile updates for this one! :partying_face:

sonic2kk:
Version bumped, ready to merge.

PR #812, opened 2023-05-25
One-Time Run: Show notifier with exename

sonic2kk:
Adds notifier to One-Time Run launches, from the GUI and commandline. It only shows two variations: For Proton launches and native launches. It doesn't show things like "using executable dir" etc because this would add some complexity to add - we'd need to create a bunch of translation strings for various use-cases and probably select which one to use, not to mention how long the strings might be. I don't see much benefit in anything outside of showing the exename and Proton version name where appropriate.

TODO:
- [x] Update langfiles


sonic2kk:
Langfiles updated, version bumped, shellcheck is green. Ready to merge.

Issue #814, opened 2023-05-27
Launching app that is allowed to outlive the game


Amzd:
## System Information
- SteamTinkerLaunch version: v12.12 
- Distribution: Nobara 37
- Installation Method: `sudo dnf install steamtinkerlaunch`

## Issue Description

I wish to make sure that OBS is launched when I launch a game, and I do not want to close OBS when I close the game. Currently I have a start script that launches OBS if it does not exist yet:

```sh
#!/usr/bin/env bash

# Launch OBS if it is not already launched
# checking for obs with a launch argument so this only works if you have a launch argument in the .desktop file
if [ "$(ps aux | grep "obs --" | awk '{print $11}')" == "grep" ]; then
  (gtk-launch com.obsproject.Studio.desktop) &  
fi
```
but when I use this as launch script in steamtinkerlaunch and quit the game, steam thinks that OBS is a child of the game process so when I press STOP it quits OBS.

When I run this same script in the steam launch option `"obs.sh" & %command%` it works the way I want (steam does not care if it is still running after the game closed and I can run a different game) but I would like to do this for all my games and steamtinkerlaunch seemed like the way to go.

I have tried all the `setsid`, `disown` combinations I could imagine but could not get it to work in start script.

Any help is very much appreciated.

sonic2kk:
I spent a while investigating this and wrote up a lot of replies which I eventually removed and re-wrote, because I actually figured  out the problem (after an unhealthy amount of caffeine ingestion): When Steam launches a process, any process (including STL), it launches a "reaper" program with it. As far as I am aware, this is launched *after* `%command%`, and this is important to note.

Reaper is some kind of Steam process that is designed to manage application processes launched from Steam, and then properly "kill" them (hence "reaper" :wink:). On game launch, Steam automatically starts the Reaper process. This was not always the case, from looking at the code I can see that the assumed behaviour before was that Reaper was started with the Steam Linux Runtime. I have confirmed that this is no longer the case as Reaper launches with any Proton game.

If you manually kill the Reaper process, and then exit a game from the game process itself, such as from an in-game quit button, then Steam will keep a forked custom command set in SteamTinkerLaunch open, and it will revert back to showing the "Play" button. However, with Reaper killed, if you press the "Quit" button from Steam, it will hang at "Stopping". This is true with and without SteamTinkerLaunch.

<hr>

[...]

PR #815, opened 2023-05-27
Reaper: Respect reaper toggle and kill reaper process

sonic2kk:
Fixes an issue reported as part of #814.

## Background
When using a forked custom command, this will successfully fork into the background and outlive the game and STL process. If STL or the game is closed, the custom command will stay open. However, Steam will still see the forked command as a child process of the game. If you press the "Close" button from Steam while a game is running, it will close the forked command with it.

This is because Steam uses `reaper` when a game is launched to track processes related to a game and clean them up on game close. This was likely done to better clean up game processes on Steam Deck in particular. In the past, this reaper command was launched separately and could be toggled on/off with SteamTinkerLaunch. Now the behaviour is entirely different and reaper is launched on each game launch.

An important distinction to make here is that reaper is only launched *after* (or with) `%command%`. Any processes that go *before* this are not subject to reaper, as shown in #814. So any processes started and forked into the background in the Steam launch options and before `%command%` will be unaffected by reaper.

## Overview
This PR changes the functionality of the reaper toggle in STL. Instead of checking if reaper is disabled and attempting to fetch and run reaper with the SLR, we remove that unnecessary logic now and instead we kill reaper if `USEREAPER` is `0` *and* if the process is running. This results in reaper being killed before game launch (though it *will* launch with STL, but this doesn't matter anyway).

We should probably also remove the "Force reaper" option as it means nothing now. Probably once #807 is merged we should investigate cleaning up the `useSLRReap` function as the logic around checking for the SLR from the Steam launch command will not apply anymore unless someone is using a heavily outdated version of Steam.

<hr>

TODO:
- [x] Update wording in langfiles to reflect new behaviour (reaper will not be used if available now, but rather "if disabled, reaper will be force killed, which means some forked processes will be kept alive after stopping a game from Steam)"
- [x] Update wiki


sonic2kk:
Wiki has been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Steam-Reaper

sonic2kk:
Pushed a couple of commits:
- Removed `FORCEREAP` as this does nothing. we can't store Reaper anymore since we don't get it from anywhere.
- Removed `HAVEREAP`, since this was used to check if we got Reaper from the Steam launch command which doesn't happen anymore, so this did absolutely nothing.
- Removed `FORCEREAP` checkbox because it did nothing now

sonic2kk:
While removing some further unused Reaper code and doing some reworks, some adjacent SLR code was removed from when the SLR was also taken from the incoming Steam args. Removing all of this code is going to be messy and can be done in the future.

The removed code appears to not break anything, I was able to launch some games no issues with the SLR enabled/disabled, but further testing should be done here.

<hr>

All langfiles have been updated now. This PR requires further testing, some conflict resolution, and then a version bump before merging.

sonic2kk:
Merge conflicts resolved. This resolution especially mandates further testing.

sonic2kk:
After testing it seems like this works. Tested a native Linux game (To the Moon) which requires the Steam Linux Runtime.

One thing I did notice is that when using native Linux games, because the `steamtinkerlaunch` part goes *before* `%command%`, `reaper` is not killed because it is not available until after game launch! This is an interesting problem.

sonic2kk:
Going to try this again, keeping the existing Reaper code and only adding the original commit.

PR #816, opened 2023-05-28
Vortex: Use Steam Linux Runtime for Installation

### LINKED ISSUE #806: Vortex broken - but easier

sonic2kk:
Should fix #806, for real this time.

If the `VORTEXUSESLR` is enabled, we attempt to install Vortex using the Steam Linux Runtime. As usual, if the SLR that a tool requires is not found, we will just run without the SLR.

This change is needed because SteamOS apparently needs the SLR to install Vortex properly at all.

Special thanks to @pikdum for spending a lot of time investigating and troubleshooting these issues.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230529-2 (vortex-slr-install)"
+PROGVERS="v14.0.20230528-3"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230528-2"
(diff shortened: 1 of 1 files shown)


sonic2kk:
https://github.com/sonic2kk/steamtinkerlaunch/pull/816/commits/59f5f60a1d22ae4e121463e90797f10b6226b58c was tested in #806, however subsequent commits were not. They only bumped the version and wrapped the SLR code in the `VORTEXUSESLR` check, so they should not break anything.

Will merge.

PR #820, opened 2023-06-03
Reaper: Respect reaper toggle for Proton games

### LINKED ISSUE #814: Launching app that is allowed to outlive the game

sonic2kk:
Fixes #814. Second attempt at #815, which was closed because the only changes needed were to the langfiles and to kill reaper if it is started before the game. Native Linux games work as expected with Reaper, since we can correctly parse `reaper` out of `%command%` (since native games use `steamtinkerlaunch %command%`, Reaper is started *after* the game as part of `%command%`).

Description below is copied from #815 and modified slightly to include updated information about native Linux games.

## Background
When using a forked custom command, this will successfully fork into the background and outlive the game and STL process. If STL or the game is closed, the custom command will stay open. However, Steam will still see the forked command as a child process of the game. If you press the "Close" button from Steam while a game is running, it will close the forked command with it.

This is because Steam uses `reaper` when a game is launched to track processes related to a game and clean them up on game close. This was likely done to better clean up game processes on Steam Deck in particular. In the past, this reaper command was launched separately and could be toggled on/off with SteamTinkerLaunch. Now the behaviour is entirely different and reaper is launched on each game launch.

An important distinction to make here is that reaper is only launched *after* (or with) `%command%`. Any processes that go *before* this are not subject to reaper, as shown in #814. So any processes started and forked into the background in the Steam launch options and before `%command%` will be unaffected by reaper.

## Overview
As a result, the reaper toggle currently works for native games. The start command with Reaper for native games is like  `~/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppID=<appid> <game_launch_stuff>`. So we don't need  to do anything extra here, we just add some logic to kill Reaper if it is started *before* a game, which only applies to Proton games (i.e. using compatibility tool).

This PR changes the functionality of the reaper toggle in STL with Proton games. Instead of checking if reaper is disabled and attempting to fetch and run reaper with the SLR, we remove that unnecessary logic now and instead we kill reaper if `USEREAPER` is `0` *and* if the process is running. This results in reaper being killed before game launch (though it *will* launch with STL, but this doesn't matter anyway).

Behaviour for native games is unaffected.


sonic2kk:
Shellcheck looks good. Version is bumped. Ready to merge.

Issue #822, opened 2023-06-07
Request: Help on reading the source code


garaiza-93:
Hello, I'm trying to make a patch for `steamtinkerlaunch` for use on NixOS, and I need some help trudging through the source code.

What I'm trying to figure out is where STL deletes a language file from `lang/` when a game is started in Steam. After looking through the source code for `rm` statements, I have a gut feeling that this file is not deleted on its own, but alongside other files. However, I'm not able to pin down an exact line number.

Any tips would be appreciated.
Here's the original issue for context and log messages: https://github.com/NixOS/nixpkgs/issues/226086

sonic2kk:
First off, I love the dedication here and that you linked to the original issue for context :-)

I'm more than happy to help, I'm currently a little busy and feeling a bit under-the-weather but later today I'll try to refresh my memory on how the language files are handled. I'm pretty sure STL removes them and re-creates them to update the langfiles to the latest each time (primarily for Steam Deck iirc where it's installed locally and not system-wide).

As for how they're removed, I think I created a helper method to do it, so that may explain why searching for `rm` on its own may not make it immediately obvious how they're being removed.

For how a game is started, I think it's `launchSteamGame`. There is a "master" launch function that makes slightly different calls for Non-Steam Games, MO2, and maybe others. I'll have to double-check on that.

And lastly, I'm not sure how things are handled with NixOS packages, but I'm *more* than happy to accept a patch/work on a patch to improve support on the STL side, to save having to potentially manage a patched STL specifically for NixOS. I don't use NixOS and I am not overly familiar with it but as long as the "GNU" flavour of the tools STL uses work then I'm happy to try improve things here to help users.

There is much room for improvement with STL and I'm looking at improving how various things are handled where I can, removing various outdated parts of the code, removing duplicated code where I can, but it's a slow process :-) Even though I've been maintainer for almost 8 months now I'm still getting to grips with some of the "intricacies" 😉 

<hr>

Once I have some more time I'll look at the code and give a more contextual and thorough answer. Once again I appreciate the initiative here and I'm happy to help!

garaiza-93:
Thanks for the kind comment! I'll look at the functions you mentioned more closely once I can get some free time 😄 

[...]

PR #824, opened 2023-06-07
Ensure write permissions before and after lang file update

garaiza-93:
Related to #822

Use `chmod` to make sure language files in the user's config directory have write permissions before AND after updating them.

I've decided on ensuring write permissions before and after the update just in case of an unforeseen edge case.

## Testing
1. Remove write permissions from all language files in the config's `lang/` directory with `chmod -R -w /path/to/config/lang`
2. In a terminal, execute `steam`
3. Pick a game. In the game's Properties, force the use of `steamtinkerlaunch`
4. Start the game
5. Check that in steam's stdout, there are no complaints of `rm` waiting for user input to proceed with removing existing language file(s)

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -2263,8 +2263,10 @@ function loadLangFile {
 
 				if [ -n "$UPDATELANGFILEPATH" ]; then
 					writelog "INFO" "${FUNCNAME[0]} - Found lang file to replace the existing user-installed file with under '$UPDATELANGFILEPATH'"
+          chmod +w "$LAFI" #Ensure write permissions before removing
 					rm "$LAFI"
 					cp "$UPDATELANGFILEPATH" "$LAFI"
+          chmod -R +w "$STLLANGDIR" #Ensure write permissions for next update!
 				else
 					writelog "INFO" "${FUNCNAME[0]} - No lang file to replace existing user-installed file with, not updating langfile"
 				fi


sonic2kk:
Wow! Thank you :smile: 

> I've decided on ensuring write permissions before and after the update just in case of an unforeseen edge case.

I had a stashed change that did exactly this (stashed because I was working on d57e57243e3ea37a52eb0fa6b5c7ceef82f01fad :wink:), so I think this is a sensible approach. I didn't have time to test it so I appreciate the work you've done here!

>Check that in steam's stdout, there are no complaints of rm waiting for user input to proceed with removing existing language file(s)

Ah, duh, I'm dumb. In #822 I wondered why this was causing issues, but it was even mentioned in the NixOS issue that `rm` waiting on user input causing the problem :sweat_smile:

<hr>

There are two very small things I would just add before merging:
1. I wonder if it's best to do this before `if [ -f "$LANGFILENAME" ]; then` (line 2236 in this PR), and then again at the very bottom of the if block that encapsulates the whole method (line 2269 in this PR). It probably doesn't matter since this issue should pretty much only affect Nix, so I will leave it to your discretion (if it turns out to be useful in other cases it can addressed at that point). I'm happy to merge it as-is :-)
2. The more pertinent question: How do we want to handle getting this into the Nix package? This will be merged into master, the """release strategy""" for STL is to basically merge to master and then, for lack of a better phrase, "once I feel like it" (and when I have a cool name in mind!) I make a tag and release from all the commits in master since the last release. That doesn't bode too well for Nix, where this fix is kind of... urgent. It might be possible to take this PR [as a patch](https://patch-diff.githubusercontent.com/raw/sonic2kk/steamtinkerlaunch/pull/824.patch) and apply it to the Nix package at install time. Of course, either way, this won't impact it getting merged, because it's a correct fix, just wondering what would come after :-) 

<hr>

Awesome work here! Thanks for taking the time to make this change and test it, this will be ready to merge after some feedback on the above questions :-)

sonic2kk:
Oh, I was just thinking, if you think the second point about how to get this into Nix would be better discussed in the https://github.com/NixOS/nixpkgs/issues/226086 I can chime in there :-)

garaiza-93:
> I wonder if it's best to do this before if [ -f "$LANGFILENAME" ]; then (line 2236 in this PR), and then again at the very bottom of the if block that encapsulates the whole method (line 2269 in this PR).

I think it's fine as-is for readability, instead of having to look at a larger scope.

> The more pertinent question: How do we want to handle getting this into the Nix package? 

Not a problem! It's just a matter of telling Nix to look for this commit hash. so we don't have to wait for the next release!  However, if the maintainer for STL's package wants to do things differently, that's up to them.

For cool edu-ma-cational purposes, here's how Nix gets STL's code:
```nix
  src = fetchFromGitHub {
    owner = "sonic2kk";
    repo = pname;
    rev = "v${version}";
    hash = "sha256-oigHNfg5rHxRabwUs66ye+chJzivmCIw8mg/GaJLPkg=";
  };
  ```
 Here, `pname` is just `steamtinkerlaunch`, just stored in a variable. `owner` and `repo` are basic repo information, as in who owns it and  the name of the repository. `rev` is the cool part though, because instead of a tag, it can just as easily be a commit hash and it'll fetch the code no problem! `hash` is used internally to identify the directory extracted from the source archive. If you're curious, all the packaging for STL is [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/games/steamtinkerlaunch/default.nix)!

sonic2kk:
> I think it's fine as-is for readability, instead of having to look at a larger scope.

Sounds good to me :-)

> Not a problem! It's just a matter of telling Nix to look for this commit hash. so we don't have to wait for the next release! However, if the maintainer for STL's package wants to do things differently, that's up to them.

That is fair enough! They are always free to reach out to me if they need any assistance as well (not sure if I can provide any :sweat_smile:)

> For cool edu-ma-cational purposes, here's how Nix gets STL's code:

That is pretty sweet and very readable actually (also, still get a kick out of seeing `sonic2kk` in other people's code heh). I wonder if pointing to a specific `rev`ision would cause all the changes from 12.12 up to now to be pulled in... But either way, like you said, that's probably up to the maintainer to figure out.

<hr>

This looks good to me, ready to merge I think :+1:

Thanks!

sonic2kk:
D'oh, forgot to ask you to bump the version, sorry about that (it's not a big deal, I forget fairly often to do it myself). Going to push a small commit to bump it.

Also, updated the [changelog](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog) to include mention of your fix, so when a release comes out and I copypaste these release notes you'll probably get pinged :-)

Issue #813, opened 2023-05-26
MO2 process stuck on exit


therob:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: Ubuntu 20.04 LTS
- Installation Method: ProtonUp-Qt

## Issue Description
I'm currently trying to mod Fallout 4 with MO2.
The MO2 process does not exit after closing the UI and needs to be killed via Steam. Also the details view of a mod in MO2 (double-clicking an installed mod) just opens a window with black content and then the process completely freezes. Disabling the Wine theme does not fix this.
This only happens in Game Mode. Standalone mode works fine. I'm using GE-Proton8-3 to run MO2 in both modes.

## Workaround
Using GE-Proton7-55 in Game Mode fixes this issue completely (hanging on exit and the details view). For me, this also requires disabling the Wine theme to fix the [Glitchy UI](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Modding#glitchy-ui) issue.
-- or --
Using GE-Proton8-3 and disabling Steam Linux Runtime (SLR) also solves this.

## Logs
There is nothing obvious in the logs. Comparing the logs between Standalone an Game Mode also shows nothing suspicious. I tried setting the logging in MO2 to the highest level, but this shows nothing when it happens.
I guess wine logging could help here, but I'm not sure what logging category to enable and full logging usually creates a lot of log spam. If you have any idea which logging categories could help, then I'm happy to try it again and provide the logs.

sonic2kk:
I guess STL could attempt to force kill MO2 on exit, but what is the issue for SteamTinkerLaunch here? Am I misunderstanding here or is this solely an issue when using a build of GE-Proton8-X (may affect other builds of Proton which use Wine 8) that is fixed by using GE-Proton7-55 (may also work with builds of Proton which use Wine 7)?

If this is the case then I don't think this is a bug on the SteamTinkerLaunch side.

> The MO2 process does not exit after closing the UI and needs to be killed via Steam. Also the details view of a mod in MO2 (double-clicking an installed mod) just opens a window with black content and then the process completely freezes

What you've described here seems like a Wine bug. One user reported issues with MO2 on builds of Proton which use Wine 8+ (#805). However the other issue:

> This only happens in Game Mode. Standalone mode works fine. I'm using GE-Proton 8.3 to run MO2 in both modes.

[...]

PR #826, opened 2023-06-07
One-Time Run: Allow Steam Linux Runtime

sonic2kk:
More work for #788.

## Overview
Using the Steam Linux Runtime can improve compatibility with executables, particularly on Steam Deck with builds of Proton 8 and above. Though this is not always the case.
- In #772, the SLR fixed a Non-Steam game not working on SteamOS.
- In #806, the SLR seems to fix Vortex on SteamOS.
- In #813, the SLR causes MO2 to crash.

In spite of this, Proton is *intended* to be used with the Steam Linux Runtime (though not fully intended to be used outside of Steam in general). Giving users the option to use the SLR is consistent with how Proton is supposed to be used, and allows users to toggle it based on their needs.

I have yet to see an instance personally where this is required for One-Time Run, but I have yet to run into a case where I required the Steam Linux Runtime on my PC. This option may be useful on Steam Deck for One-Time Run where applications may fail.

## Implementation
We fetch the SLR to use using Pressure Vessel Funtime in `setSLRReap`, called from `setNonGameSLRReap`. We tell `setNonGameSLRReap` whether to get the native or Proton SLR based on `ISWINDOWSEXE`.

The SLR can be enabled from the GUI by checking the "Use Steam Linux Runtime" option. On the command line, it can be toggled with the `--useslr` flag. STL can then infer what SLR it needs to get. Example usage:

```
$ steamtinkerlaunch otr 1290000 --exe="/path/to/app.exe" --useslr

$ steamtinkerlaunch otr 1290000 --exe="/path/to/script.sh" --useslr
```

## TODO
- [x] Test for regression with "Save" option
- [x] Update help screen to include `--useslr` option
- [x] Update notifier to note when we are/aren't using the Steam Linux Runtime (will require a new language string)
- [x] Update wiki
- [x] Update langfiles


sonic2kk:
Options same to all be saving correctly except for the Proton version, odd...

sonic2kk:
Fixed the saving issue.

sonic2kk:
Notifier added. Langfiles and wiki need updated now.

sonic2kk:
Wiki updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/One-Time-Run

sonic2kk:
Langfiles updated, this should be ready to merge

sonic2kk:
Did one final round of testing (using Committee of Zero STEINS;GATE patch installer) and everything works as expected.

Issue #828, opened 2023-06-08
Vortex deploy fails with Skyrim SE - vortex.log says "EXDEV: cross-device link not permitted"


Kaylebor:
## System Information
- SteamTinkerLaunch version: steamtinkerlaunch-v14.0.20230607-2
- Distribution: SteamOS, EndeavourOS (multiple systems)
- Installation Method: GitHub clone

## Issue Description
I had a working Vortex install, which broke after a recent update; after this was fixed on SteamTinkerLaunch, and tinkering with options, Vortex now works.

After further configuring Vortex (logging in on my Nexus account, setting Vortex profiles), and trying to deploy, Vortex shows non-descriptive error messages, and none of the hardlinks are created.

Looking around on my own, I found where Vortex stores logs (attached below): there are a lot of errors, at least one per hardlink, all of them with the message `EXDEV: cross-device link not permitted`.

I managed to fix this locally on my Steam Deck, by installing Vortex normally, shut it down, configure SteamTinkerLaunch to not use Steam Runtime with Vortex, and then running Vortex again.

Could there be some permissions issue when trying to hardlink with Steam Runtime?

Things I have tried, deleting the `vortex` folder completely between tries:
- Change drive in which Skyrim is installed (**no difference**)
  - Both on an EndeavourOS desktop computer, and Steam Deck; same error, no difference
- Change Proton version (**no difference**)
  - GE-Proton from 8.2 to 8.4
- Run without Steam Runtime (**works**)
  - Only tested on Steam Deck; need to run once with Steam Runtime to install .NET

## Logs
Vortex logs can be found in `~/.config/steamtinkerlaunch/vortex/compatdata/pfx/drive_c/users/steamuser/AppData/Roaming/Vortex/`.

Each is hundreds of lines, and bigger than the allowed Pastebin free limit; I will upload it if needed, but the only thing that jumped to me was this line (repeated once per mod file):
```
[...]

Issue #823, opened 2023-06-07
Vortex hangs on prepareAllInstalledVortexGames


jbbandos:
## System Information
- SteamTinkerLaunch version: v14.0.20230603-1
- Distribution: SteamOS 
- Installation Method: ProtonUp-QT

## Issue Description
Vortex isn't launching. It hangs after "prepareAllInstalledVortexGames - Applying Vortex settings for all games via autogenerated cmd '/dev/shm/steamtinkerlaunch/vortset.cmd"
I've tried deleting the prefix dir, installing again, and doing a reset with "steamtinkerlaunch vortex reset", but it still hangs. It has worked a couple of times, without changing anything, which is even more frustrating, and now it just doesn't work at all.

## Logs
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/11676828/steamtinkerlaunch.log)

[installVortex_protonrun.log](https://github.com/sonic2kk/steamtinkerlaunch/files/11676228/installVortex_protonrun.log)

sonic2kk:
I will test when I can but it was working about a week ago on PC and Steam Deck.

How long have you left it for? This process can take a long time, reportedly upwards of 10 minutes (for me it takes ~5 max).

jbbandos:
I've left it for over one hour, it never passes that point.

sonic2kk:
The log you have attached is for `steamtinkerlaunch version`, not for your attempted Vortex launch. Try attaching the log after an execution of i.e. `steamtinkerlaunch vortex start`.

jbbandos:
My apologies, I've uploaded the correct log now. 
Just tried launching vortex again, now exits, and the log doesn't have much.
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/11676842/steamtinkerlaunch.log)

sonic2kk:
Just tested and it installed and ran fine for me on my Arch PC, but stupidly I forgot I had the SLR disabled for Vortex... I'll try again with it enabled but I expect the results to be the same. Vortex installed for me with GE-Proton8-4 and I installed Vortex 1.8.4.

An interesting log to check might be `/dev/shm/steamtinkerlaunch/vwrun.txt`

[...]

PR #829, opened 2023-06-08
Vortex: Only use the Steam Linux Runtime to install Vortex

sonic2kk:
May solve #828, partially related to #823.

Adds a separate option to use Vortex with the SLR. Now the default is to only use the SLR for a Vortex install and if dotnet6 is not found, as the SLR Is required to install it.

If this doesn't work, I'm done with Vortex. Nexusmods, make a native Linux port. And to users, just manually mod if Vortex doesn't work.


sonic2kk:
Works according to both #823 and #828. Merging.

PR #833, opened 2023-06-11
One-Time Run: Add option to reset entries back to default

sonic2kk:
Further work for #788.

Clears all One-Time Run related variables in the per-game config to be back to their default values. Blank values essentially reflect the defaults, as blank is false for checkboxes, none for paths, and the Proton version will be handled appropriately too.

Currently this is only implemented from the commandline, example usage:

```
$ steamtinkerlaunch otr 223280 --default
```

Also adds 'STEAM_COMPAT_CLIENT_INSTALL_PATH=', which the Proton script now seems to require to launch with Proton.

<hr>

TODO:
- [x] Add button to One-Time Run GUI
- [x] Update help screen with new option
- [x] Update langfiles (if needed, for the "Default" button)
- [ ] Update wiki

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230612-2 (otr-default)"
+PROGVERS="v14.0.20230612-1"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Defaults button was added to the One-Time Run GUI.

image

The language string was re-used from the Main Menu. This is probably fine!

sonic2kk:
Help screen updated, shellcheck is green. Wiki will be updated and this will be merged.

sonic2kk:
Wiki has been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/One-Time-Run/77d57dcd272729a1d29ee4e0800be4e9837a2adb

PR #834, opened 2023-06-11
One-Time Run: Use notifier for invalid executable, Proton, and run directory

sonic2kk:
More work for #788.

When something goes wrong during a One-Time Run, there is some logging but it is not immediately obvious what went wrong visually. Users are expected to check the log to trace through execution and figure out what the problem is, but this PR adds a little bit of visual niceness using the notifier (if enabled, of course) and also uses `echo` to detail when something went wrong.

This is purely a quality-of-life improvement, with minimal functional change for One-Time Run. The only functional change added was to check to ensure an executable exists before trying to launch it, by adding a `&& [ -f "$OTEXE" ]`.

<hr>

TODO:
- [x] Update langfiles


sonic2kk:
Langfiles updated, version bumped, shellcheck is good, no regressions that I can see.

PR #840, opened 2023-06-18
Commandline: Allow set command to take 'global' instead of AppID

sonic2kk:
This allows for setting a value in the Global config.

TODO:
- [x] Test for any breakages
- [x] Test command output
- [x] Update helpscreen


sonic2kk:
Low on time currently so this may have to wait a bit. Anyone interested is free to test as well though. I tested a little bit after I made this PR and it worked ok

sonic2kk:
Command output works fine, it outputs the correct variables when incorrect usage is performed.

The help screen should also be updated for this.

sonic2kk:
As for breakages, this appears to be strictly a user-facing option. I also used this branch for a while without realising it and it worked fine, so this is probably good to merge.

Issue #838, opened 2023-06-16
[Enhancement] Upcoming GameScope Filtering Changes


sonic2kk:
## System Information
- SteamTinkerLaunch version: git
- Distribution: Arch Linux

## Feature Description
GameScope in git has changed how it takes upscaling and filtering arguments. The existing `-Y`, `U`, and `-n` flags for NIS, FSR and Integer Scaling respectively have been **removed** and **replaced** with the following ([as noted on the Readme](https://github.com/ValveSoftware/gamescope/blob/7a1fe2def38e9c3ed4d12adb9f1bb81b6f9dcbe5/README.md?plain=1#L66-L69) as of the latest commit at time of writing: ValveSoftware/gamescope@7a1fe2def38e9c3ed4d12adb9f1bb81b6f9dcbe5):
- `-F <filter>`, where `<filter>`  can be `linear`, `nearest`, `nis`, `fsr`
- `-S <scale>`, where `<scale>` can be `auto`, `integer`, `fit`, `fill`, `stretch`.

Since this is in git, some STL users may be affected. In future when this makes it into a stable release, this will affect all STL users making use of GameScope. That is, once those users update and a newer version is available to them (package managers, etc).

<hr>

STL will need to update in the meanwhile to accommodate **both** the old *and new* usage, STL will need to make some changes:
- Check the GameScope manpage and make some assumptions about the GameScope version used (if it has the new vs. old docs we can make assumptions about what syntax to use)
- Update to accomodate the new syntax
- Add support for the (new?) `stretch` option.

I am not sure of the depth required for these changes. I am hoping that the overhaul to GameScope (#744) will make this a bit smoother.

These changes will stay in a branch for a while to give these changes a chance to "settle" upstream, in case there are any changes in the meanwhile to account for.

<hr>

I am opening this issue to remind myself that this change will be required ASAP. To test, I will test against builds of GameScope before and after the commit for this refactor.

TODO:
- [x] Refactor UI to use dropdowns for scaling and filtering options instead of checkboxes
- [x] Refactor code to allow `-F` for setting FSR, NIS, and Nearest Neighbor *if* the current GameScope help screen uses those options
- [x] Refactor code to allow `-S` for setting Integer scaling
- [x] Investigate support for new Filter option: `linear` (probably default so might not be needed)
- [x] Investigate support for new scale options: `fit`, `fill`, and `stretch`
- [x] Investigate support for `auto` scale (probably default so might not be needed)

[...]

PR #844, opened 2023-06-25
GameScope: Add default button

sonic2kk:
A somewhat prerequisite for #838, because it will make the transition easier for users. 

Adds a Default button to the GameScope button bar, so users can reset all options. This will be useful for #838 because it will introduce a breaking change **for all GameScope GUI elements**, since the order of elements will change entirely with several being merged into one.

This also fixes a bug that should never occur but that may occur in certain circumstances. When a dropdown value is blank, that option would be written out to the options because we only check if the value is not the default and not if it is not defined at all. So if `GSFOOR` was `""`, that is `!=` to the default, so we wrote out `--force-orentation ` with an empty value. This should never really occur but it happened during testing, so I fixed it when a `-n` check for GameScope dropdown variables.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230626-1"
+PROGVERS="v14.0.20230626-2 (gamescope-default-button)"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Shellcheck is good on this, this is probably good to merge.

sonic2kk:
This option does not impact the gamescope checkbox itself, as this is stored in a different variable. Imo it makes more sense to keep this as whatever it was before the defaulting.

PR #853, opened 2023-07-08
GameScope: Add Keyboard Grab option

sonic2kk:
Adds one of the new GameScope options, `-g`, which allows for locking keyboard input to the GameScope window. Not sure how useful this is but probably for users in embedded mode it'll be more useful.

Tested this out and existing options didn't seem to break, but a slightly more thorough test and an update to the langfiles should be all that's left for this one.


sonic2kk:
Langfiles updated, version bumped, ShellCheck looks good :+1: Ready to merge

PR #854, opened 2023-07-09
Steam Deck: Add command to remove installed dependencies

sonic2kk:
A simple stop-gap resolution for #719.

This adds a new command `cleardeckdeps`, which will remove the Steam Deck dependencies directory (usually `/home/deck/stl/deps`). When this is removed, STL will redownload up-to-date dependencies on next launch on Steam Deck.

Currently if we bump the required dependency version in the script, the dependencies will not be updated because we don't track the version, so STL will see that the files are there and assume no download is needed. One solution is to always re-download the dependencies but this is messy. Ideally, eventually, we'll track the installed versions and then check if we need to update, but right now this PR is a simple solution.

When the dependencies folder is removed, and a version of STL with an updated dependency version is used, the updated files will be downloaded because the download logic will be called when the files are missing. A user can manually remove the deps folder, but a command is slightly more accessible I think.

This PR is entirely untested for now. I don't feel like testing on my Steam Deck yet because I'm lazy :stuck_out_tongue_winking_eye: But this should help #719. Afterwards I will probably consolidate the improvement mentioned in this PR and that issue into one bigger initiative issue describing various Steam Deck improvements. Someone from the community may be able to pick up the mantle for that, as I very, very rarely use STL on my Steam Deck.

TODO:
- [x] Testing on hardware


sonic2kk:
This change should also be documented on the wiki, and in the changelog.

sonic2kk:
This works on hardware, but an additional change to update dependencies which will just run this and then update dependencies in the same cycle may work too. Eh, it could go in another PR. This is a stop-gap anyway.

sonic2kk:
Side note: Commandline usage on Steam Deck is a bit of a pain with the automatic update stuff. That needs a big overhaul. I'm hoping to make an issue describing Steam Deck improvements I would like to see as part of STL, mainly opt-in auto updates and changes to constantly checking for new dependencies. But I'll make an issue about that *eventually*.

sonic2kk:
Tested on hardware as mentioned and this works. ShellCheck is good too, and the version has been bumped. This is good to merge.

PR #855, opened 2023-07-09
GameScope: Update --hdr-sdr-content-nits default value

sonic2kk:
The default value for `--hdr-sdr-content-nits` is now 400 instead of 203. This updates the check for the default value and a couple of comments around it.

This PR also documents the `--sdr-gamut-wideness` option. I initially considered adding that as part of the PR, but I was not sure what the default value is. I briefly documented it in the code, so if in future if we want to add this/if it's requested, it can be added.


sonic2kk:
ShellCheck is good, this one is straightforward enough to add.

PR #856, opened 2023-07-10
Update chinese.txt

Juijote:
I have translated the supplementary text, and when I checked the previous translated text, I found that it was rather rough, so maybe I will translate it all again in a uniform way when I have time.


sonic2kk (approved):

sonic2kk:
You are legendary for doing this :rocket: Thank you so much!

> so maybe I will translate it all again in a uniform way when I have time.

If you do, that would be great! Even if you want to do it in sections that's fine too. Translations are all community-submitted and maintained so the quality may vary.

Thanks again! This looks good to merge to me.

PR #861, opened 2023-07-16
commandline: Command to install Steam Linux Runtime required by game Proton version

sonic2kk:
Different Proton versions require different Steam Linux Runtime versions. For example, Proton 5.13 through Proton 7 require the "Steam Linux Runtime - Soldier", whereas Proton 8.0 requires "Steam Linux Runtime - Sniper". Native Linux games will use "Steam Linux Runtime", which is for native games (internally this is called "Scout"). These are available from Steam under the "Tools" section of the library, but it is not immediately clear which Steam Linux Runtime is required by a given game's Proton version.

## Overview
This PR adds a command which will check the required Steam Linux Runtime for a game, by checking its Proton version's `toolmanifest.vdf` and telling Steam to install the `require_tool_appid`. We do this by running the command with the required SLR AppID like so: `steam steam://install/<appid>`. **This will only work for games previously launched with SteamTinkerLaunch**, as we rely on getting the Proton version from the game's config. However the benefit of this is that users don't need to know which SLR is required, they can just tell STL to prompt Steam to install the SLR. We can't do this ourselves before a game launch because there is no way to wait until the SLR is installed before launching a game.

When no AppID is passed to the command, we default to the last game used with SteamTinkerLaunch. For Native Linux games, we can't really tell when we have a native or Proton game (native games still have a `USEPROTON` specified), so instead we have a separate argument "`native`" to tell STL that we want to get the SLR for a native game here.

In the case of no SLR, no Proton version, no game config, no tool manifest, or no `require_tool_appid`, we simply log and do nothing. If the required SLR is already installed for a given game, we tell the user and do nothing. 

Example usage:
```
$ steamtinkerlaunch getslr  # Will get the SLR for the most recent game launched with STL (will always assume Proton)
$ steamtinkerlaunch getslr 638970  # Will get the Proton version for this game and get the required tool AppID from its toolmanifest.vdf
$ steamtinkerlaunch getslr 1999740 native  # Will get the native Steam Linux Runtime
```

## Future Improvements
This command could be extended to work on the GUI in the future. I am considering replacing the "WikI" button with a button to install the Steam Linux Runtime for the current game. This way it can be installed by the user and they'll be responsible for ensuring it is installed before a game's launch.

<hr>

TODO:
- [x] Update help screen


sonic2kk:
Help screen updated, ShellCheck looks good, this should be ready to merge.

Issue #866, opened 2023-07-22
Compatdata redirect to mayor version folder


cyberconan:
## System Information
- SteamTinkerLaunch version: 12.12
- Distribution: Arch Linux

## Feature Description
I'm using compatdata redirect to global-proton directory, is an incredible and amazing option because I have a big compatdata folder with all requirements for diferent games (ubisoft connect, rockstar, directx, vcruntimes, ...). The only problem than I see happens when a Proton version have a minor upgraded because a new compatibility folder is created.

For solve this, I keep manually a structure folder like this:

![Screenshot_20230722_103757](https://github.com/sonic2kk/steamtinkerlaunch/assets/1497109/841c54c5-8af4-48b4-ad54-032926f78e95)

As you can see, for example, I'll keep a "compatdata-proton-8.0" and -2d and 3c versions are linked to "real folder". All my games using proton 8 version are always connected to this folder regardless they are using -2d or 3c version. Same with proton GE-8 or experimental versions.

Is possible to implement an option than can active compatdata folder names have mayor version names? Thank you so much for your work and dedication.

sonic2kk:
This is done for prefix compatibility, as updating prefixes to different Proton versions is risky and can result in breakages. And going between different Proton major versions, such as Proton 7.0 to Proton 8.0, can be especially risky and *will* require the prefix to be updated to account for new DLLs, etc.

If I am understanding correctly,you can an option to only have new compatdata folders redirect for major Proton version changes, such as Proton 6/7/8, and probably also counting Experimental as a separate version. Then your folder structure would only have, for example, `compatdata-proton-8`.

I think this would be an ok idea to implement. It is definitely ***NOT*** desirable, but it is possible and we could add a checkbox. As far as implementation goes, I think we should disregard minor Proton version bumps when the option is enabled, such as Proton 8.0-3c. The same should go for Experimental bumps as well, just have one `compatdata-proton-experimental`. But we *should* still bump for new GE-Proton/tkg/etc versions.

One way to go about implementing this would be to derive the redirect folder name from the Proton folder name. Having this behind a checkbox means it will not be the default behaviour, so we don't need to worry about this being a "breaking" change. So in cases where a user already has `compatdata-proton-8.0-3c`, when they enable the checkbox they will then get `compatdata-proton-8.0`.

Deriving the folder name should be as simple as replacing spaces with `-` with a simple `sed`.

<hr>

[...]

PR #867, opened 2023-07-22
Compatdata Redirect: Add option to only redirect on major Proton versions

sonic2kk:
Implements #866.

Adds a checkbox on the Game Menu that will only create a new compatdata redirect folder on major Proton version bumps.

The default behaviour right now is to make a new compatdata redirect on each new Proton version, including minor fix versions. So a user could end up with `compatdata-proton-8.0-2d`, `compatdata-proton-8.0-3c`, `compatdata-GE-Proton8-6` and so on. This new option will only create new compatdata directories for major version bumps, such as `compatdata-proton-8.0` or `compatdata-GE-Proton8`.

SteamTinkerLaunch should be able to automatically pick up and switch to using `compatdata-proton-XX` over the folder with the minor fix version. In testing going between both of these options worked without issue, the logs also indicated that the compatdata for the game I was testing with was being selected correctly.

TODO:
- [x] Testing from OP who reported the issue
- [x] Update langfiles


sonic2kk:
Initial testing from OP is looking good, this is probably ready to merge once langfiles are updated, unless anything last-minute comes up.

sonic2kk:
Updated the description string slightly. ShellCheck is good for this and the version is bumped, so this is good to merge.

PR #869, opened 2023-07-28
Global Menu: Add option to force XWayland for Yad

sonic2kk:
Adds an option to the Global Menu which allows for forcing XWayland when using Yad, by exporting `GDK_BACKEND=x11`. This can be useful for compatibility in some instances, if XWayland is buggy, or if some Desktop Environment settings cause issues with Yad's Wayland support. One example is KDE Plasma's "Apply Scaling Themselves" option, which really breaks Yad's sizing functionality on multi-monitor displays with various different scale factors.

The `export GDK_BACKEND` may cause issues in future with some programs that rely on this, for example launching a native Linux custom program which actually has good Wayland support will force it to launch with the X11 backend. If this becomes an option we'll need to investigate some alternative, such as finding a way to only set this for Yad.

TODO:
- [x] Update langfiles


sonic2kk:
In future it may be possible to fix any issues with the GDK variable export by not exporting it and just setting it. This raises a ShellCheck warning, [`SC2034`](https://www.shellcheck.net/wiki/SC2034), which if needs be we can always suppress if not exporting fixes it. I tested and not exporting still keeps the functionality of this PR, so it is a potential future solution!

For now, it works fine.

sonic2kk:
Version bumped, langfiles updated, ShellCheck is green. Ready for merge!

PR #870, opened 2023-07-29
Main Menu: Add button to install Steam Linux Runtime

sonic2kk:
Implements #871.

## Overview
This PR replaces the wiki button on the Main Menu with a button to tell Steam to fetch the Steam Linux Runtime required by a given game. It uses the current game AppID and passes that to `commandlineFetchGameSLR`, which manages calling out to get the SLR required by the given Proton version. We also attempt to check for native games with an `ISGAME -eq 3` check (though this will always be true if not running from Steam).

This is essentially an attempt to emulate the `getslr` command but from the Main Menu (#861). In quick testing it works and fetches the right Steam Linux Runtime, but we will see. There is scope in future to extend this to allow selecting the SLR in general, from a checkbox menu kind of like the Steam Eval Script menu. We could have a list of SLRs and by default highlight the one we think the game needs. But for now this initial implementation works.

## Rationale
There are a couple of reasons for opting to replace the Wiki button on the Main Menu:
1. The wiki button only takes you to the wiki homepage, and various wiki pages can be accessed from the F1 menu (though this is less obvious than a wiki button).
3. There is an option to open the wiki in the browser from the app icon menu.
4. There is a command to open a wiki page in the browser.
5. The wiki button relies on Yad WebKit support, which is not present on SteamOS or other builds using the Yad AppImage. When the Yad WebKit support is not available, the wiki command defaults to using the browser, and no one has complained about this.
6. Using the wiki button doesn't really work from things like GameScope sessions or other scenarios where multiple windows aren't well supported (though the same could be said for prompting to install the SLR from Steam itself).
7. For most users, installing the SLR for the selected Proton version is more useful as it can help fix games that won't launch. 

## Remaining Work
There are some pieces of remaining work. Namely, we should update the Steam Linux Runtime wiki page before merging this, so if a user reads it and thinks "What is the Steam Linux Runtime?", we can have a wiki page for them to read with some background information on it, why it needs to be downloaded separately with STL, and what issues it can fix by enabling/disabling it.

I would also like to add a notifier to note when the selected Steam Linux Runtime is either installed, or when Steam has been told to install it. This won't show anything in SteamOS Game Mode but that's not really our problem :-)

TODO:
- [x] More testing
- [x] Add notifier
- [x] Update langfiles


sonic2kk:
Latest commit added notifier support for the various parts of the SLR installation. This was done by adding an additional flag to the `commandlineFetchGameSLR` function which toggles whether to use the notifier. Doing it here instead of in the GUI function means we have the flexibility to show the notifier at more points during the installation, and to show the notifier when we fail to get information on which SLR we need to install (such as no `tool_manifest.vdf` for the given Proton version).

Langfiles will need updated but this should be good after that.

sonic2kk:
Langfiles updated, version bumped, ShellCheck is green, this is ready  to merge!

PR #880, opened 2023-08-10
Update flatpak .desktop generation

termdisc:
The flatpak version of steamtinkerlaunch resolves its realpath to be `/app/utils/bin/steamtinkerlaunch`, which is incorrectly set in a .desktop shortcut, i.e., `Exec=/app/utils/steamtinkerlaunch/bin/steamtinkerlaunch hedgemodmanager start auto "%u"`

The shortcut should be directed to `flatpak run --command=steamtinkerlaunch com.valvesoftware.Steam <args>` instead.

The changes are directed to adding an `if` condition that changes the `Exec=` line based on whether `[ $INFLATPAK -eq 1 ]`, such that if stl is running in a flatpak environment, the appropriate flatpak command is used

PR #882, opened 2023-08-23
Update Vortex URL handler

termdisc:
Accidentally left a reference to HMM in the Vortex URL handler and an extra `$`.

Fixing my fix from #880


sonic2kk:
Looks good to me, sorry for missing this originally but thanks for the fix! 😄

sonic2kk:
Also just a side note: This change will probably be left out of the changelog since it's a fix for a fix that wasn't in a release yet. Just in case you're wondering why it's not mentioned. But your work is highly appreciated all the same!

termdisc:
Thanks, and sorry for the mistake.

Issue #457, opened 2022-04-23
⚠ steamtinkerlaunch wiki discussion/contributing


zany130:
Have some free time so I'm correctly going through the wiki to make some things clearer or adding relevant links. 

Also as we found out on in several  old issues [such as here](https://github.com/frostworx/steamtinkerlaunch/issues/378) gamemode can break a few options such as GDB and cheat engine so I would like to add notes about that. BTW do you know why so i can add it to the wiki? I'm assuming it's a limitation on how gamemode works 

anyway take a look here for my changes whenever you want https://github.com/zany130/steamtinkerlaunch/wiki/ 

going to have to figure out how I can merge these back into your wiki since I don't think github has pull requests for wikis 😅

zany130:
made a few changes [here](https://github.com/zany130/steamtinkerlaunch/wiki/Wine-Debug) and [here](https://github.com/zany130/steamtinkerlaunch/wiki/Steam-First-Time-Setup) to add a warning about an issue I have encountered where if you use steamtinkerlaunch to delete the proton prefix, steam will overwrite steam cloud saves for that game(mentioned the backup feature though and that it does that by default)

some additions to the [gamemode](https://github.com/zany130/steamtinkerlaunch/wiki/GameMode) page and examples of what you can do with it (especial the scripting function which I think is really nice)

some grammar and spell checking in the [cheatengine](https://github.com/zany130/steamtinkerlaunch/wiki/Cheat-Engine) page as well as a note on my experience with some cheat tables which are now "protected" by a new app (should prob mention that fearless revolution is the main source for cheat tables and worked with upstream I think) and that was all I think, for now at least

EDIT: Oh forgot to mention I also added a link to the WineHQ debug chanels page to [Wine Debug](https://github.com/zany130/steamtinkerlaunch/wiki/Wine-Debug)

also any critic (such as if the above warning was necessary to add) or comments would greatly be appreciated

frostworx:
Oh nice! Thank you very much for improving the wiki! :)
You forgot a "n" in channels in your [Wine Debug note](https://github.com/zany130/steamtinkerlaunch/wiki/Wine-Debug), but apart from that I haven't found anything which should be changed.
Just invited you as collaborator, so you can make the changes directly in the wiki. :)

zany130:
[...]

PR #883, opened 2023-08-23
Enable using a symlinked "Public" user folder along with "steamuser" with the same behavior.

diegoortizmatajira:
# Need for this change
A central copy for my "steamuser" and "Public" user folders allowed me to make backups for my game progress when using multiple proton prefixes. Public folder is used sometimes for some game versions to store data.

I found that I can manage that feature using:
```
## symlink the steamuser directory
USESUSYM="1"
## symlink the steamuser directory globally to one directory
USEGLOBSUSYM="1"
```
But that option only symlinks my "steamuser" folder, so I have added the code also to symlink "Public" user folder.

# Proposed change
With the code I have added, source folders are located in:
```
$HOME/.config/steamtinkerlaunch/proton/steamuser
$HOME/.config/steamtinkerlaunch/proton/Public
```
and both of them can contain a `global` or multiple `$AID` subfolders as required.

Symlinks are created on game prefix inside `pfx/drive_c/users` folder.

Doing so will ensure that both folders are symlinked per game or globally with the same behavior as steamuser.


sonic2kk:
Thanks for opening a PR! This feature sounds very useful, essentially allowing for the `Public` folder in `pfx/drive_c/users` to be shared amongst other prefixes, just like how the `steamuser` symlink works. Awesome work! :-)

The implementation also looks sound to me, it mostly follows what the steamuser symlink does. I am wondering if it might be a good idea in future to create a "shared" function that can manage creating symlinks. Though this can always be tackled later on, it's not like this is followed very well in the rest of the codebase :sweat_smile: 

Having a separate method could make it easier to redirect other folders, though there's a point where that might start to overlap too much with redirect compatdata, so it's really not *that* big of a deal. Plus touching too much of the existing SteamUser symlink stuff could be a bit of a delicate operation.

The explanation for this sounds logical, I can see why this would be desirable. Just wondering, could there be a (reasonable) case where a user may want to symlink the `steamuser` directory but *not* symlink the `Public` directory? Could there be a case where this could be problematic? If so, with the way this is implemented, it would be straightforward to gate this behind a separate checkbox (which could be on by default). But that might not be necessary, please let me know if you think having the ability to toggle this behaviour is warranted.

<hr>

Overall this looks really good! Just that outstanding question over whether this could pose a problem. If you don't think it will or are at least reasonably confident in it, I'm happy to merge this as-is and deal with complaints like that down the road (it should be a straightforward change if it's ever needed).

Thanks again for all your work on this!

sonic2kk (approved):

diegoortizmatajira:
Hi, in regards your question:
> The explanation for this sounds logical, I can see why this would be desirable. Just wondering, could there be a (reasonable) case where a user may want to symlink the `steamuser` directory but _not_ symlink the `Public` directory? Could there be a case where this could be problematic? If so, with the way this is implemented, it would be straightforward to gate this behind a separate checkbox (which could be on by default). But that might not be necessary, please let me know if you think having the ability to toggle this behaviour is warranted.

For most games, it boils down to storing data inside `steamuser` OR `Public` but for reduced tinkering, I usually link both folders in game prefixes, and let the game use whatever it wants to. 

Also, I had to take time to experiment a little before finding this option in Steamthinkerlaunch, as it wasn't entirely clear to me from the Wiki article, I had to track down the code to fully understand how it works. 

Being so, splitting this into two options in settings (for `steamuser` and `PublicN individually) may be even harder to understand for someone new. We can keep them together, and work on it if someone needs it.

If I get the chance, I'll try to add more detailed documentation to Wiki.

sonic2kk:
> Also, I had to take time to experiment a little before finding this option in Steamthinkerlaunch, as it wasn't entirely clear to me from the Wiki article, I had to track down the code to fully understand how it works.

There's lots of room for improvement on the wiki :-) If you look at the page edit history it's probably pretty sparse and/or pretty old. I even had to look up the code for this to understand how this feature fit together (I didn't write the steamuser symlink stuff, I think the previous maintainer did).

This is ideal too imo, since because I didn't write this code, improvements coming from someone who uses this feature and understands how to improve it will always be more useful! 😄 

If symlinking both folders results in less tinkering and if you haven't experienced any issues linking both, then I'm happy with that!

<hr>

For wiki updates, I think GitHub has relatively poor collaboration for wikis. If there's a way for you to "open a PR" so to speak for a wiki improvement that would be awesome, but if GitHub doesn't facilitate this, feel free to submit a markdown document or something along those lines to #457, but there is _no pressure_ to contribute there :-)

diegoortizmatajira:
I had this proposed change applied on `steamthinkerlaunch v12.12` on my machine and it works properly, The only issue happens with `Borderlands` series and `Tiny Tina Wonderlands` displaying:

>  "Your Settings Have Failed to save. Would You Like to Try Saving Again?" 

Then, I had to stop using symbolic links with them. But every other game so far works properly. Anyways is a game issue, not related with steamthinkerlaunch.

Now, It is up to you to merge the change.

sonic2kk:
Thanks a bunch for your contribution to SteamTinkerLaunch, I have updated the changelog to include this change under the "Features" section: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog

I think others find this feature useful as well :-)

PR #887, opened 2023-08-27
langfiles: Update Steam Deck Compatibility localisation strings

sonic2kk:
This PR is a work-in-progress that will update the language file strings for the other languages that SteamTinkerLaunch supports, to improve localisation.

These strings are taken from SteamDatabase's [extracted localization data](https://github.com/SteamDatabase/SteamTracking/tree/master/ClientExtracted/steamui/localization) from the Steam Client. To make the process a bit less painful, I created a Python string that can take the input strings copypasted from the `shared_langname.json` file on the SteamDatabase repo, and convert them into the language string format that SteamTinkerLaunch expects.

```python
with open('strings.txt') as instrings:
    for s in instrings:
        keyval_split = s.split(':')
        
        string_key = keyval_split[0].strip()
        string_val = keyval_split[1].strip()[:-1]
        
        formatted_string_key = string_key.replace('"', '').upper()
        
        print(f'{formatted_string_key}={string_val}')
```

This will be used to put the data for other languages that SteamTinkerLaunch supports into their respective language files.

TODO:
- [x] Other languages
- [x] Ensure all strings end with full stops
- [x] Check that all langfile syntax is fine and not broken


sonic2kk:
It turns out having a full stop at the end of Steam strings is inconsistent between languages and even some strings in the same file are missing them, so there are inconsistencies there too.

For consistency here, I will manually check the Steam Deck Compatibility strings to ensure they all end in a full stop before merging.

sonic2kk:
Chinese will be left out of this PR because I am not sure which of the Steam Chinese files to use. The language strings for Chinese are mostly up-to-date anyway thanks to the efforts of community contributors, so for the most part the strings are translated already except for the new ones in #887.

sonic2kk:
Punctuation should be consistent now, and fixed up a couple of issues where the updated strings were not ended properly.

Things should be ready for this PR now.

PR #888, opened 2023-08-27
ReShade: Combine USERESHADE and INSTALL_RESHADE into one option

sonic2kk:
Implements the remaining feature request from #873, meaning the only required work left on that issue is tracking down a bug or two.

This PR combines the two ReShade checkboxes, `USERESHADE` and `INSTALL_RESHADE`, into one checkbox. This makes for a better UX. We also have some migration logic, so that if `INSTALL_RESHADE` was enabled before, we migrate that over to `USERESHADE` so that they are equivalent. This should make the transition easier with little to no intervention in the Game Menu needed to re-eanble ReShade.

The `INSTALL_RESHADE` checkbox existed before to allow the user more granular control over whether ReShade should be installed. A user could enable it, but if ReShade was not installed, nothing would happen. They would have to explicitly tell SteamTinkerLaunch that they want to use ReShade. Likewise, a user could choose to install ReShade but not use it with that game.

There is no much value anymore in my opinion in having these options, and even when developing #881 I found this somewhat confusing. So this PR combines the two options into one. If ReShade does not exist, it will be downloaded.

All existing functionality should remain untouched, this PR just removes a step in enabling ReShade.

TODO:
- [ ] Further testing
- [x] Update langfiles


sonic2kk:
Updated all langfiles to remove references to `INSTALL_RESHADE`, and updated English and English UK to use slightly different wording for `USERESHADE` strings.

Pending further testing, this should be ready to merge.

sonic2kk:
Going to merge this and wait for feedback in #873, to avoid keeping this PR in limbo :slightly_smiling_face: And also because I am impatient.

PR #895, opened 2023-09-05
GameScope: Add chckbox for ENABLE_GAMESCOPE_WSI environment variable

sonic2kk:
This can be used by DXVK v2.3 for some Vulkan optimisations when disabling vsync, as well as some HDR options within GameScope it looks like.

This still needs testing.

TODO:
- [x] Test this change (i.e. make sure it doesn't break any existing GameScope stuff, can't actually test the functionality)
- [x] Update langfiles


sonic2kk:
Tested and the option works, doesn't seem to cause regressions when toggling other options or other STL Game Menu options (as those impact game config, global conf should be entirely unaffected).

Did not test if this breaks any games yet. Once this is confirmed to be safe, I will merge.

sonic2kk:
Pending a version bump, this change looks good to go. Tested with DOOM 3, GameScope appeared to work as expected and the option did not break anything (though I was not using DXVK v2.3). Seems to work as expected, at least with the option simply being an environment variable toggle before GameScope launch.

Issue #893, opened 2023-09-05
tracked_files needs first run in order to be created in prefix; Special K still shows warning, but works fine if vcrun and d3dcompiler_47 are installed through winetricks


CartoonFan:
## System Information
- SteamTinkerLaunch version: v14.0.20230903-1
- Distribution: Arch Linux
- Installation Method: AUR

## Issue Description
First off, I apologize: this issue was mostly due to me deleting steam_api64.dll (which NieR: Automata--the game I was testing--requires to **not crash**), thinking it was a part of the Special K installation. The next big problem was the tracked_files file not appearing in the game prefix, which happens if you don't do a first run beforehand to create it in there. Not sure if that last part's a bug or not--I feel like STL should create it if it's not there--but overall, it's probably not a huge issue by itself. 

A slightly more annoying thing is that vcrun2022 and d3dcompiler_47 have to be installed through Winetricks (the former for Special K, and the latter for both Special K and ReShade), which makes the launch process **much** longer. The warning about Special K not finding the specific dll it wants still shows up, and the setup process for it doesn't trigger, but installing both vcrun2022 and d3dcompiler_47 allow it to run properly. Again, I'm not sure if that's a bug (it feels like it) or not. I think it took two launches for the Winetricks verbs to actually install, as well, making the process take even longer. As far as I can tell, this needs to be done for any game that uses Special K.

## Logs
[524220_gamelaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/12525088/524220.log)
[steam-524220.log](https://github.com/sonic2kk/steamtinkerlaunch/files/12525093/steam-524220.log)
[524220_steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/12525118/524220_steamtinkerlaunch.log)

sonic2kk:
I saw your initial issue earlier with your "Ted Talk", I apologise that I didn't get a chance to fully dig into it (took a nap after work to beat a heatwave). But I am assuming that all of that from earlier is now resolved, and that the core of the issue is what you've described here, which should boil down to:
1. `tracked_files` won't exist in the game prefix (i.e. `/home/Gaben/Games/steamapps/compatdata/524220`) until the game is launched at least once, and this causes Special K to fail (because it relies on `vcrun2022`).
[...]

PR #898, opened 2023-09-06
Special K: Remove dxsetup and vc2022 installation

sonic2kk:
Should resolve the final parts of #893. In short, this PR should make SpecialK work out-of-the-box now.

When installing Special K, SteamTinkerLaunch performs two additional steps: Install `dxsetup` and `vc2022` (previously `vc2019` before some recent changes). It installs these packages by using the Steamworks Shared packages to do so, which are the packages Steam uses to install these components when required by a game.

These must have been required previously, but in testing, I found that `vc2022` specifically causes games to crash. When this package is **not** installed, SpecialK works as normal. `dxsetup` will be installed if required by a game, as this sets up Direct3D.

Currently on master, for Special K to work, one must install the `vcredist2022` package. This creates a duplicate prefix for some reason, and I suspect this is why SpecialK works during this step. When the prefix is entirely unmodified as part of the SpecialK installation (that is, `dxsetup` and `vc2022` are not installed), SpecialK works with our existing logic which is mostly just fetching the correct DLLs/INI files and putting them into the same folder as the game executable.

This PR also does two other minor things:
1. Adds a notifier for when SpecialK is being installed (currently a hardcoded string, not in the langfiles yet)
2. Removes the `d3dcompiler_47` DLL installation, as this is not required for SpecialK to run -- It *is* required to use SpecialK and ReShade (which is problematic but that's for another time), but STL will install this as part of its ReShade installation, so there is no case where this DLL has to be installed just to use SpecialK. 

I tested this with NieR:Automata and it works. I will test NieR:Replicant as it is the only other SpecialK compatible game I am willing to test with at the moment.

TODO:
- [x] Test with more games
- [x] Remove commented code
- [x] Put notifier string in langfiles


sonic2kk:
User confirmed in #893 that this resolves the remaining issues with installing and using SpecialK (aside from problems with ReShade+SpecialK, which are tracked separately at #894).

I'm going to test NieR:Replicant with SpecialK and see if the install process works for it, too. So far NieR:Automata is the main game that has been tested, and NieR:Replicant is the only other one I'm comfortable testing for now. If issues crop up we can address them separately.

It's worth noting that some parts of SpecialK may not work, but this is likely outside of the scope of SteamTinkerLaunch (ex: the FPS cap remover for NieR:Automata). It could be the case that SpecialK, for compatibility reasons perhaps, disables some tweaks for compatibility on Linux.

sonic2kk:
Everything except for a version bump is ready for this PR to be merged, just waiting on NieR:Replicant to download so I can test it :-) Once I confirm SpecialK installs correctly with it, I will bump the version and merge this.

sonic2kk:
Confirmed that this worked with NieR:Replicant. Ready to merge after a version bump!

Issue #738, opened 2023-02-10
Add Poster, Banner, Logo, etc to steamtinkerlaunch add Non-Steam game as launch options


trentondyck:
## System Information
- SteamTinkerLaunch version: N/A
- Distribution: Steam Deck

## Feature Description

@sonic2kk Since we have app id, this also unlocks the ability to add poster, banner, logo per app. Would be a nice addition to steamtinkerlaunch:

Working example in bash/python:

https://github.com/trentondyck/horizon_scripts/blob/70bd701f06596ff4f4f2fa96054abc6a7c03a0a5/install-or-update-horizon.sh#L107

Context:

https://github.com/sonic2kk/steamtinkerlaunch/issues/729

sonic2kk:
This would be great, but it looks like it relies on Python to get the AppID for the Non-Steam Games? Otherwise nice find!

Also wondering that even though it relies on Python, can it generate the AppID? The discussion in ProtonUp-Qt that I was involved in ended up just parsing it instead of calculating it.

SteamTinkerLaunch is a Bash script and will not use Python in any capacity, so we need a way to calculate  this from Bash. It also probably has to be written out to the shortcuts file.

As an aside, if we can generate this with pure Bash, this can also be included in the Add Non-Steam Game UI, which I originally tried and failed to do in a PR several months ago.

(This specifically refers to Non-Steam Games it seems so just editing the title of your proposal to reflect this 😄)

trentondyck:
It does. Line 104 parses shortcuts.vdf and returns appid. 

That's too bad about python, seems like the correct way to parse these files, though it does have the requirement of both python and the vdf module which I assume is what you're trying to avoid.

If we can pick apart https://github.com/ValvePython/vdf/blob/d76292623e326fb165fe3bdb684832cdf30959d4/vdf/__init__.py 
maybe there's a way.

sonic2kk:
> It does. Line 104 parses shortcuts.vdf and returns appid.

That sounds like reading, not calculating :-) We need to calculate the AppID, since we are adding artwork etc before Steam calcuates the AppID. Please let me know if I'm misunderstanding here 

> seems like the correct way to parse these files

[...]

PR #902, opened 2023-09-14
Add Non-Steam Game: Fix Non-Steam AppID Generation

sonic2kk:
Part of work for #738, at long last!

We need to generate a 4byte little endian hex from a signed 32bit integer. This is what gets written to the shortcuts.vdf file. Steam will reject the AppID if it is not in this format, and will overwrite it.

This fixes the AppID generation, allowing us to 'predict' the AppID and will allow for implementing Non-Steam Game artwork selection.

## Implementation
*(A more thorough explanation of how Steam stores Non-Steam AppIDs can be found at https://github.com/sonic2kk/steamtinkerlaunch/issues/738#issuecomment-1720263749)*

It was discovered that the shortcut AppID is not actually generated by Steam itself, it can be anything so long as it is written out to the `shortcuts.vdf` file as a 4byte little endian hexadecimal representation of a signed 32bit integer. We aren't currently doing this, we're generating a random integer, converting it to hex, and attempting to write that out, which is why the AppID is always overwritten.

To fix this, essentially we have to generate a random nine-digit negative number and write that out as little-endian hex. Numerous functions were created to do this, but I'll explain the overall process here.

To start, we use `shuf` with a seed based on an optional argument to generate our signed 32bit integer. For our seed, we pass it a seed of the EXE name and path to it (ex: `Bad Piggies/home/Gaben/Games/Bad Piggies`), though the seed is an optional component to this function. Since `shuf` can't have a signed range, we cheat and add a `-` in front. This AppID is the decimal representation of what gets stored as a shortcut AppID in the `shortcuts.vdf` file (this is how the Python VDF library displays it, for example).

Once we have this signed integer, we need to convert it to a 4byte little-endian hexadecimal number. Since the `shortcuts.vdf` file is a binary file, it needs the value in hex representation. We first convert the signed integer to hex, and then convert that hex from big endian to little endian (which is basically the reverse byte order, this [Integer Encoder](https://cryptii.com/pipes/integer-encoder) website is very useful for checking out converting integers been signed, unsigned, hex, big endian, and little endian).

This value then needs some formatting to prefix each byte with `\x`, which Frostworx already wrote code for. This ensures it's written out to the binary file correctly. Once we do this, we've successfully written out our own custom AppID to the `shortcuts.vdf` file!

One final step which is useful for #738 specifically, is converting from our initial signed integer to unsigned. This unsigned integer is the one used to set the game artwork in the `/path/to/steamroot/userdata/<userid>/config/grid` folder, allowing that feature proposal to be implemented since we now have a knowable AppID.

## Future Implementations
This opens the door to various other features, one main feature that comes to mind is the ability to remove game configs for Non-Steam games which are no longer on Steam. For example adding a game shortcut to Steam, using STL with it, and then removing the game. That shortcut is going to hang around!

We can check if AppIDs over a certain length are in the `shortcuts.vdf` file, and if not, we can assume they were removed shortcuts, and remove these config files. We could also implement something similar for `compatdata` directories, as Steam does not clean these up. These compatdata directories are always stored on the root Steam install's `steamapps`, so we can simply check for and remove these folders.

Of course, both of these features should be implemented as commands with warnings, as checking based on AppID length is not the strongest test, but it should work for the meantime.

<hr>

Pending a tiny bit of further testing on a last-minute change to appease ShellCheck, and a version bump, this is ready to be merged!

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230915-1 (ansg-writeout-appid)"
+PROGVERS="v14.0.20230915-1"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230912-1"
+PROGVERS="v14.0.20230915-1 (ansg-writeout-appid)"
 PROGCMD="${0##*/}"
 SHOSTL="stl"
 GHURL="https://github.com"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Fixed a few silly mistakes which I forgot to refactor and pushed the version bump. As long as ShellCheck is still green, this will be good to merge now!

PR #576, opened 2022-09-05
[WIP] Add more game artwork paths to Non-Steam Game UI

sonic2kk:
## Overview
This PR adds the ability to set more game artwork (background, logo, boxart) when adding Non-Steam Games. It also has some minor formatting changes to the Add Non-Steam Game GUI.

image

The current file picker on Steam for Linux is tedious to use, so adding these paths manually can be really annoying. Being able to do it from a GTK file picker is a lot more user-friendly.

## Implementation
These files are set by copying them to `$STUIDPATH/config/grid` with appropriate names. This is where Steam reads them from if you set the manually, they aren't written to any VDF file from what I can see. Manually setting artwork for regular Steam games also writes them here.

Steam expects the files for the games to be named in a specific way.

- Background/Hero files are named `<app_id>_hero`
- Logos  are named `<app_id>_logo`
- Box art is named `<app_id>p`
- Small artwork is named `<app_id>` with no special suffix (this is the artwork that shows up when the game is the most recently played)

## Problems
This PR is marked as WIP because currently there's a big problem. The files are copied over to the correct folder with a set AppID, but **this AppID might be wrong**. When opening Steam after adding a shortcut, no game art shows up, and when you run the game, the AppID that shows up for the Proton prefix is different from the one that the files are given when they're copied over to the `grid` folder.

There should be a reliable way of doing this, as [Steam-Rom-Manager](https://github.com/SteamGridDB/steam-rom-manager) is able to do this when adding Non-Steam shortcuts (I've verified this myself by using it on Steam Deck). I'm not sure how it does this though. The AppID might only change when the game runs with Proton, which may mean we have to set STL as the compatibility tool (**if** it's an EXE) and *then* update the AppID? 

Aside from the language files, this is the main thing holding this PR back. I decided to open the PR in its current state to figure out a good way of getting the AppID. If we do find a way, we might have to move the artwork copying to the bottom of `addNonSteamGame`, so I'll try to remember to do that too :sweat_smile: 

## Todo
- [ ] Get Steam AppID
- [ ] Probably change the wording of the strings in the language files to use better terms, e.g. "Background/Hero" and better explain what the small artwork is
- [ ] Update all language files

## Considerations
I have no idea how this works with Flatpak, so I'm hoping however `$STUIDPATH` is gotten works on Flatpak too :sweat_smile: 

No rush on merging this or anything, but it's something I was tinkering around with and wanted to get some guidance on further implementation. Thanks! :smile:

PR #903, opened 2023-09-15
Add Non-Steam Game: Implement Artwork Selection Paths

sonic2kk:
Implements #738.
Salvages #576.

## Overview
This PR adds the ability to select artwork (hero, logo, boxart, tenfoot) for a Non-Steam Game from the Add Non-Steam Game GUI. Thanks to #902 we properly generate and write out a Non-Steam Game AppID, so this feature request is now possible after over a year of investigation work!

image

I tested this with adding Bad Piggies with some grids I downloaded from SteamGrid, and it worked like a charm.

image

## Implementation
Thanks to #902, we can now generate a correct AppID. This made implementation of this feature trivial, essentially moving over most of the work from #576 for the GUI portion (including the headings, etc).

The GUI was changed to allow users to select Non-Steam Game artwork. This opens a file selection dialogue where they can select their images. These are stored as paths. This UI portion was also essentially copied from the Set Game Artwork functionality from #757, the language strings were directly re-used and the variable names were made to match.

The `addNonSteamGameGui` function takes in all these parameters, and passes them to `addNonSteamGame`, which is a function also used for the `steamtinkerlaunch addnonsteamgame` command. The work done in this PR to allow Non-Steam Game artwork selection also applies to the commandline usage as a result! We simply call `addNonSteamGame` and pass it some extra flags now: `-hr`, `-lg`, `-ba`, `-tf` -- By design, identical to how we implemented Set Game Artwork, for consistency and easier maintenance.

The `addNonSteamGame` function was updated to parse these new arguments, and uses them to call `setGameArt`. We basically just take in the paths from the user in `addNonSteamGameGui`, send them to `addNonSteamGame`, and this in turn sends them (with the generated AppID) to `setGameArtwork`. This function can manage whether the images exist or not, so we don't have to worry about vetting this here -- Woohoo, abstraction!

This PR more or less brings the functionality of `setGameArtGui` and puts it into `addNonSteamGameGui`, where each of these functions' respective commandline functions handles the rest.

There were a couple of other minor tweaks and discoveries made along the way, like how we can pass a string `"--flag"` to functions and they'll parse this as a flag. This means we can get rid of some nasty `if` blocks around the codebase (I think One-Time Run and Custom Command logic may benefit here, but setGameArt will benefit a lot too). This can all go in a separate PR, though :slightly_smiling_face: 

<hr>

In my testing, this PR is feature-complete, and only has a couple of things left:
- [x] Update addnonsteamgame help screen text
- [x] Minor UI text updates
- [x] Update langfiles
- [x] Version bump


sonic2kk:
Version bumped, langfiles updated, code changes are ready, and shellcheck is good. Ready to merge!

PR #910, opened 2023-09-20
Game Menu: Replace vkBasalt checkbox with generic Vulkan Post Processing dropdown

sonic2kk:
Replaces the "Use vkBasalt" checkbox on the Game Menu with a dropdown which can contain various Vulkan post-processing options. The primarily motivation for this is have a nicer UI for future Vulkan post-processing tools in future (namely [vkShade](https://github.com/ralgar/vkShade), without having to extend the game menu with a bunch of checkboxes.

image

As a cool trick, we use `eval` to export the variable, since right now we only use vkBasalt, and in future we may use vkShade. The export names are just `ENABLE_<toolname>`, and since `<toolname>` is what we store in the game config, we can just eval an export string like `eval "export ENABLE_VKBASALT=1"`.

In future, if there are other Vulkan post-processors that require more, we can do some conditional checks to see which post processor is selected, and leave this as our fallback.

There is intentionally no `migrateCfg` option, as the checkbox for toggling vkBasalt directly controlled `ENABLE_VKBASALT`. It wasn't stored in the game config, and I didn't want to run the risk of modifying users' custom environment variables. Therefore, users will have to re-enable vkBasalt using this dropdown, which is `none` by default.

vkShade is not included in the dropdown yet because it is not available without building from source yet. Once a package is at least available on the AUR or Chaotic-AUR, I will include it. The refactor that this PR introduces will make it trivial to introduce though; just add `vkShade` to the Vulkan Post Processor dropdown. 

TODO:
- [x] Logging
- [x] More testing
- [x] Test Non-Steam Games
- [x] Update langfiles


sonic2kk:
`ENABLE_VKBASALT` seems to get exported correctly for Steam games, tested with Cookie Clicker. Tested with MangoHUD and adding `vkbasalt` as a config option.

Tested "Touhou Fumo Racing" as a custom command, it displays vkBasalt as working as well.

For both games, when the post processor is `none`, vkBasalt is displayed as off, as expected.

sonic2kk:
Testing is a success on this. Shellcheck looks good on this, langfiles updated. This should make it easier to add vkShade in future :-)

sonic2kk:
Noticed a typo in the commit name too late, shit. Oh well.

PR #912, opened 2023-09-21
Special K: Initial logic for choosing SpecialK DLL name

### LINKED ISSUE #894: ReShade compatibility with Special K

sonic2kk:
Some work for #894.

## Overview
This PR adds a combobox entry dropdown where a user can enter a custom name for the SpecialK DLL. Similar to the ReShade DLL name (#881), this is useful to fix issues where the wrong DLL name may be chosen, as well as to avoid conflicts with ReShade itself or other DLLs which have conflicting names with SpecialK.

Screenshot_20230921_231507

The options in the combobox are `auto`, `d3d11`, `dxgi`, `d3d9`, `opengl32`. The `auto` option is the default value, and will fall through to the existing logic of picking the SpecialK DLL based on the rendering backend. The other options will, as expected, set the DLL name to the specific value chosen. The user can also enter their own DLL name, though just like with ReShade, a manual DLL override will need to be set so that the game knows to use it.

## Remaining Work
There are some features I would like to implement as part of this PR.

### Removing Old DLLs
~~I would like to have a way to clear out old SpecialK DLLs once the name changes, similar to what we do for ReShade. If the user selects `auto` and it copies over `d3d11.dll`, then they find out that doesn't work and switch to `dxgi.dll`, STL should clear out the previous DLL and copy over the new selected  `dxgi.dll`. We should be able to do this feasibly as we track the installed DLLs with `SpecialK_Enabled.txt` very similarly to how we handle ReShade. There's probably room for a broader refactor here to make the SpecialK and ReShade DLL-related logic more shared, but that can come later.~~

We didn't implement this for #881, so we don't add it here.

### DLL Naming Conflicts
Another feature I would like to add is some way to manage DLL name conflicts. ReShade goes to great lengths for this but if ReShade and SpecialK conflicted, we should catch that first. For example, if both ReShade and SpecialK tried to use `dxgi,dll`, we should handle this somehow.

We should also consider cases where ReShade/SpecialK were installed first, and how to manage the DLL naming conflicts. In future, we will add an option to not load ReShade as a SpecialK plugin, so we should account for this scenario. In general there could already be a scenario where the DLL names conflict such as with a mod manager, ReShade is just one example. Here's two scenarios:
- If ReShade is already installed and has its custom DLLs `dxgi.dll,d3d9.dll`. We then try to install SpecialK, even just with `auto`, and it conflicts with the existing ReShade DLLs.
- If SpecialK is already installed and uses the name `dxgi.dll`, then we try to install ReShade which by default uses `dxgi.dll` by default. This will be a conflict.

I am unsure how to handle this at time of writing. Perhaps the best solution is to simply not do anything except log/show notifier, and document this on the wiki.

### Appending `.dll` Extension
A straightforward one, make sure the chosen SpecialK DLL always ends with `.dll`, to avoid problems with copying over a file with an invalid/nonexistent extension. We do the same thing for ReShade.

<hr>

TODO:
- [x] Investigate some basic logic to catch if there is a conflict between the chosen SpecialK name and the EXISTING ReShade DLL name(s)
- [x] Investigate some basic logic to catch if there is a conflict between the ReShade DLL name(s) and the EXISTING SpecialK DLL name
- [x] Make sure SpecialK DLL names end with `.dll`.
- [x] Update langfiles


sonic2kk:
Seems like we didn't add any logic in #881 to track renamed ReShade DLLs, so we don't do it here for SpecialK.

sonic2kk:
Added basic logic to check for single ReShade DLL name conflict, however this won't work when there are multiple ReShade DLL names chosen. We should add that.

sonic2kk:
The ReShade+SpecialK DLL name checking logic should be useful in future when we add the option to load ReShade and SpecialK independently, without loading ReShade as a SpecialK plugin.

sonic2kk:
Improved the logic on the SpecialK side for checking for these DLL name conflicts. Now we will check all ReShade DLL names for conflicts and we won't install SpecialK over ReShade if we find a conflict.

We also try to backup existing DLLs with conflicting names which seemingly aren't SpecialK, or at least aren't a tracked SteamTinkerLaunch SpecialK installation (similar to what we do with ReShade in `installRSdll`). If we can't, and the `<dllname>.bak` filename already exists, we simply abort install.

We need similar logic to this on the ReShade side, and then some testing to ensure basic functionality still works. After that, this is good to merge.

sonic2kk:
DLL conflict check is now in place on the ReShade side. Aside from the usual langfiles and version bump, we need  to do some more testing before this is ready.

sonic2kk:
Dear god, this turned into a massive refactor of how SpecialK is handled. I did not expect this change to take this much work, but here we are.

The latest commit did... a lot of changing, and we now cover several cases for ReShade DLL name conflicts. In developing this, I covered basically every test case I could think of, and normal happy-path installation still works as expected (and SpecialK won't reinstall each time, I made sure that this was the case).

sonic2kk:
This is ready to be merged.

Issue #900, opened 2023-09-08
Information request on a MO2 cosmetic issue font size and moving MO2 compatdata to save on disk space.


tetoNidan:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: Garuda

## Feature Description

So this request isn't so much of a feature request as it is a request for some advice. 

1. MO2's font is so small I cant read it without pressing my nose to the monitor. What would be the recommended method to increase the font size? I cant remember if you can do it in winecfg, change system wide font scale I mean? Will adjusting the font size in the winecfg STL game menu work? It is using the PFX for the game not the MO2 PFX right, so I would have to manually set the PFX with WINEPREFIX? Also if I do that mono will be installed for the winecfg GUI, will that brake anything? Asking because after I installed mono a long time ago to the MO2 PFX I lost the text to LOOT. At least that is what I think broke the text output to LOOT, really the only thing I can think of that I did to that MO2 PFX. I could be wrong about that though. It was also over a year or so ago but I was never able to fix that. I also stumbled onto this forum that said I could [use a different MO2 theme](https://stepmodifications.org/forum/topic/14088-basic-font-tiny/) but I haven't been so bold as to try it yet. Any advice is appreciated.

4. Can I just move my MO2 compatdata folder to an external drive? I currently have it installed to my $USER in it's native location. I have limited space left as I have 2 heavily modded games SKAE, FO4 and probably soon Starfield. 

I was wondering the best way to go about this. Should I move the compatdata via the Global MO2 compatdata location setting in STL? Should I reinstall MO2? Maybe I should just use a symbolic link `ln -s source dest`?

Thanks sonic2kk!

sonic2kk:
The font scaling issue should be possible to fix using the DPI scaling option in Winecfg, under its "Graphics" tab :-)

![image](https://github.com/sonic2kk/steamtinkerlaunch/assets/7917345/76d17766-00a9-4dbc-9468-63a676fa0b1c)

[...]

PR #915, opened 2023-09-23
Add option to set Per-Game and Global Wine DPI value

sonic2kk:
This topic was previously discussed in #900, and the ideas discussed there have now been implemented more broadly.

## Overview
This PR adds an option to configure a Global and Per-Game Wine DPI scaling value. This option is available for configuration from Winecfg, but this PR exposes it on the SteamTinkerLaunch GUI on the Global Menu (Global Scale) and Game Menu (Per-Game Scale).

Effectively, this option allows for increasing the font size, and thus the rendering size of a given window under Wine. This option is useful when the text that an application may display is too small to be read comfortably, MO2 suffers from this in my experience and others' experienced as well (though it may be configurable from a theme). This option is also useful for people who are using fraction scaling on Wayland with a multi-monitor setup, and have unscaled monitors. In this scenario, this option complements KDE Plasma Wayland's now-default legacy application scaling setting, "Apply Scaling Themselves".

I implemented this option because I pretty much needed it when playing Oblivion with ModOrganizer 2, as not only was Oblivion's launcher rendered incorrectly on my displays (needed to bump the resolution to 3840x2160 once), so was the ModOrganizer 2 UI. 

## Implementation
The Wine DPI scaling can be configured from Winecfg, but for convenience we now exposed the option on the UI. To set the scaling value, we can write directly out to the current game prefix's registry. We write under `HKEY_CURRENT_USER\\Control Panel\\Desktop` to the key `LogPixels`, and write out the value in decimal (the `reg` command can convert it to hexadecimal, which the registry actually stores).

We write this value out on launch if the DPI scale value is enabled, however, we can't unset this. There is no good way to know when the DPI value should be updated without creating a messy system to track the previous preferences state and so on. Therefore, this value is applied permanently to the prefix, though a user is free to modify the value back. Just like with Winecfg, the user is expected to know how to tinker with this properly. We also offer ways to directly access Winecfg should the need arise through a checkbox on the Game Menu to launch Winecfg before game launch, as well as on the One-Time Run menu.

## Future Work
In future, it may be worth exposing this option for third-party tools which have their own prefixes, such as MO2 and Vortex Mod Manager.

We should create a wiki page, perhaps a general section for limitations, workarounds, and tips when running under Wayland, as this option is useful as well in combination with Steam's lesser-known `STEAM_FORCE_DESKTOPUI_SCALING=1.5` value. There are also some funky Yad behaviours in this mode as well which are annoying but at the moment have no workarounds. The notes about how this DPI value is applied should also be documented on this wiki, as well as how to change these values back if a user is stuck.

I wrote out a big PR description for this, and when I tried to create the PR, it disappeared into the aether. I have no idea where the PR went, but it's disappeared, and so too has my motivation to describe this feature in as much detail again :sweat: 

TODO:
- [x] More testing to ensure Global Scale applies properly
- [x] More testing to ensure Per-Game Scale applies and overrides Global Scale
- [x] More testing to ensure this setting is only applied when the settings are enabled
- [x] Update wiki with information about this, including that this may not work as expected with all applications depending on how the deal with reported DPI
- [x] Update langfiles


sonic2kk:
Untested, but this option may not work for Vortex. We may need to pass the argument `--force-device-scale-factor=<val> to the executable independent of this scale factor.

This `--force-device-scale-factor` argument is used for Electron applications. It is generally only needed for very old native Linux Electron applications, or Electron applications running under Wine (such as Cookie Clicker or Vortex Mod Manager).

sonic2kk:
A comprehensive wiki page covering scaling options for various different applications, toolkits, Wine, and SteamTinkerLaunch itself has been created: [HiDPI Scaling wiki page](https://github.com/sonic2kk/steamtinkerlaunch/wiki/HiDPI-Scaling).

sonic2kk:
This all appears to be working in testing. Per-Game Scale overrides Global Scale correctly, and each option works individually. When the Global Scale is disabled, and a new game is launched, it will not pass any DPI scaling.

Probably this is ready to merge after some more testing, and a langfile update.

sonic2kk:
Apparently I was setting `USEGLOBALWINEDPI` to `DEFWINEDPI` by default, this is what I get for copypasting...

This didn't seem to cause issues, as I guess any value not 0 or 1 was defaulting to 0, since this option was off by default in my testing.

sonic2kk:
Shellcheck is good, testing is good, I think this is ready!

PR #917, opened 2023-09-24
Special K: Add option to specify Global SpecialK ReShade Version Override

### LINKED ISSUE #894: ReShade compatibility with Special K

sonic2kk:
Further work for #894.

## Overview
Adds the ability to override the ReShade version when using ReShade and SpecialK together, in the interest of improving compatibility. In initial testing for #894, it was found that ReShade 5.4.2 is generally more compatible when loading ReShade as a SpecialK plugin. Therefore, this PR adds the option to do this, and enforces a default of `5.4.2`. We can bump this and allow users to change this in future, should compatibility improve.

There is also a separate option to disable this feature, if the user doesn't want to use any override version anymore and simply wants to use the global ReShade version by default with SpecialK.

image

This option does not take precedence over the per-game ReShade Override. Therefore, if a certain game has more compatibility when using ReShade+SpecialK together, the override will still apply. Also, this option will not apply if ReShade is used on its own, or if SpecialK is used on its own. It will only apply if both are used together.

The order of precedence with ReShade versions is as follows:
1. ReShade Override Version on Game Menu -- This will always apply when ReShade is enabled, and takes the **most priority.**
2. SpecialK ReShade Version (this PR) -- This will take priority over the Global ReShade version, but not over the ReShade Override version.
3. Global ReShade version -- This is the default, base value for which ReShade version to use.

## Future Considerations
In future, we should make sure this option is IGNORED when the option to use ReShade+SpecialK independently is added. This option should really only apply if ReShade is loaded as a SpecialK plugin.

<hr>

TODO:
- [x] Testing with some more games
- [x] Some light regression testing to make sure this option is properly ignored when not enabled
- [x] Update wiki documentation on this feature and the order of precedence, under the "ReShade and SpecialK" section for both the ReShade wiki page and SpecialK wiki page
- [x] Update langfiles


sonic2kk:
Fixed a regression from #912 which caused `SpecialK_enabled.txt` to lose its tracked DLLs. We remove this file so we can update it with only our tracked SpecialK DLLs (so we don't end up with duplicates and/or entries for renamed SpecialK DLLs), but we didn't write out to it if `SHOULDINSTALLSPEK -eq 1` - this is `0` if SpecialK is already installed.

This has been fixed in 2c1a16c.

sonic2kk:
I tested this PR with HoloCure and it seemed to work. The correct ReShade version was used only when SpecialK was enabled, and I could change, override, and disable this.

This PR should be ready for merging soon.

sonic2kk:
[SpecialK](https://github.com/sonic2kk/steamtinkerlaunch/wiki/SpecialK#reshade-and-specialk) and [ReShade](https://github.com/sonic2kk/steamtinkerlaunch/wiki/ReShade#using-reshade-and-specialk-together) wiki pages have been updated to note this new option, its default behaviour, how to change it, and the order of precedence for ReShade versions.

Bumped version, updated langfiles, Shellcheck is good, this is ready to merge.

PR #918, opened 2023-09-24
GameScope: Add support for selecting ReShade Effects

sonic2kk:
Newer GameScope releases have support for selecting a ReShade effect file to load and display over the game. This doesn't actually use ReShade, GameScope now just has a way of compiling ReShade shaders and using it on the compositor itself, rather than the game.

This PR adds two new options to our GameScope GUI which expose these options. This is under the "Filtering" options as I felt it was nnecessary to create a new category for these options. If ReShade support in GameScope matures enough to have several options which warrant a standalone heading, we'll add it in future.

image

This support in GameScope upstream is still in its infancy, but this is a good starting point for us downstream and better than having no option.

The `--reshade-technique-idx` is something I don't really understand, but from looking at the code, it appears that it will reject blank values or `0` values, so we do that here. It also appears that since it uses `std::atoi`, that it expects an integer value, hence why I went with a spinner for this option.

<hr>

TODO:
- [x] Test to ensure menu order is still working fine
- [ ] Option somewhere to download [Joshua Ashton's GameScope Shaders](https://github.com/Joshua-Ashton/GamescopeShaders) repo?
- [x] Update langfiles


sonic2kk:
Menu order seems to still work fine.

sonic2kk:
Adding the other shader repo will come another time in a different push/PR. For now, this feature is ready!

PR #919, opened 2023-09-24
Special K: Add toggle for loading ReShade as Special K Plugin

### LINKED ISSUE #894: ReShade compatibility with Special K

sonic2kk:
Another piece for #894.

## Overview
Currently, when ReShade and SpecialK are enabled together, SteamTinkerLaunch does some magic and takes a different install path for ReShade. We copy over the DLLs as `ReShade32.dll`/`ReShade64.dll` instead of using the DLL name(s). We take an entirely different install route and name the DLLs this way so that SpecialK can find the ReShade DLLs names and load them *instead* of the game doing it. I am not sure why this is the default but it seems to work and be the more desirable behaviour, so we'll leave it as the default.

However, for compatibility, sometimes a user might want to load ReShade and Special K separately, where they are entirely independent of each other, and it's up to the game to load the DLLs. Some newer versions of ReShade don't work with SpecialK in some games (NieR:Automata and NieR:Replicant don't support versions of ReShade above 5.4.2 when using it as a SpecialK plugin, but iirc *do* work when both are loaded independently).

This PR adds an option to toggle whether or not ReShade is loaded as a SpecialK plugin to the Game Menu, grouped with the SpecialK options.

## Future Work
For now, the toggle won't work for existing installs, it only works for clean SpecialK installs with no ReShade installed yet. If a user installs ReShade as a SpecialK plugin, then disables the option and launches the game again, the existing ReShade DLLs for SpecialK are NOT removed. This feature could potentially be added as part of this PR, to make the toggling more useful.

This feature could also go in a future PR where we clean out existing ReShade installs when SpecialK is enabled. That PR could be a general one that adds better DLL management when using ReShade+SpecialK.

<hr>

This PR is a draft as it has had minimal testing.

TODO:
- [x] More testing
- [ ] Potentially add toggle logic to check for and clear ReShade installs automatically if required (remove non-plugin or plugin ReShade installs depending on status of checkbox) 
- [ ] Test ReShade+SpecialK existing DLL conflict logic
- [ ] Update langfiles


sonic2kk:
Tested fresh install of both tools with HoloCure, works as expected.

There was some strange behaviour observed where the SpecialK DLL seemed to get copied over again for some reason. This may be independent of this PR and should be investigated separately, but if possible, fixed in this PR.

sonic2kk:
Will work on simplifying some of the duplicate check logic where I can, think I overcomplicated it a bit on the ReShade side in `installRSdll`.

sonic2kk:
Test cases (2 consecutive runs per test):
- [x] Installing SpecialK Standalone
- [x] Installing ReShade Standalone
- [x] Installing SpecialK, then trying to install ReShade with no conflict (PLUGIN DISABLED)
- [x] Installing ReShade, then trying to install SpecialK with no condlict (PLUGIN DISABLED)
- [x] Installing SpecialK, then trying to install ReShade with a DLL name conflict (PLUGIN DISABLED)
- [x] Installing ReShade, then trying to install SpecialK with a DLL name conflict (PLUGIN DISABLED)
- [x] Installing SpecialK and ReShade together in one go with no conflict (PLUGIN DISABLED)
- [x] Installing SpecialK and ReShade together in one go with a DLL conflict, should prefer SpecialK over ReShade since ReShade calls are made before SpecialK calls, and we check if the DLL names conflict (PLUGIN DISABLED)
- [x] Installing SpecialK, then installing ReShade (PLUGIN ENABLED)
- [x] Installing SpecialK and ReShade together in one go (PLUGIN ENABLED)

sonic2kk:
As an aside, we should improve the logging+notifier message for ReShade+SpecialK DLL name conflict

sonic2kk:
This is mostly working now and can be merged soon

sonic2kk:
In testing, I had trouble getting ReShade+SpecialK to move into the game files, turns out I had the logic wrong for the `overrideReShadeVersion`, we were incorrectly checking if `USERESHSPEKPLUGIN -eq 0` instead of `1`, which meant the override incorrectly applied to, and only to, ReShade loaded independently from SpecialK. Fixed this.

There are a couple of issues left that I'm investigating with ReShade and SpecialK sometimes complaining about a conflict when there isn't one.

sonic2kk:
~~The case of installing ReShade+SpecialK independently but with a DLL name conflict doesn't quite work. It only installs one tool, but it should prefer ReShade over SpecialK. Instead it does the opposite, it prefers SpecialK over ReShade.~~

No it shouldn't, I'm a dumbass. This is working as expected, I just made a mistake.

sonic2kk:
Okay, fixed the logic so that ReShade is actually preferred. Will push.

sonic2kk:
We should also note on the wiki that ReShade takes priority over SpecialK when there is a DLL conflict if both are installed fresh at the same time, because ReShade is installed early.

sonic2kk:
Hmm, when using ReShade+SpecialK as a plugin (Default behaviour), it's incorrectly copying the ReShade DLLs as though they were independent. Will need to fix this.

sonic2kk:
Fixed, minor logic mistake in `installReshade`.

sonic2kk:
At long last, everything in this PR is working as expected. I'll do some minor code review to refresh myself on some of the messier sections and see if any improvements should be made before merging.

The logic for removing ReShade installs on toggling the ReShade+SpecialK plugin option will come in a separate PR, which will generally manage installing ReShade on top of SpecialK, and better install management. I may also be able to include some logic in that PR to remove renamed DLLs, by checking the values of `SpecialK_enabled.txt` / `ReShade.txt` and comparing it with our existing values, and seeing if the DLLs for an existing install were renamed, then we can remove those DLLs and let them get reinstalled.

sonic2kk:
Need to bump langfiles

sonic2kk:
Langfiles updated, version bumped, shellcheck is good, this is ready to merge.

PR #920, opened 2023-09-26
ReShade + SpecialK: Automatically export DLL names

sonic2kk:
One more piece of work for #894.

This PR adds the SpecialK and ReShade DLL names to the `WINEDLLOVERRIDES`, without any `.dll` extension if it exists. This means custom DLL name should be loaded by the game without any DLL overrides having to be set by the user. Unsure why this wasn't done in #881, but because of that we didn't do it in #912. I guess I thought it would be more complex than this :sweat_smile: 

The logic here is a bit messy, as it will export DLL names twice in some scenarios. If we leave the DLL names as defaults, for example `dxgi.dll`, then it will technically be in `WINEDLLOVERRIDES` twice. This shouldn't cause issues, though. Logging just might look a little odd at first :-)

The DLLs will only be exported for SpecialK if it actually reaches the block for installation, so they won't be exported in any other case. They were not exported before afaik so I guess the game just picked it up manually, hopefully this won't break anything :-)

<hr>

This is mostly untested, so will need more testing to ensure this doesn't break anything.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20230926-1"
+PROGVERS="v14.0.20230927-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Seems to work! Tested:
- Installing SpecialK standalone
- Installing ReShade standalone
- Installing SpecialK, then installing ReShade as a plugin
- Installing SpecialK and ReShade as a plugin together in one go
- Installing SpeicalK and ReShade independently

In all cases the DLLs were picked up correctly! This is good to merge then.

PR #926, opened 2023-10-02
Add Non-Steam Game: Add Option to Find Artwork in EXE Directory

### LINKED ISSUE #906: SteamGridDB support for Non-Steam games
## System Information
- SteamTinkerLaunch version: 6f5f68aa98084838640df59c66ad287623bca6a6
- Distribution: Arch Linux

## Feature Description

I'm sure you've got a **ton** of stuff on your plate at this point :laughing: , but after glancing at #738 and the recent commits, I was wondering if it'd be possible to have SteamGridDB grab (and assign) the game art for a specific game.

My current workflow for adding Non-Steam games is:

1. Install through Lutris and and use the built-in setting to add it to Steam
2. Use steamgrid to add the artwork to Steam
3. Rename the game to its Steam ID (where possible) to take advantage of Steam Input

Probably doesn't seem like a lot, but when you've got a bunch of Non-Steam games, it gets to be tedious fairly quickly. So...I was wondering if Steam Tinker Launch could maybe simplify the process somehow :thinking: 

Thanks!

sonic2kk:
Part of #906.

## Overview
This PR adds a checkbox to the Add Non-Steam Game GUI which will look for artwork in the same folder as the game exe directory if it is named appropriately. If the user enables this option and DOES NOT select/pass any artwork (such as if they pass no artwork or if they only pass a banner image), we will search for artwork with a relevant name in the same folder as the game EXE.

image

This can be useful if you don't want to remember/enter long paths, or if your game already has `icon.png` or something along those lines in the game files, which is reasonably common. It's worth nothing as well that this option **will** respect the copy method option.

We will look for the following:
- Icon - Image files named `icon`
- Banner - Image files named `hero`
- Logo - Image files named `logo`
- Boxart - Image files named `boxart`
- Tenfoot - Image files named `tenfoot`

Any file format can be used as long as `files` sees that it is image data, however not all file formats may be recognised by Steam.

This feature is also available as a commandline flag with `--auto-artwork`.

## Future Work
This is part of #906, so in future, this option can be used alongside an option to use SteamGridDB artwork, where there can be a fallback for any artwork that is not found (or if the call fails).

<hr>

TODO:
- [x] More testing
- [x] Update langfiles
- [x] Update wiki


sonic2kk:
Testing looks good. Wiki will be updated on merge.

sonic2kk:
Shit, forgot to update the helpscreen. Will do that when implementing the next part of #906.

PR #927, opened 2023-10-06
Add Non-Steam Game: Add Artwork Fetching by SteamGridDB Game ID

sonic2kk:
Implements #906.

## Overview
This PR adds a textbox to the Add Non-Steam Game GUI, as well as a commandline option (`-sgid`), which accepts the a SteamGridDB Game ID. This is then used to hit a SteamGridDB endpoint to fetch the most popular boxart (specifically boxart, more on that in a second).

image

We facilitiate this by making a couple of tweaks to, but mostly reusing, the existing SteamGridDB support in SteamTinkerLaunch, mainly the `getGrid` function. This PR modifies this function to accept a custom endpoint (since by default it hits one that searches for grids by Steam AppID, not SteamGridDB Game ID), as well as an option to enforce copying artwork over to the Steam grids folder, and finally allowing the function to accept custom AppIDs - since previously it always got the AppID from the grid filename (which doesn't work here because we specifically need the grid filename to have the Non-Steam AppID that we generate, so it's picked up by Steam.

This functionality will respect the SteamGridDB options set in the Global Menu, so users can specify custom dimensions, image types, etc.

## Limitations
We re-use the existing SteamGridDB functionality, there are a couple of limitations purely because the existing implementation has limits.

### Only Boxart is Fetched
Right now , only the Steam Boxart is fetched (files suffixed with `p`, i.e. `123123p.jpg`). All other artwork is ignored, and this is likely because SteamGridDB only fetches boxarts at this endpoint, and there is a separate endpoint for each artwork type:
- `https://www.steamgriddb.com/api/v2/heroes/game/{gameId}`
- `https://www.steamgriddb.com/api/v2/logos/game/{gameId}`
- `https://www.steamgriddb.com/api/v2/icons/game/{gameId}`
- The existing grids endpoint covers boxart

We can change this in the future in a refactor, but it's not in scope for this PR, which is solely about bringing Non-Steam Game Grid Selection up to par with the option on the Game Menu.

### Need to Generate SteamGridDB API Key
We need to generate an API key to use this function, and this is noted in the tooltip, but we need to make this clearer to the user I think.

### Refactor SteamGridDB integration
We should rework `getGrids` or perhaps make multiple functions with more generic logic to fetch other artwork, i.e. `getHero`, etc. We should probably make an overarching function and do some refactoring, and structure it like this.
```bash
function getSteamGridDBArtwork {
    function getSteamGridDBGrids {
        : # ...
    }

    function getSteamGridDBHeroes {
        : # ...
    }

    function getSteamGridDBLogos {
        : # ...
    }

    function getSteamGridDBIcons {
        : # ...
    }

    : # ...
}
```

This would allow all Steam games to have custom artwork fetched; regular Steam and Non-Steam Games. 

### Integrate Custom AppID Passing to SteamGridDB Command Line Option
We have a commandline option to fetch grids from SteamGridDB, we should extend it to support Non-Steam Games. This should also open a GUI with options similar to the ones on the Global Menu.

Doing this eliminates the need to integrate this with `setGameArt`, which is specifically for selecting artwork files.

### Automatically Select a Tenfoot if not Specified
Note that a "tenfoot" artwork (the artwork used when the game is the most recently played in the little box) is absent, perhaps we can add some logic to re-use the the hero or boxart artwork (in that order) ***IF*** they're present.

Maybe this should be a checkbox option, a general one under SteamGridDB settings.

<hr>

TODO:
- [x] Check/fix order of priority, ensuring we have expected behaviour for selected and fallback artwork when used in conjunction with this feature
- [x] Code cleanup
- [x] Update help screen (including missing options from #926)
- [x] Update text on Add Non-Steam Game GUI to add information about artwork selection, including some information on this new feature and that an API key is required, and that SteamGridDB configuration is applied here and can be changed on the Global Menu
- [x] Fix invalid (outdated?) text for how to generate a SteamGridDB API (should be done on profile section)
- [ ] Update wiki
- [x] Update langfiles


sonic2kk:
[Wiki has been updated](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Add-Non-Steam-Game).

sonic2kk:
Doing a final shellcheck on this, but everything else here has been implemented. This should be ready to merge momentarily

PR #932, opened 2023-10-08
Special K: Add toggle for installing d3dcompiler_47

### LINKED ISSUE #894: ReShade compatibility with Special K

sonic2kk:
Final piece of #894, though quite minor and seems to only affect me.

## Overview
This PR implements a toggle for installing `d3dcompiler_47.dll` for Special K. This DLL can cause crashes on some systems if used, particularly if used with ReShade, so this toggle allows a user to choose whether they want this DLL and they can turn it off if it causes a game crash.

image

The goal here is to provide an optional way to use `d3dcompiler_47` to prevent this crash. In testing it seems that Valve Proton requires this DLL, but GE-Proton and Proton-tkg will crash with it, possibly due to differences in patches and so on. Users can use this option in combination with a Proton version change, so the hope is this feature provides even more flexibility for tinkering.

To make clear: **This only applies for Special K!** ReShade is not relevant here, as I have not seen a case where the DLL causes a crash when present.

## Implementation
The existing `installd3d47` code will skip installing the DLL if the file exists, unconditionally. Even if SteamTinkerLaunch didn't install it (i.e. it's not tracked in `SpecialK_enabled.txt`). So if this DLL already exists in the game files, such as with the XCOM 2 launcher, we would not install it.

If the DLL already exists in the game files, we will use that, otherwise we'll install it ourselves. When the checkbox is disabled, we will remove the DLL only if we were the ones that installed it (if it's in `SpecialK_enabled.txt`).

<hr>

TODO:
- [x] More testing (SpecialK on/off, SpecialK+ReShade combinations, etc)
- [x] Update SpecialK wiki page with information about this option
- [x] Update langfiles


sonic2kk:
Fixed a few edge case issues identified in testing. Will do some testing with ReShade+SpecialK, and if that works, I will bump the langfiles and merge.

sonic2kk:
Interesting, the crash I was having no longer occurs with the newer Mesa version and newer GE-Proton versions, but it could re-occur I suppose. I think this option is still useful to have, though, just in case.

sonic2kk:
Looks like SpecialK DLLs were not exported properly in some cases, fixed that.

sonic2kk:
Tested with SpecialK and it works, it also works with SpecialK+ReShade.

Pending a langfile update, this should be good to merge.

sonic2kk:
Done!

sonic2kk:
[SpecialK wiki page](https://github.com/sonic2kk/steamtinkerlaunch/wiki/SpecialK) was updated as well.

PR #934, opened 2023-10-10
SteamGridDB: Overhaul SteamGridDB Integration

### LINKED ISSUE #933: [Feature Request] Overhaul SteamGridDB Integration
## System Information
- Distribution: Arch Linux

## Feature Description

*(expansion of #906)*

### Current Behaviour

### Proposed Enhancements

<details>

image

    - Logo
    - Hero
    - Banner
    - (Non-Steam Game Only) Icon
      - None (default)
      - Hero
      - Boxart
      - Logo
      - Default: Unchecked
      - Split text on `x` in `0x0`

</details>

<hr>

### Part 1

### Part 2

sonic2kk:
Will implement #933 or most of it, once completed. This is NOT ready for testing yet.

Massively WIP, but checklist for now:
- [x] Generic function to download single SteamGridDB artwork
- [x] Standalone function which can download Hero/Logo/Boxart based on given AppID/Game ID, filename (default to AppID for Steam games) and other optional parameters, and can optionally apply it for Non-Steam Games
- [x] Function to download **STEAM GAME** (not nonsteam) artwork based on a list of **STEAM APPIDS** (calls the above commandline function in a loop)
- [x] Migrate Owned/installed grid download functions to use this loop method mentioned above
- [x] Wire up/overhaul SteamGridDB commandline usage to allow using this command
- [x] Overhaul Global Menu SteamGridDB options to allow specifying styles/dimensions/etc for all new artwork
- [x] Investigate using notifier
- [x] Fully replace all instances of `getGrids` with the new functions
- [x] Fetch icon when adding Non-Steam Games
- [ ] ~~Add option to fetch grids for all Non-Steam Games only (could be handy, especially when combined with the replace/skip options etc)~~ (will go in a separate PR down the road)
- ~~[ ] Add optional field on Game Menu for downloading SteamGridDB artwork~~ Superseded by commandline option

Right now, this works for Non-Steam Games, so that's a win! However it doesn't respect any of the SteamGridDB settings, because we need to overhaul these settings to accept dimensions etc for each type of grid.

We will also need to massively overhaul the wiki to document the changes here.


sonic2kk:
Added commandline usage. The SteamGridDB wiki page will need to be updated with detailed usage instructions. It should have examples of searching both for Steam games and Non-Steam games, instructions on how to get the AppID for Non-Steam games, instructions for how to apply the artwork manually from Steam (in case the user just wants to download it but wants to set it themselves in Steam/elsewhere), and more constellations for usage.

<hr>

### Usage Examples
*(Note that there are more options than this, these are just high level examples for common usage)*

Get artwork for game in Steam library using Steam AppID
```bash
# Download artwork for "The Elder Scrolls IV: Oblivion Game of the Year Edition" using its Steam AppID
# Note that the filename AppID is not specified. We use Search ID to name the file automatically as it is the Steam AppID
steamtinkerlaunch sgdb --search-id="22330" --steam --apply --skip-existing
```

Get artwork for game in Steam library using SteamGridDB Game ID:
```bash
# Download artwork for "The Elder Scrolls IV: Oblivion Game of the Year Edition" using its SteamGridDB Game ID
# Searches using the SteamGridDB Game ID but since the game is owned on Steam, name it after the actual Steam AppID
steamtinkerlaunch sgdb --search-id="5258102" --filename-appid="22330" --nonsteam --apply --skip-existing
```

Get artwork for Non-Steam Game that is available on Steam using Steam AppID
```bash
# Download artwork for "The Elder Scrolls IV: Oblivion Game of the Year Edition" using its Steam AppID despite it being added as a shortcut
# Assumes this is a Non-Steam Game (shortcut) but we're searching on the Steam AppID, yet naming the file after the Non-Steam AppID in the library
steamtinkerlaunch sgdb --search-id="22330" --filename-appid="12123123" --steam --apply --skip-existing
```

Get artwork for Non-Steam Game using SteamGridDB Game ID
```bash
# Download artwork for "The Elder Scrolls IV: Oblivion Game of the Year Edition", added as a Steam shortcut, using the SteamGridDB Game ID
# We search using SteamGridDB's Game ID and then name the downloaded file after the Steam shortcut ID
steamtinkerlaunch sgdb --search-id="5258102" --filename-appid="12123123" --nonsteam --apply --skip-existing
```

<hr>

We will also need to update other wiki pages, such as the Add Non-Steam Game wiki page, to note that as much artwork as is found will be downloaded, and that it will use the settings on the Global Menu to do so.

Of course before we do that this PR will need to be ready for merging, and we're not there yet, as we don't have the GUI options.

sonic2kk:
I may revisit the commandline usage either before merging this, or sometime after, as I have a feeling some improvements could be made... I'll let it simmer for a bit and see how it fares in my own usage, and how everything fits together once I actually get the GUI options set up so that dimensions and things can all work as expected. Once I can do a "full test" maybe I'll be more satisfied.

I'm happy with how everything is architected, I like how I was able to break things out here and re-use `getGrids` as a standalone function to get grids from SteamGridDB based on various parameters (and how it can use a specific endpoint based on Steam AppID/Game ID). But I dunno, the commandline usage almost doesn't feel as good as it could to me. But again, once I've had time to use it, maybe I'll feel differently.

sonic2kk:
The next steps for this PR is to overhaul the SteamGridDB Global Menu options, so that we can specify settings for the other artwork (dimensions is the primary one, but also categories, etc).

We should also consider how to handle tenfoot artwork. I'm apprehensive about setting the hero artwork. Perhaps we could simply add a setting for the tenfoot with different dimensions? I'll have to see how they're categorised on SteamGridDB and how applicable this is, but we could simply categories the tenfoot separately yet internally have it reflect hero/boxart, but we give the endpoint the dimensions of the tenfoot.

After this, we'll need to update all usages of `getGrids` with probably a direct call to `downloadArtFromSteamGridDB` since the remaining usages are for internal SteamTinkerLaunch usage.

A nice to have feature would be fetching and setting icons when adding Non-Steam Games. Currently we can't easily update these, as we don't have a good way of parsing and updating the `icon` field of the binary `shortcuts.vdf` file, and to my knowledge there's no other way to set icons for Steam shortcuts. So right now the only possibility is to download the file and set this when we're adding the shortcut. In future, we may be able to update the icon path in `shortcuts.vdf`, but not right now. We don't have a parser for it right now, so we can't search and update the fields :sweat_smile:  

Finally, after this, we'll need to do extensive testing. To protect my library on my PC, I may just test on my laptop (sine I'll need to dig it out for testing other PRs) and test the `installed`/`owned` options with a small subset of my library. I'll also test to ensure a "fresh" SteamTinkerLaunch install works too. Then finally, once we confirm all this to work and we confirm the commandline usage to set SteamGridDB works, we'll need to make sure there is no regressions anywhere. We need to make sure that, game artwork etc is not incorrectly updated (such as if/when SteamTinkerLaunch fetches grids for new games, we need to **make sure it doesn't set this artwork** for the game, only for STL itself).

sonic2kk:
SteamGridDB does have a preset horizontal grid option that we could try to search on for tenfoot.

image

I don't think we can explicitly pass this to the endpoint, though, so for tenfoot we'd have to just hit the grid endpoint with dimensions specified. So boxart and tenfoot would use the same endpoint, but with different dimensions. The dimensions on the UI should reflect the recommended Valve one, as well as the preset ones for the Steam GUI.

We can't implement this on the commandline until we add the GUI options I think, unless we were to hardcode it.

<hr>

In testing, these grids appear to look fine, if not a little small at times.

image
image
image
image

Big Picture Mode also looks fine, though sometimes the low resolution artwork is very noticeable at high resolutions.

image
image

There's not much we can do about artwork looking strange except to prioritise higher resolution artwork. And if a user knows what they're doing, they can specify custom dimensions.

<hr>

I think using the grid endpoint again with different dimensions parameters is the way to go here. For grids we should give resolution options that SteamGridDB lists for Steam first, then GOG Galaxy 2.0. Square grids don't really apply and would look really off I think. 

When artwork is not found, this should already be logged, but it would be helpful to echo this out as well so users on the commandline know this as well.

sonic2kk:
Especially given this new approach, we will need to consider migrating the cfg options, though we will probably *ignore* dimensions, since the existing dimensions options don't make sense to migrate.

<hr>

In other news, I temporarily hardcoded a call to SteamGridDB to fetch artwork with a size of `920x430,460x215`, and it's working great.

Some less popular and some older games may not have grids with this size, but that isn't really an STL problem. That's up to users to manage.

sonic2kk:
Seeing an error now, `find: ‘realpath’ terminated by signal 13`. This may be related to finding the banner to display on the settings menu, and may just need silenced.

sonic2kk:
The new SteamGridDB GUI options are now in place! Logo does not have dimensions, but aside from this, we can now expose all options that would get passed to the SteamGridDB endpoint!

image

This is untested for now, aside from spot-checking a couple of UI elements to make sure they save.

We will need to check the following:
- [x] Test with single resolutions (i.e. for tenfoot/hero) to ensure only the selected one is used
- [x] Test that disabling some elements truly disables them, i.e. disabled hero option will not download hero artwork
- [x] Test that the expected values are being passed to the SteamGridDB endpoint (shown in the logs), i.e. ensure tags and so on are actually passed as query parameters

sonic2kk:
The way we're saving some of the default options for types (animated/static) is wrong and needs fixed.

sonic2kk:
This is working from the commandline now, and it seems the options on the UI are working. Further testing is needed, but it seems like this should work now!

Tenfoot is working great, I found out one of the styles we can remove from our search is `no_logo`, which is very useful for tenfoot. If there is no logo, it may not be obvious what the game is. So we can, by default, ensure downloaded tenfoots have the logo.

sonic2kk:
Big remaining things left for this PR:
- ~~Fetch icon when adding Non-Steam Games~~
- ~~Replace `getGrids` calls with our new function `downloadArtFromSteamGridDB`~~
- ~~Investigate the realpath error~~
- LOTS of testing
- ~~Cleanup todos~~
- Cleanup code
- Update langfiles
- Overhaul wiki

sonic2kk:
We can now get icons for Non-Steam Games if they exist on SteamGridDB! Very happy with how that turned out.

sonic2kk:
When the remaining instances of `getGrids` is replaced with our new commandline function, we'll need to do some strict testing!

sonic2kk:
Got around to replacing the remaining `getGrids` calls with calls to `commandlineGetSteamGridDBArtwork`. For now, since these remaining calls are for all game launches, it is untested, just like the installed/owned game grids downloading.

Most of what remains now is further testing.

<hr>

Upon further thought, I think having an option on the Game Menu to specify the SteamGridDB Game ID is useful, because if a user is downloading grids they may want to, for example, download artwork for "Sonic Colors" instead of "Sonic Colors: Ultimate", because the former has substantially more artwork (and better artwork imo) than the latter. That could also go in a separate PR though, as this one is getting pretty big.

For Non-Steam Games, we could maybe store this? Though I am not sure if this would be feasible, it would be nice to store this as a preset value. We'd have to create a config file for the game at add-time, though...

sonic2kk:
Marked the PR as ready now, because it is in a state where it could be tested (though should still be considered heavily in-flux).

Another change I think I will make is to the commandline usage, to add some more sensible defaults perhaps. I will also continue to use and test this feature myself because it's fun :smile: Once I get more testing in and ensure that this integration doesn't break or do things it shouldn't (such as incorrectly download artwork when it's not supposed to), this should be ready to merge.

Hopefully it will be merged this month :partying_face:

sonic2kk:
There is a SteamGridDB endpoint we can hit to try and search by Game Name, but I would need to investigate how this works. Endpoint is: https://www.steamgriddb.com/api/v2/search/autocomplete/<name>

This is the same endpoint probably that the website hits to return game info, so we'd need to be careful that we manage the response properly. This would yet again be brand new functionality, so it would be implemented separately. Still, it would be really cool to have this especially for Non-Steam Games. It means we wouldn't have to rely on the SteamGridDB Game ID alone.

It would also likewise be nice if we could optionally pass the Steam AppID when adding a Non-Steam Game, instead of just the SteamGridDB Game ID.

The functionality in this PR would be a prerequisite to any work related to both of these, but it is something we can consider for future work.

We could make searching SteamGridDB for the Game ID a commandline function as well, as it could be generally useful.

sonic2kk:
Made a few more changes, still testing:
- Made `steam` the default when setting game art now, now for Non-Steam you'll need to pass `--non-steam`.
- Changed how game artwork is set, so that grid artwork is the fallback and will not overwrite regular game artwork
    - This means if we can't get artwork from Steam, we'll fall back to resized hero artwork saved by the user
    - When using this artwork, we convert and save as jpg instead of png now, which should fix some checks
- Updated helpscreen

Not sure if I'll make any other changes to usage. I considered making `--apply` the default, but this will be the default if the user enables the "Download grids to Steam" option on the SteamGridDB options. I'll still keep usage in mind in case I think of anything that should be changed, such as adding shorthands for some of the ID flags. With usage I'm also keeping automated scripts in mind, and in future, I may add the option to simply search on Game Name, but that will first go to Non-Steam Games as a test.

I also investigated the `realpath` error and couldn't get it to happen again. I suspect it was happening in `setShowPic` though and was fixed after the above changes.

I have also cleaned up a lot of the remaining TODOs, the rest (max file size option etc) can go in different PRs in the future as enhancements, as they're very much just nice to have and not crucial (the animated/static filter should handle most of that).

I have done a good amount of testing with this already, replacing artwork, adding Non-Steam Games, and testing games never launched with SteamTinkerLaunch before. It seems to be working as expected, and won't default in any way to using SteamGridDB without the user explicitly using it.

<hr>

So remaining is just more testing, and then the wiki overhaul. But this should be ready very soon!

sonic2kk:
I think this is working well enough that it's ready to merge once the wiki update is ready.

There are still areas for improvement, such as finalizing commandline usage syntax/wording, and searching SteamGridDB for the Game ID using the Game Name, but overall it's in a good enough state for a merge.

sonic2kk:
Wrote up a big wiki page overhaul, once this PR is merged the wiki can be updated.

sonic2kk:
Added last minute change for `--no-apply` option.

sonic2kk:
Langfiles updated, shellcheck is good, version bumped, code looks good as well and this seems to work great in my testing. This is ready to merge!

PR #935, opened 2023-10-14
ModOrganizer 2: Add option to Select Custom Installer Executable

sonic2kk:
## Overview
This PR adds an option on the Global Menu under ModOrganizer 2 options, to specify a custom installer executable. This is useful if there is a custom/unofficial build that a user wants to install, such as a build that was rumoured to be floating around on a Discord server with Starfield support. A user could also build MO2 from source and this option would allow them to install it.

image

This option is gated behind a checkbox on the Global Menu, so that it is not accidentally enabled. Also, if the file does not exist, we will fall back to downloading MO2 from GitHub.

<hr>

Right now, this change is **entirely untested**, as I am using MO2 on my PC right now, and I don't mod on my Steam Deck. I'll need to dig out my laptop+charge/update etc, so it may be a while before I can test.

It would be great if someone could test this and see if it works!

## Future Work
I'm hoping in future to apply a very similar change to Vortex. I had originally wanted to do this for Vortex (#810), but went with MO2 first because I actually like and use MO2, and it was a slightly more straightforward change compared to Vortex (as with Vortex we would have to work into the logic to ignore the version/channel etc).

<hr>

TODO:
- [x] Update commandline usage to allow passing this option to `install` command (will be a delicate operation no doubt
- [x] Do some testing (ensure existing functionality still works without interruption+make sure this option actually works too)
- [x] Update helpscreen with custom executable usage
- [x] Update wiki to document this new feature
- [x] Update langfiles


sonic2kk:
Commandline usage added.

sonic2kk:
Helpscreen updated.

sonic2kk:
Langfiles updated.

sonic2kk:
The main remaining work on this PR now is testing, and then writing up the wiki updates. Once this is confirmed to be working, I'll update the wiki.

It may be a while before I can test, so I'd appreciate any community testing.

sonic2kk:
Should also more on the wiki that this option could be particularly useful if using two ModOrganizer compatdatas and that you could swap between versions this way.

sonic2kk:
May have to change the UI element type, think I got it wrong...

sonic2kk:
Fixed exe file picker element type and resolved conflicts.

sonic2kk:
Hoping to test and merge this over the weekend.

sonic2kk:
Okay, did some fixups and testing here. It works as expected. Tested this PR with using ModOrganizer 2.5.0 and ModOrganizer 2.4.2 (2.5.0 does not install but I used it as a test executable that was not 2.4.4).

- ModOrganizer 2 download and install works correctly when `USEMO2CUSTOMINSTALLER -eq 0`
- Same as above when `USEMO2CUSTOMINSTALLER -eq 0` and a custom MO2 EXE is selected on the global menu, so the custom EXE is correctly ignored if we don't enable the option to use it.
- ModOrganizer 2 custom executable from Global Menu is used if `USEMO2CUSTOMINSTALLER -eq 1` and we have a valid game EXE
- ModOrganizer 2 custom executable from commandline is correctly used if we pass it and if it is a valid file (`USEMO2CUSTOMINSTALLER` is forced on in this case for that run only)
- If ModOrganizer 2 custom executable from Global Menu or Commandline is invalid file we correctly ignore it
- ModOrganizer 2 custom executable from commandline correctly takes precedence over one on Global Menu
- ModOrganizer 2 custom executable commandline usage works even if nothing is set on Global Menu
- When none of these options are set (all of the work in this PR is **OFF** by default) there is no impact, so there is no impact to existing functionality.

I'll do a quick ShellCheck on this and make any required changes, then update the wiki and merge this. Finally, this PR is out of development hell!

sonic2kk:
No changes required for ShellCheck. Will do a final glance over the code, bump the version and merge once the wiki is updated.

sonic2kk:
Code looks good, ShellCheck is green, wiki is updated ([Commandline Usage](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Mod-Organizer-2#command-line) updated, [Global Menu configuration](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Mod-Organizer-2#global-menu) section added), version is bumped. This is ready to merge!

PR #938, opened 2023-10-17
SteamGridDB: Add --search-name option to commandlineGetSteamGridDBArtwork

### LINKED ISSUE #933: [Feature Request] Overhaul SteamGridDB Integration

sonic2kk:
Another piece for #933.

This PR adds a `--search-name` flag to the `commandlineGetSteamGridDBArtwork` function. This uses the standalone function to search SteamGridDB for a Game ID using the Game Name, added in #937.

When `--search-name` is passed, since we'll be using the SteamGridDB Game ID, the user must provide the AppID for the game to use in the filename with `--filename-appid`. Otherwise we won't be able to save the file with the correct AppID.

Also when this flag is passed, we force use the SteamGridDB `game` endpoint, so the user doesn't have to pass `--nonsteam` (it's fine if they do, it will simply be ignored).

Tested this with "Doki Doki Literature Club" and it worked no problem.

TODO:
- [x] Update helpscreen
- [ ] Update wiki


sonic2kk:
[SteamGridDB wiki page](https://github.com/sonic2kk/steamtinkerlaunch/wiki/SteamGridDB#single-game) has been updated.

PR #946, opened 2023-10-21
Non-Steam: Disable Steam AppID file

sonic2kk:
Will resolve most of #941.

This PR adds a function to check if we have a Non-Steam Game, and uses this in `checkSteamAppIDFile`. If we have a Non-Steam Game, we won't run the code to generate the `steam_appid.txt` file and we will turn the option off in the config file. This is useful for users who have this option on by default, as for Non-Steam Games only this will be ignored, so it is safe for them to keep it on by default.

Steam uses `steam_appid.txt` if it is in the same folder as the game exe to set its SteamAppId variable for Non-Steam Games, and when this is set via this file for Non-Steam Games, we get a crash if this value is defined and not `0`. This check is done before STL even has a chance to run, so we can't fix it from our side. Instead, we can only prevent the code that creates this file from running when we have a Non-Steam Game. Any user who has this file in their game dir currently will have to manually remove it.

In future, this function could be used to set `ISGAME=4`, but then we'd need to determine if the game is Windows or not, so probably `ISGAME=4` and `ISGAME=5`. For now, though, this PR is sufficient.

<hr>

TODO:
- [x] Update langfiles


sonic2kk:
Langfiles updates, version bumped, some other minor todo cleanup I did in this PR for giggles, and shellcheck is green. Ready to merge!

PR #948, opened 2023-10-21
Non-Steam: Use GRIDBANNER for initial fallback instead of CUSTPIC

sonic2kk:
CUSTPIC always expects PNG, GRIDBANNER uses find and can take any filetype.

Fixes STL not replacing 0byte showpics if the hero grid was a jpg instead of a png.


sonic2kk:
Testing looks good, ShellCheck is green, this should be ready to merge!

Issue #947, opened 2023-10-21
x64dbg debug not working


abc123098123:
## System Information
- SteamTinkerLaunch version: v14.0.20231021-2
- Distribution: Arch Linux
- Installation Method: Package Manager

## Issue Description
I select the option to `Debug with x64dbg` and put the game in a steam collection called `x64debug` but the game always starts without x64dbg.

From the logs, that you can find below, it seems `steamtinkerlaunch` tries to download x64dbg without success and because of that, x64dbg archive extraction does not work.

I also tried to install x64dbg in the expected directory. After that, it does recognize that x64dbg exists and tries to run it, giving the following error, extracted from other logs:

```
Sat Oct 21 03:40:09 PM WEST 2023 INFO - extWine64Run - Command in Proton WINEPREFIX is: WINE="" WINEARCH=win64 WINEDEBUG="-all" WINEPREFIX="/run/media/romeu/GAMES/SteamLibrary/steamapps/compatdata/230050/pfx"  /home/romeu/.config/steamtinkerlaunch/downloads/x64dbg/release/x32/x32dbg.exe /usr/bin/steamtinkerlaunch: line 6872: : command not found
```

It says `command not found`, but I don't know which command. Steamtinkerlaunch is inside `/usr/bin/steamtinkerlaunch`, and `x32dbg` executable file is inside the expected directory

## Logs

```
Sat Oct 21 03:33:06 PM WEST 2023 INFO - setflatpak - started SteamTinkerLaunch from /usr/bin/steamtinkerlaunch
Sat Oct 21 03:33:06 PM WEST 2023 INFO - initAID - Set AID to PLACEHOLDERAID '31337'
Sat Oct 21 03:33:06 PM WEST 2023 INFO - main - Current SteamTinkerLaunch working directory is '/home/romeu'
Sat Oct 21 03:33:06 PM WEST 2023 INFO - loadLanguage - First load the default language 'english' to make sure all variables are filled
Sat Oct 21 03:33:06 PM WEST 2023 INFO - loadLangFile - Language from command line is 'english'
Sat Oct 21 03:33:06 PM WEST 2023 INFO - loadLangFile - SYSTEMSTLCFGDIR is '/usr/share/steamtinkerlaunch'
Sat Oct 21 03:33:06 PM WEST 2023 INFO - loadLangFile - Command line language 'english' is no file - trying to find its absolute path
[...]

PR #592, opened 2022-09-20
Fix logic for fetching MO2 and Vortex version

sonic2kk:
Fixes #591, #589. Will be needed for #540.

## Problem 
GitHub changed how assets are loaded on the releases page, they are loaded after the page with lazy loading, so we can't `wget` on the releases page and get the executable when trying to get the executable version and name for MO2 and Vortex Mod Manager. This means MO2 and Vortex can't be downloaded, as STL doesn't catch their version. `MO2SETUP` and `VORTEXSETUP`, which hold the setup executable, end up blank since the `wget` code fails.

## Solution
Instead, we can get the latest tag version from the release tags page and get the release executable name from the `expanded_assets` page for that version. On this page, we can use the previous logic to get the setup executable name.

This solution uses a generic function that works for both MO2 and Vortex. This function only changes how we fetch the latest version for these tools, the actual download logic is the same since that URL did not change.

## How This Works
(adding this in an edit because I forgot to include it initially in my rush to get this up, oops!)
Basically, the way we get the version now is as follows. See my [comment with an initial attempt](https://github.com/frostworx/steamtinkerlaunch/issues/591#issuecomment-1252724630) at getting this working and [my initial generic function attempt](https://github.com/frostworx/steamtinkerlaunch/issues/591#issuecomment-1252758125) for some background/context on how this solution came to be.

We have a new function called `getLatestGitHubExeVer`. This function takes two inputs: The exe name (`SETUPNAME`) to look for, and the project URL (`PROJURL`). This is why the two new project URLs were added, they were added to be passed to this function.

We check the tags page (e.g., for MO2 this is https://github.com/ModOrganizer2/modorganizer/tags) and grep all of the elements that have a specific tag link. For example with MO2 this would be getting all elements on the page that have `/ModOrganizer2/modorganizer/releases/tag`. We take the first one which should always be the latest **release** (we are only looking for tags associated with a **release**, and in the case these two projects, this seems to work as we don't get any MO2 development tags). 

There were probably other ways to check this but the tags page seemed to return the least amount of data, so I chose that method.

[...]

PR #950, opened 2023-10-21
x64dbg: Add toggle to attach game process on startup

sonic2kk:
Addresses part of #947.

## Overview
This PR adds an option to checkbox toggle to control a new variable `X64DBG_ATTACHONSTARTUP`. This controls whether the game EXE is attached to `x64dbg` on startup, and it is enabled by default.

image

### Default Attach Behaviour
It's on by default because this should have been the default behaviour, but it was not due to a bug. `winepath -w` was returning a lot of spew before the game path, which meant `x64dbg` could not actually read the standalone game path. This has been fixed by getting `tail -n1` of the `winepath -w` output, and now `x64dbg` will start with the game process as it should have (and as it may have done previously).

There was another issue with this default process flow, where when x64dbg was closed, the game would start anyway. This has been fixed by doing a check on this new `X64DBG_ATTACHONSTARTUP` control variable before we launch the game in `launchSteamGame`. If `USE_X64DBG` and `X64DBG_ATTACHONSTARTUP` are both enabled, then we can assume x64dbg will manage the game launch and that we don't need to do it.

### New No Attach Feature
Finally, a new flow has been introduced based on discussion in #947, which was the main reason for adding this checkbox. In some cases a user may want to disable this, for example if they're having trouble attaching the right process by default, or if they want to only attach to a certain game process. So we allow the user to disable attaching the game process by default, and allow the game to launch normally, and x64dbg to launch independently. We do this by sleeping, then running x64dbg, and continuing on to launch the game process. We need to sleep otherwise x64dbg will open and for some reason prevent the game from opening.

The sleep value is a magic number, hardcoded to `5` for now, because `1` didn't work and I didn't encounter a case in testing where any value above `5` was required. It worked in all my tests with as low as `2` as well, but I decided to go with `5` to be as generous as possible without introducing too much of a wait. We could introduce an option to adjust this in future if needed, though.

Since on this path the game is launched by STL normally and has to be attached to x64dbg after-the-fact, we don't have to worry about it launching after x64dbg is closed. Likewise, when the game process is attached to x64dbg, closing the program will close the game process. So in every instance, the game processes and x64dbg should now open and close more cleanly.

## Future Work
GitHub changed how it renders the releases page a while ago, and this broke MO2 and Vortex. We fixed that (#592), however it seems we missed x64dbg. Since x64dbg doesn't use tags in the same way, we'll need to figure out another way to find and then download the correct release.

<hr>

This PR was tested by me and also tested by the reporter in #947, so it should be ready to merge after some quick code review and langfile updates. This should work for Steam games and Non-Steam Games. As far as I know, this is not applicable for Native Linux games.


sonic2kk:
Tested with this new option on and x64dbg off, everything works good. When x64dbg is not installed, `RUN_X64DBG` is set to `0`, so there are no worries of any problems when x64dbg is not installed.

Langfiles bumped, shellcheck green, this works in testing, so it should be ready to merge!

Issue #949, opened 2023-10-21
Non-Steam Game Categories Not Working


seanryanmcewan:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: SteamOS
- Installation Method: ProtonUp-Qt

## Issue Description
Hello! I'm using steamtinkerlaunch for it's ability to add non-steam games to the steam library via command line options.

I am uncertain whether the 'tags' option is currently supposed to be working. [This](https://github.com/sonic2kk/steamtinkerlaunch/issues/190) issue from several years ago mentioned that it wasn't working, but then the following comments were unclear as to whether that was resolved or not.

To be clear, I am running the script, closing Steam, then reopening Steam. 
SteamTinkerLaunch successfully added the game to the library, but it was not added to any Collections.
I also tried the linked solution of running `steam://resetcollections`. I rebuilt my collections and tried again, and it still doesn't seem to be working. 

Is this intended to be working? 
If it's not, perhaps looking at [Steam Rom Manager](https://github.com/SteamGridDB/steam-rom-manager) would be helpful - they are somehow successfully adding items to the Steam Library and adding them to collections.

Thanks for reading!

## Logs
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/13061113/steamtinkerlaunch.log)

sonic2kk:
Have you tried using the latest SteamTinkerLaunch (SteamTinkerLaunch-git from ProtonUp-Qt with Advanced Mode enabled)? You're using v12.12 which is very old. Also, Add Non-Steam Game got a lot of useful improvements lately. It also correctly sets the Steam AppID now, which *may* resolve the problem.

I'll take a look and see if tags are working, perhaps how we write them out is wrong or something. This worked before a long time ago, but I'll take a look when I have some time.

> If it's not, perhaps looking at [Steam Rom Manager](https://github.com/SteamGridDB/steam-rom-manager) would be helpful - they are somehow successfully adding items to the Steam Library and adding them to collections.

That program is written using custom libraries in TypeScript, SteamTinkerLaunch writes out directly to the binary VDF file and is written in Bash, so it is not super helpful unfortunately.

sonic2kk:
[...]

PR #951, opened 2023-10-21
Add Non-Steam Game: More closely match Steam casing and include missing fields

sonic2kk:
Small change to how we write out to `shortcuts.vdf`. This now includes some extra fields to match newer ones that Steam writes out, and also changes some casing to better match what Steam writes out.

Found while messing around to figure out the cause of #949, but doesn't fix that issue.


sonic2kk:
This didn't break anything in my tests and ShellCheck is good. A very small change, so now that the version is bumped, this is good to merge.

PR #956, opened 2023-10-23
Primerun: Ensure nvidia_icd.json is used

### LINKED ISSUE #953: primerun setting doesn't work
## System Information
- SteamTinkerLaunch version: 12.12
- Distribution: Manjaro
- Installation Method: AUR (yay)

## Issue Description
If I launch a game through tinkerlaunch after enabling the primerun option in tinkerlaunch's game settings, it will just use the iGPU instead.

## Logs
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/13067468/steamtinkerlaunch.log)

sonic2kk:
Should fix #953.

When `USEPRIMERUN` is on, we export several environment variables. But if the `nvidia_icd.json` is not set in the current environment's `VK_ICD_FILENAMES`, it will not be able to use the Nvidia GPU. For cases likee this, we should explicitly export `nvidia_icd.json`.

The fix for this was given by OP in #953 (they informed which environment variable to use), this PR just applies it in the appropriate spot in the code.

This path is not likely to change, and if other custom ones are necessary, a user can manually set these in their STL env variable config file.

Putting this up as a branch in case this fix is not sufficient, and we have to make other changes.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20231023-1"
+PROGVERS="v14.0.20231024-1 (primerun-change)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
OP confirmed it to work, shellcheck is green. Ready to merge.

PR #963, opened 2023-10-27
Non-Steam: SteamGridDB and getid/gettitle Commandline Support

### LINKED ISSUE #933: [Feature Request] Overhaul SteamGridDB Integration

### LINKED ISSUE #960: [Feature Request] Non-Steam Game Improvements
## System Information
- Distribution: Arch Linux

## Feature Description

For each of the commands:

```
Name: Dungeon Keeper 2
AppID: 3849204543
```

*(see also #934)*

<hr>

TODO:

sonic2kk:
Heavily WIP, currently does the following:
- Add a bunch of commands to facilitate parsing `shortcuts.vdf`
- Support Non-Steam Games with `getIDFromTitle`
- Support Non-Steam Games with `getTitleFromID`
- Add support for Non-Steam Games with command `steamtinkerlaunch update grid nonsteam` (alternatively, `update grid shortcut`)

Remaining work before I edit this PR description with something more informative and structured:
- Add support for setting icon for Non-Steam Games with the update grid command
- Shellcheck fixes

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20231106-1 (get-sgdb-for-all-nonsteam-aids)"
+PROGVERS="v14.0.20231107-1 (get-sgdb-for-all-nonsteam-aids)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
This PR adds many, many functions for interfacing with the `shortcuts.vdf` file, including fetching and editing specific entries. These functions will be generally useful and will also allow for, in this PR, updating the icon for Non-Steam Games.

The next step for this PR is to fetch the icon for Non-Steam Games.

sonic2kk:
Oh yeah, we'll also need to run shellcheck and do some fixups. I haven't checked yet but I suspect it'll be a real mess...

sonic2kk:
Untested implementation for setting the shortcut icon. There are now three mostly untested changes:
1. `getSteamGridDBNonSteamIcon` was broken out into a separate function. This works with `getGridsForNonSteamGames` but I didn't test if it works for `addNonSteamGame` yet. This needs tested
2. `getSteamGridDBNonSteamIcon` now returns the icon path, which we didn't check for before. We need to make sure that this icon path is valid at least for `addNonSteamGame`, and that it works as expected.
3. `editSteamShortcutEntry` was tested during debugging and updated a hardcoded icon path for a given shortcut, but has not been tested in `getSteamGridDBNonSteamIcon` yet.

I also did a quick run of ShellCheck. No glaring errors, just a couple of style, globbing, and unused/unnecessary variable changes that need addressed. Once the above testing is done and this is all working, it should be straightforward to address those and get this merged.

We're almost there!

sonic2kk:
Can't set the icon in `getSteamGridDBNonSteamIcon` in its current form just yet, because it sets the icon to all the downloading output. Damn! We'll figure out a way though, probably tomorrow.

It is updating the `shortcuts.vdf` file though, so it will work once we correctly pass it the path for the icon.

sonic2kk:
We should also add some more logging before merging this PR.

sonic2kk:
Got some rough changes locally that I'll push soon which resolve the remaining functionality issues, and allow icons to be downloaded and set when updating grids for Non-Steam Games! We correctly edit `shortcuts.vdf` and insert the value.

Some more testing needs to be done, and there is some potential to break out a couple of areas into separate functions, bbut functionally the PR is complete once I push my local changes. Major progress, really just testing and cleanup remains!

sonic2kk:
Tested adding Non-Steam Games, setting the icon appears to still work.

sonic2kk:
Shellcheck fiixes in place, this is probably ready for merging after a couple more tests...

sonic2kk:
Add Non-Steam Game still works, getid/title still works, editing shortcut fields works, I think this is ready to merge! Testing and shellcheck are green and version is bumped. No langfile changes here.

The way we handle which column to edit in `editSteamShortcutEntry` is perhaps a little inefficient in that it uses a case statement, but eh, I think it's fine. It's for internal use only anyway, in future if there's demand it would be easy to expose on the commandline though.

<hr>

I never did update the PR description but I think it's fine.

PR #965, opened 2023-10-30
Add Non-Steam Game: Refactor Shortcut Entry Creation

### LINKED ISSUE #960: [Feature Request] Non-Steam Game Improvements

sonic2kk:
Refactors how we generate shortcut entries to write out to `shortcuts.vdf`. This PR is in part motivated to simplify the codebase as well as reduce overall script size.

Instead of having a lot of if/else checks for whether to use `\0x01` or `\0x00`, we instead just printf regardless and if the value is blank, default to `0`. Then, to ensure the value is blank if it isn't 1 or 0 (since we can pass values on the commandline like `--hide="no please"`, anything that is passed to one of the following tags will be defaulted to blank, and so set as `0` in the `shortcuts.vdf`:
- `--hide`
- `--allowdesktopconf`
- `--allowoverlay`
- `--openvr`

I had originally wanted to change the usage to simply be `--hide`, and if the flag was present at all, set `NOSTHIDE=1`. But with how we use this function in `addNonSteamGameGUI`, that was going to require a lot of if/else checks, or some messy workarounds in that function when we call `addNonSteamGame`, so I didn't bother. In future if someone can figure out a clean way to do this, a PR would be weelcome!

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -23463,7 +23463,7 @@ function addNonSteamGame {
 				NOSTVR="$( checkValidVDFBoolean "${i#*=}" )"
 				shift ;;
 			-t=*|--tags=*)
-				NOSTTAGS="$( checkValidVDFBoolean "${i#*=}" )"
+				NOSTTAGS="${i#*=}"
 				shift ;;
 			-stllo=*|--stllaunchoption=*)
 				NOSTSTLLO="${i#*=}"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -23321,9 +23321,7 @@ function addNonSteamGame {
 	SCPATH="$STUIDPATH/config/$SCVDF"
 
 	function checkValidVDFBoolean {
-		if [ "$1" -eq 1 ] || [ "$1" -eq 0 ]; then
-			echo "$1"  # Only return value if 1 or 0, then other places can do "${var:-0}"
-		fi
+		[ "$1" -eq 1 ] || [ "$1" -eq 0 ] && echo "$1"
 	}
 
 	function getCRC {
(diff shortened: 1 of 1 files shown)


sonic2kk:
Fixed a mistake that wrote out the binary values as strings. Instead of writing out `\x00` or `\x01`, it was writing out `\x0` and then a string `0` or `1`. This is invalid, and caused Steam to invalidate the whole `shortcuts.vdf` file.

This has since been fixed, and the values look correct when inspecting the bytes.

During this, I discovered that `AllowOverlay` and `IsHidden` are actually ignored in `shortcuts.vdf` and this is instead controlled in `localconfig.vdf.`

sonic2kk:
Tested this a lot tonight and it seems to work fine. Need to test it against a Steam generated `shortcuts.vdf` and a `shortcuts.vdf` generated and updated from master.

After that this should be ready to merge with a shellcheck and a version bump. Will probably do that after work tomorrow, them take a look at fixing up some of the more recent Non-Steam property changes (tags, hidden, openvr) since they're broken at the moment. But this cleanup should be one step towards a cleaner codebase overall :-)

Hoping to make more of these PRs as I go and put a focus on cleanup once the Non-Steam game stuff is fixed. Housekeeping has not been enough of a priority and I have really only done it with features I develop (i.e. gamescope). Though some parts could still use a cleanup (ReShade+SpecialK).

Gotta get the script under 1mb again at the very least!

sonic2kk:
Testing against a Steam generated shortcuts.vdf works (added a Non-Steam Game through the Steam Client, then closed the client and added a game through SteamTinkerLaunch).

sonic2kk:
Generating a new `shortcuts.vdf` from master and adding a shortcut on this branch also appears to work, so this should be good to merge.

sonic2kk:
Shellcheck is green and version is bumped, this is ready to merge!

PR #969, opened 2023-11-07
Commandline: Add support for Non-Steam Games to getGameExe

### LINKED ISSUE #960: [Feature Request] Non-Steam Game Improvements

sonic2kk:
More work for #960.

This PR adds support for getting the Game EXE for Non-Steam Games. It returns in the same format, and can also search on game name. This also removes the need for the Non-Steam Game hack of setting the game EXE.

This shouldn't break Steam games either, all existing functionality should be untouched. This just adds support for Non-Steam Games, the same way #963 added support in `gi` and `gt`.

There is also a small refactor to those functions in this PR. We disable searching on shortcuts by default, to prevent any breakages when trying to search for unknown games names and erroneously falling back to `shortcuts.vdf` when we shouldn't. Searching on Non-Steam Games for these functions is only enabled when we're running from the commandline.

Side note: We don't write this out to the meta file, because the information in `shortcuts.vdf` is volatile, as the user can edit the game name or EXE at any time.

<hr>

This PR is ready for merging after a version bump and some more testing.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20231108-3 (nonsteam-get-exe)"
+PROGVERS="v14.0.20231108-3"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Tested it when troubleshooting another issue, it should be ready to merge.

PR #970, opened 2023-11-08
Commandline: Add support for Non-Steam Games to getCompatData

### LINKED ISSUE #960: [Feature Request] Non-Steam Game Improvements

sonic2kk:
Another piece for #960.

This PR adds support for Non-Steam Games with `getCompatData`. We already have partial support, if a game is launched with SteamTinkerLaunch, because we can get the game name and EXE already with existing functions, and we can get the compatdata with the symlink SteamTinkerLaunch creates.

However, for games not launched with SteamTinkerLaunch before, it fails. To resolve this, we can simply generate a default compatdata path. In testing, it looks like Non-Steam Games always have a compatdata created in the Steam Root library folder (i.e. `~/.local/share/Steam/steamapps/compatdata/<appid>`). Since we can parse `shortcuts.vdf` now it's very easy to loop through all shortcuts and simply generate a path like above. As a safety, we check to make sure this actually exists before returning it. We build the return string using the information from `shortcuts.vdf` even if we have information from earlier in the function, in case the shortcut name has changed or something, just to make sure we have the most up-to-date information.

There is a possibility that the compatdata location assumption is *not* correct, and that Steam actually creates the compatdata at the "default" library folder. By default Steam does set this to the Steam Root library folder, but it is possible that if Steam uses some other location that this Steam root path will not have the compatdata, and so this command will not find the compatdata. I think that's a big "if" though, so for now, this works :-)

<hr>

After a shellcheck and some more testing, I think this is good to merge.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20231108-4 (nonsteam-getcompatdata-fix)"
+PROGVERS="v14.0.20231108-4"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Shellcheck is green, doesn't seem to have broken anything, and Steam games still work. This is ready to merge.

Issue #975, opened 2023-11-14
Add Discord Update channel to SpecialK


zany130:
Been awhile been very busy :sweat_smile: 

## System Information
- SteamTinkerLaunch version: steamtinkerlaunch-v14.0.20231114-1
- Distribution: Garuda Linux

## Feature Description

SpecialK has 3 release channels.

**Stable**: this is what you get from https://sk-data.special-k.info/SpecialK.exe and https://github.com/SpecialKO/SpecialK/releases; they are the same. For our purposes, I guess GitHub is better.

**Discord beta**:  https://sk-data.special-k.info/repository/SpecialK_xx.xx.xx.xx.exe (the format is 2 digit year . 2 digit month date.exe ex:`23.11.13.2.exe` These are curated from the discord nightly by one of the devs, Aemony who is also the main dev behind SKIF (SpecialK windows installer/manager)

**Discord nightly**: I don't think there is any way to download these as they are hosted directly on Discord (as the .dll, so there is no need to extract them)

The SpecialK dll can be extracted from the exe with `inoextract`.

I believe this used to be an option, but it wouldn't download the file; instead, the user had to provide their own.

Also, the `default` and `latest` are the same version, which is the stable channel.

sonic2kk:
This is a good suggestion, but I have no good idea how to implement this. Is there any way to get the Discord beta release naame directly? We could *perhaps* try to guess, by trying to "calculate" the name from the current date and working backwards if we can't find it. But, I don't like this approach. I would prefer to be able to fetch it more directly.

It looks like SpecialK has[ GitHub Actions integration](https://github.com/SpecialKO/SpecialK/actions/workflows/build-windows.yml), we could fetch the release from here using nightly.link. The naming scheme seems similar, except the actions have a commit hash i.e., the current latest is `SpecialK_23.11.13.2_83bed401`, where `83bed401` is the hash that the workflow built from. These workflows don't even come with SKIF, they are the raw 32bbit/64bit DLLs, which is perfect for our case, we don't even need `innoextract`.

Though the builds on Discord are "curated", I don't think we need to care about that. These builds are probably more in line with the "Discord Nightly" release, and I would be fine with simply naming this channel "Nightly".

If we can get the file and the DLL into a folder matching the dropdown selection name (i.e. `discord`), the rest of the logic should just work, since we just pick it from the folder. However, picking the release is the tricky part.

<hr>

[...]

PR #977, opened 2023-11-15
SpecialK: Add option to download latest nightly from GitHub Actions

sonic2kk:
Implements most of #975.

This PR adds logic to fetch the latest GitHub Actions nightly archive. Since the archive name is not fixed, we do this by using `fetchLatestGitHubActionsBuild` to get the latest workflow commit hash for the relevant workflow name we want, then build the nightly.link URL for the page that will list the latest GitHub Actions build, wget it, and parse out the artifact archive nightly.link download URL we want based on a pattern for how our desired archive name will look.

Once we have this, we set the basename of this ZIP download URL to the `SPEKARC` which is the destination path for the archive. `extSpek` takes care of extracting it, and the rest of the paths are built dynamically based on the version name, so the rest of the SpecialK installation logic works without any tweaks.

We did have to make a couple of changes to achieve this:
- `SPEKDL` is now set after wee set the `SPEKDLURL`, because the `nightly` option overrides `SPEKARC` since in this case, it's a ZIP instead of a 7z file.
- `extSpek` can now extract ZIP archives and checks the file extension to do this
- `fetchLatestGitHubActionsBuild` was tweaked to now optionally  take a name to fetch the first workflow from. For SpecialK it was pulling from its `CodeQL` workflow. Now we filter the workflow API response with `jq`.

This branch was tested in #975 and should work, and it also works in my tests, so I will merge it soon. In future, other changes to allow custom SpecialK versions may be implemented as well.


sonic2kk:
Need to bump langfiles.

sonic2kk:
Added some more logging, updated langfiles, and bumped the version. ShellCheck is green on this so I think it's ready to merge. I also gave a brief mention to the nightly release on our [SpecialK wiki page](https://github.com/sonic2kk/steamtinkerlaunch/wiki/SpecialK).

PR #979, opened 2023-11-18
SpecialK: Allow Custom SpecialK Versions

sonic2kk:
Implements #975.

## Overview
This PR adds a new SpecialK version called "custom", and removes the old `default` and `latest` versions in favour of simply `stable` which is equivalent to the older `latest`. Previously,`default` pointed to GitHub and `latest` pointed to the fixed download link on the SpecialK website. There is a minor (1 month?) difference in these versions.

We also remove the `createDLSpekList` since it isn't used anymore. This was created when we downloaded the SpecialK versions list from the GitHub versions, but now we instead have a fixed list of versions we download from since a while ago (before even this PR), so we never needed to use this. It would just default to appending `latest`, since the version list was supposed to look like `latest!x.y.z!a.b.c!d.e.f` etc. Just a bit of cleanup that should not cause any regressions :-) Hopefully...

### Custom SpecialK
When using the `custom` version, either selected on the Game Menu or when "downloaded" with `steamtinkerlaunch specialk download custom`, SteamTinkerLaunch will attempt to do the following:
- Look for `SpecialK32.dll` and `SpecialK64.dll` in the SpecialK download directory, in a folder called `custom`
- If not found, try to extract these DLLs from a manually placed SpecialK installer DLL

image

The idea behind this is to allow more flexibility with installing SpecialK versions that STL may not be able to fetch, such as those with hardcoded URLs, Discord previews, or other custom builds. This also allows for restoring the ability to use the older GitHub releases, as someone can drop the DLLs or installer into this folder, and STL will pick it up.

This also allows for using an older nightly, since we always point to the latest nightly and there may be a regression. A user can get the previously working build artifact and put the DLLs from it into the `custom` folder.

There is **no** fallback if a custom version is not found, so SpecialK will simply not be used.

### Stable SpecialK
The `stable` option replaces the old `latest` option. This makes the dropdown cleaner and since there is probably no real reason a lot of the time to use `default`, this change streamlines the SpecialK version selection.

### SpecialK Download Checks
The logic for checking existing SpecialK downloads has also changed in two main ways.

First, if `AUTOSPEK` is off, instead of checking if the archive exists (which a user may manually remove) we instead now check if `SpecialK32.dll` and `SpecialK64.dll` are downloaded. This acts as a better valid download check, since we need these DLLs. It makes more sense to check that the two files we need are already present instead of checking on the archive name.

The only other change around this is that after a download, we remove anything from SPEKVERS folder that isn't `SpecialK32.dll` or `SpecialK64.dll` to clean up after ourselves. There's no reason to keep the archive hanging around.

<hr>

This PR was tested pretty well in #975. But before I merge this PR I will need to give the wiki a touch up to reflect the new versions, what they point to, and how to use them. The SpecialK wiki probably just needs a revamp at this point :sweat_smile:   

TODO:
- [x] Investigate notifier and more logging? i.e. don't say "downloading SpecialK", say "extracting" or "looking for" custom SpecialK, etc
- [x] Automatically create `custom` folder?
- [x] Update wiki


sonic2kk:
Added notifier (no extra logging was needed), we also now auto-create the SpecialK `custom` download dir anytime SpecialK is downloaded. Didn't make sense to create this any other time, since we don't explicitly create the SpecialK DL directory outside of `dlSpecialK`.

The langfiles needed updated for the notifier commit, so now we have to bump them too.

Next up, resolve conflicts, update wiki, and then bump langfiles. Then we can merge.

sonic2kk:
[SpecialK wiki page](https://github.com/sonic2kk/steamtinkerlaunch/wiki/SpecialK) has been updated.

sonic2kk:
Feature has been tested and seems to work, refactor to use simply `stable` seems to work also, notifier added, langfiles bumped, version bumped, just ShellCheck is green. This is ready to merge!

PR #1000, opened 2023-12-27
ReShade: Add Toggle for Creating INI

### LINKED ISSUE #894: ReShade compatibility with Special K

sonic2kk:
Another piece for #894.

## Overview
This PR adds a checkbox to toggle whether or not we should create the ReShade INI. This option is ON by default. This was requested in #894, to allow some new SpecialK functionality that won't work with an existing ReShade INI. SteamTinkerLaunch will create this each time a game is launched and uses ReShade, so this PR adds a checkbox to disable it.

image

This checkbox will not remove any existing ReShade INI, so nothing will be lost. But this means a user will have to manually remove the ReShade INI file. We will need to document this on the wiki.

For readibility, I also took this chance to split this logic into a separate function called `createReShadeINI`. This means we can avoid some nested conditionals in this huge `checkReshade` function. We should really do this elsewhere in the codebase.

## Remaining Work
The main thing left to do here is test, and update the ReShade and SpecialK wiki pages for this option. The ReShade wiki can note that this option can disable creating an INI, meaning a user can use their own existing one if they want more easily, and also we should not that this will not remove an existing ReShade INI, to further avoid confusion. On the SpecialK wiki page, we can note the use case for this ReShade checkbox under the ReShade compatibility section, linking back to here and the comment in #894.

<hr>

TODO:
- [x] Further testing
- [x] Update wiki
- [x] Update langfiles
- [x] Version bump


sonic2kk:
Been sick and haven't gotten around to doing anything with this, but I recall an issue with one game had this disabled by default for some reason (may have been a bad config file, some boolean vars in it were empty without quotes and had blank quotes). Will have to test this some more hopefully later this week (famous last words).

Did a quick test with the offending game and it worked. Will test a game I previously tested ReShade with (to simulate a normal path going from master -> this feature), and then test this branch with a fresh game that I have never used ReShade with before (should be good to verify a brand new game will use the INI by default and will work when the checkbox in this PR is disabled).

sonic2kk:
"Create ReShade INI" is correctly enabled by default in 2 games I tested now and will create the INI (HoloCure, NieR:Automata).

~~This also appears to work correctly when using ReShade + SpecialK (NieR:Replicant). Only tested quickly though. The `checkReshade` path that we create the INI on is called before any ReShade+SpecialK specific checks so we should be fine on this front to begin with.~~ Sometimes (though not always), SpecialK will create a ReShade INI file itself when using ReShade+SpecialK. This is for both using ReShade as a SpecialK plugin, and loading them separately. SpecialK is definitely creating this, because the content is different than what STL creates (STL only creates a few lines, these files are 70+ lines).

sonic2kk:
Actual feature appears to work as expected. Will need to prepare some wiki updates, bump langfiles and bump version, and then this can be merged. Once I'm feeling better I'll write up those wiki changes.

sonic2kk:
Looks like only the ReShade wiki page needs updated, the SpecialK page simply links to the ReShade wiki page on using ReShade and SpecialK together for more information (https://github.com/sonic2kk/steamtinkerlaunch/wiki/SpecialK#reshade-and-specialk).

Wrote up the ReShade wiki update, will bump langfiles and version and merge this PR. It appears to work as expected and the other changes are ready to go.

sonic2kk:
Version bumped, langfiles updated, ShellCheck is green, this is ready to merge!

sonic2kk:
ReShade Wiki has also been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/ReShade/5809de44c2b175362890e90895f8fde6f428c613

PR #1008, opened 2024-01-11
Langfiles: Remove remaining mentions of Cheat Engine

sonic2kk:
Remove remaining mention of Cheat Engine from langfiles. These strings are unused, and were missed in f791df8575b1f24c29342b0a7282d5d4f0e30ff9.


sonic2kk:
Will confirm that only CheatEngine-related strings were removed and do a quick test to ensure no crashes/regressions tomorrow. After that I will merge this. Should be a straightforward change.

There may be other unused language strings that I will check out and perhaps remove in a future PR, as a bit of a cleanup effort before v14.0.

sonic2kk:
Langfile diff looks good, bumped the version, this should be fine to merge now. Did a quick launch test and didn't have any issues.

PR #1007, opened 2024-01-10
SteamGridDB: Implement GUI for getSteamGridDBArtwork command

sonic2kk:
Implements #933.

## Overview
This PR adds a GUI dialog frontend for the `commandlineGetSteamGridDBArtwork`  SteamGridDB command, fully implementing the work for #933 and completing the SteamGridDB Overhaul. It allows a user to enter information on the frontend instead of manually entering it all on the commandline, which is a small Quality of Life enhancement.

The options on this dialog are essentially a form for taking input for the SteamGridDB Command.

image

## Implementation
Not much to add to this since the bulk of the implementation detail is for `commandlineGetSteamGridDBArtwork`.

Usage: `steamtinkerlaunch sgdb <appid>`

This works by the user running the `sgdb` command and give it an AppID, and then a dialog appears, allowing them to enter information to download artwork from SteamGridDB for that specific game. They can search for artwork using a Steam AppID (it doesn't have to be the AppID of the current game, i.e. they can get artwork for a version of a game on Steam that they don't own if it has better artwork), the SteamGridDB Game ID, or they can attempt to get artwork from a fuzzy Game Name search. SteamGridDB has an endpoint for fetching artwork by Game Name. This is all implemented on the `commandlineGetSteamGridDBArtwork` function, we just massage some of the entered data to make it easier to pass to the command.

#### Managing Game Type Search
We use a variable to denote if we pass `--steam` or `--nonsteam` to the `commandlineGetSteamGridDBArtwork` command, and we set this variable based on whether the user 

If the user enters a Steam AppID, it will take priority and we will assume we're searching for artwork for a Steam game. Otherwise, we will search based on the SteamGridDB Game ID and assume we're using its game endpoint.

Since the command takes the AppID of a specific Steam game, we can force the filename AppID to this file, as we have the context of what AppID the user wants to set the SteamGridDB game artwork for. Therefore we should be able to search for artwork for a Non-Steam Game using a Steam AppID (for example, if using the GOG release of Oblivion, the user could still use the Steam AppID to fetch the game artwork, since the grids will use the `--filename-appid`).

### Some Specified Defaults
For selecting how to handle existing grids, we have a combobox, and this defaults to the option selected on the Global Menu just like the command does, except we select it visually as the default option on the GUI.

We also default to having the "Apply" option enabled, as it is probably sensible to have this on by default for the GUI portion of this work. A user can toggle it off and it is more likely that from the GUI a user will want the default behaviour to be that grids are set on Steam.

## Remaining Work
[...]

PR #1011, opened 2024-01-12
Commandline: Refactor getTitleFromID to be more reliable with Non-Steam Games

sonic2kk:
## Overview
I noticed an issue with a Non-Steam Game that I have launched with SteamTinkerLaunch before not returning the game title correctly with `getTitleFromID`. I found this while working on #1007, where using it for Non-Steam Games sometimes had a blank name. I narrowed the cause down to the game having an STL meta file that was missing the title (since we don't have the title for Non-Steam Games). If we didn't find the title in the meta file, we weren't continuing to search, which was causing the issue since Non-Steam Games have a meta file but not a title, and so we were never continuing beyond this to search on `shortcuts.vdf`.

This PR refactors `getTitleFromID` significantly, making it less nested and now checking on SteamTinkerLaunch meta file, then falling back to appmanifest, and then falling back to `shortcuts.vdf` to try and find the corresponding Title for a given AppID. Previously, we were checking only one of these. This meant if there was a meta file without a title (which is potentially the case for Non-Steam Games), we would not find the App Name.

## Implementation
The function has been refactored so that we use a variable to track the game title. We set this each time we search for the game title. If it is blank, we assume we can move onto the next search method. It works like this:
- By default, `FOUNDGAMETITLE` is blank.
- If we have a SteamTinkerLaunch metadata file, search for the game title in it and store the result in `FOUNDGAMETITLE`.
- If `FOUNDGAMETITLE` is still blank after this, search the Steam AppManifest file for a `name` key and store the result in `FOUNDGAMETITLE` again.
- If `FOUNDGAMETITLE` is once again still blank, and if we have Steam Shortcuts that we have enabled searching on (`SEARCHSTEAMSHORTCUTS`), then search `shortcuts.vdf` for the Game Title. If we get a match, store the Game Title for this shortcut entry in `FOUNDGAMETITLE`
    - If there are no shortcuts OR if searching on Steam Shortcuts is disabled, this step is skipped
- If after all of this, `FOUNDGAMETITLE` is still empty, display that no game is found.
- Otherwise, echo out the game title at the end of the function.

To me, this is much more logical and also easier to read.

<hr>

This PR needs a bit more testing before merging. We need to test that game names are still displayed correctly on the Game Menu, and ensure that any parts of the code that use `getTitleFromID` still function correctly. I tested quickly with `getExe` and it still works correctly.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7834,40 +7834,51 @@ function getIDFromTitle {
 
 function getTitleFromID {
 	SEARCHSTEAMSHORTCUTS="${2:-0}"  # Default to not searching Steam shortcuts
+	FOUNDGAMETITLE=""
 
 	if [ -z "$1" ]; then
 		echo "A Game ID is required as argument"
-	else
-		if [ -f "$GEMETA/${1}.conf" ]; then
-			if grep -q "^NAME" "$GEMETA/${1}.conf"; then
-				grep "^NAME" "$GEMETA/${1}.conf" | cut -d '"' -f2
-			else
-				getAppInfoData "$1" "name"
-			fi
+		return 1;
+	fi
+
+	# Search meta file if we have one for Game Title
+	if [ -f "$GEMETA/${1}.conf" ]; then
+		if grep -q "^NAME" "$GEMETA/${1}.conf"; then
+			FOUNDGAMETITLE="$( grep "^NAME" "$GEMETA/${1}.conf" | cut -d '"' -f2 )"
 		else
-			# Try to get AppManifest based on passed ID if we don't have a data file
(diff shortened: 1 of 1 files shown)


sonic2kk:
This seems to work fine. We will need to update usage of `getTitleFromID` in various places (`getSteamGridDBArtworkGUI`, `setGameArt`, `fixShowGnAid`) to search on Steam Shortcuts as well as currently they will not display the title for Non-Steam Games. That change will go after this PR though.

sonic2kk:
Tested and this doesn't break displaying the game name on the Wait Requester/Main Menu, generating a fresh meta file also works for these games. Should be a safe merge.

sonic2kk:
Following this PR, fixed the mentioned functions above not pulling Non-Steam Game titles when using `getTitleFromID` (2fb17b8cadd13067c5554b6feafd58eb963851a2).

Issue #1016, opened 2024-01-15
Steamtinkerlaunch crashing with “installing modorganizer” page


Mbhuff03:
As far as I can tell, steamtinkerlaunch was working fine two days ago. Today I launched a game on steam deck, and I typically go into STL menu periodically to see if I’m using the latest proton. 

I like to use aurora from Cheathappens for some games (don’t hate, it saves time and I don’t use it on multiplayer) which requires STL. 

Today I updated it to the latest proton experimental and it crashed the load into the game. Tried a few other games also and they crashed as well. 

The only symptom is the screen right before the crash with “installing modorganizer” and a load bar at the top. This displays for a second or two (load bar doesn’t seem to load anything) and it goes back to the game launch screen. When I try to relaunch the game, it won’t even show the STL menu before going back to the install modorganizer screen. 

I’ve tried to uninstall and reinstall STL a few times using protonup-qt. Checked that yad was installed by using the “steamtinkerlaunch yad ai” command. 

I can’t figure out what’s wrong. 

I WAS able to reload the STL menu on a fresh install and get the game to open. But when I try to load aurora again, it seems to cause the crash again.   I realize this could be an issue with aurora, but since the games would launch with aurora fine a few days ago and with old proton builds, I don’t think it should cause such an issue.

sonic2kk:
**Do not ignore the issue template, this issue will be closed as you did not follow it. You NEED to provide logs when reporting reproducible bugs.**

Someone else reported something similar and said a clean install fixed it. Very strange, as I haven't seen this on the Linux Desktop. This seems to be limited to Steam Deck though.

I no use SteamTinkerLaunch on my Steam Deck so I cannot test. I have no idea what Aurora is and do not support tinkering on Steam Deck, so I cannot give much help.

Are you definitely using the latest SteamTinkerLaunch from git and not v12.12?

sonic2kk:
Upon re-reading, it seems this Aurora tool uses SteamTinkerLaunch. You should report this downstream then and see if they can figure it out, maybe they're trying to install ModOrganizer 2?

> Checked that yad was installed by using the “steamtinkerlaunch yad ai” command.

Why? SteamTinkerLaunch installs Yad itself on SteamOS, which the Steam Deck wiki page tells you. It is installed to `~/stl/deps/usr/bin` if memory serves.

<hr>

[...]

Issue #1012, opened 2024-01-13
Harvestella won't run on stl(cannot find usable proton version)


Himoteph:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: Steamdeck SteamOS 3.5.7
- Installation Method: ProtonUp-Qt

## Issue Description
Whenever I try to run Harvestella with steam tinker launch as a compatibility tool, and set it up to fork with another program(have tried with a trainer and cheat engine), I get an installing ModOrganizer page, then it crashes. Subsequent attempts to run the game with stl lead it to bypass the option to open the menu, and jump straight to the ModOrganizer install screen. Has not occurred with any other games in my library.

## Logs

[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/13926307/steamtinkerlaunch.log)

Himoteph:
Also occurs if I completely uninstall 12.12, and install 12.0 instead.

sonic2kk:
SteamTinkerLaunch V12.12 is very old. ModOrganizer 2 doesn't work with v12.12 either which is noted on the wiki and closed issues already. It tries to use the latest MO2 which is v2.5 and it's not compatible with Valve Wine yet. Use SteamTinkerLaunch from git, it's available from ProtonUp-Qt. You'll have to uninstall MO2 from the game prefix and from the global compatdata at `~/.config/steamtinkerlaunch/mo2`.

If STL can't find a usable Proton version, try removing `/dev/shm/steamtinkerlaunch`.

> Distribution: no idea

What do you mean? How do you not know what Linux distribution you installed on your computer?

If you are not an experienced Linux user, I don't recommend using an enthusiast tool like SteamTinkerLaunch.

<hr>

I will also need you to be a lot more specific here. You say the game won't run, then you mention MO2 suddenly. Does it work standalone or only show this issue with MO2? Is your issue about MO2 or the game not running vanilla?

sonic2kk:
Inspecting the log, it looks like no Proton version was found for MO2 or the game as you state, but a lot of things about this log seem really, really wrong. It could simply be because v12.12 is so outdated though.

After clearing `/dev/shm/steamtinkerlaunch`, re-run STL and if the issue persists, inspect `/dev/shm/steamtinkerlaunch/ProtonCSV.txt` to see what STL was actually able to find.

[...]

PR #1017, opened 2024-01-16
ModOrganizer 2: Prevent MO2MODE from being 'none', only checkMO2 if MO2MODE is 'GUI'

sonic2kk:
Resolves an issue observed in the log for #1012.
May help #1016, very little information was provided to dianogse if it was the same problem.

## Overview
It has been reported that a couple of games somehow get into a state where `MO2MODE` is `none` (or `$NON`). This shouldn't happen, and so far has only been actually confirmed to happen with one game for one user and only on SteamOS (#1012, happened with Harvestella). A fresh install fixes this issue. In the log for #1012, I observed `MO2MODE` was `none`.

This would explain two things:
1. ModOrganizer 2 is launched in `checkMO2` if `MO2MODE != disabled`. Of course, `none != disabled`, so it tries to launch whenever a game config gets in this state, where `MO2MODE = none`.
2. A clean install fixes this because it re-creates the config files, thus re-creating the config file properly.

Nowhere in the code is `MO2MODE` set to `none`, so this must be some config file defaulting that incorrectly sets the value. However, none of my config files have this value, including a couple of games never launched with SteamTinkerLaunch until just now. So I'm not sure how this actually happens. I suspect it could be somehow isolated to Steam Deck usage, but I don't know what would be causing it there.

## Implementation
As a fix, I have made the followiing changes:
1. Add a migration in `migrateCfgOption` that forces `MO2MODE = disabled` if `MO2MODE` is `none`. This should fix config files that got in a bad state somehow with this invalid value.
2. Change `checkMO2` to only launch on `$MO2MODE = gui`. Previously there were plans for a silent mode, but as I don't use MO2 and don't work on that, and no one else has shown any interest, for now we will only launch on `gui`. This makes sense
3. Exit early in `checkMO2` if we have `MO2MODE = none`. This should never happen now because of `migrateCfgOption`, but it is an extra bit of insurance. We also have some logging for this.
4. When a user tries to launch MO2 in silent mode, since this is not implemented, default to `gui`. This was not an issue before since we only prevented launch on `MO2MODE = disabled`, but now we are skipping that.
5. Exit early if `MO2MODE = disabled`, not all that important to functionality but results in cleaner code.

This all needs extensively tested and some keen review from myself, but this should hopefully resolve the problems. If more information is provided in #1016 we can see if it resolves that issue too.

<hr>

I suspect, somehow, this is related to custom commands in the case of #1012. But I have no idea, really. At the very least this should prevent it from happening.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240116-5 (mo2-migrate-none)"
+PROGVERS="v14.0.20240116-5"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -6958,7 +6958,7 @@ function migrateCfgOption {
 			updateConfigEntry "SPEKVERS" "$SPEKVERS" "$STLGAMECFG"
 		fi
 
-		# stop MO2MODE from being 'none' in weird scenarios
+		# stop MO2MODE from being 'none' in weird scenarios -- Issue was confirmed to be exclusive to SteamOS and may be a (temporary?) SteamOS regression
 		if [ "$MO2MODE" == "$NON" ]; then
 			MO2MODE="disabled"
 			touch "$FUPDATE"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240116-3"
+PROGVERS="v14.0.20240116-4 (mo2-migrate-none)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Tested a game that uses ModOrganizer 2 on my system (The Elder Scrolls IV: Oblivion Game of the Year, AppID 22330). Works as expected. MO2 starts up properly when enabled.

sonic2kk:
Tested a Proton game that does not use ModOrganizer 2 (PowerWash Simulator, AppID 1290000). Works as expected, no MO2 prompt.

I inspected the calls to `checkMO2` and they look the same as before but with extra logging, so the flow should be normal.

sonic2kk:
Tested forcing `MO2MODE` to `none` in the game config file for a Proton game (Travellers Rest). It was migrated to `disabled` correctly on startup. When launching the game via STL, `checkMO2` is correctly skipped because `MO2MODE` is `disabled`.

sonic2kk:
I considered an option to migrate if MO2MODE was not `gui` or `disabled`, but we don't know how this will change in future. This would be brittle, and even though currently `gui` and `disabled` are the only valid UI values, there is no need for this change now.

This PR updates `checkMO2` to only launch on `gui` (and this is an easy change if we add `silent` or any other future mode), and if the `MO2MODE` values get in a bad state that isn't `none`, MO2 won't try to launch / install now. Also, should that happen, it can be resolved on the GUI, and we have better logging now.

sonic2kk:
There seems to be no regression in behaviour, this should work as expected. I have asked another user to investigate to see if this problem is "widespread", and the OP of #1016 to provide some more details and optionally to test if this PR resolves their issue.

This is ready to merge but I am waiting for more feedback from various places before merging.

sonic2kk:
There is potential in future to restructure `checkMO2` to be less nested, but that is for another PR.

sonic2kk:
Issue is confirmed to be reproducible on and exclusive to SteamOS, not sure why though. Maybe related to #935, but surely it would've come up before now.

I will merge as-is after bumping the version, although this is more of a hack than a real fix. I would rather not leave this PR open though and have this issue remain widespread.

Issue #1024, opened 2024-01-24
Issues with gamescope and gamemode


Martmists-GH:
## System Information
- SteamTinkerLaunch version: 12.12-1
- Distribution: Arch Linux
- Installation Method: yay (package manager)

## Issue Description
Enabling gamescope sometimes errors with `/usr/bin/gamescope: unrecognized option '--verb=waitforexitandrun'`

Gamemode on the other hand is unable to find libgamemodeauto.so.0, which is located at `/usr/lib32/libgamemodeauto.so.0`

## Logs
[steamtinkerlaunch.log](https://transfer.sh/WLp1jPtKhe/steamtinkerlaunch.log)

sonic2kk:
In future, please upload your log as a file to GitHub and not to a filehost. For what it's worth, the filehost also warned that your log was laced with a virus. I have found no such traces, however.

Your log looks mostly normal. However, your GameMode path looks wrong. Check where it is installed with `which gamemoderun`, and ensure that you have the 64bit version installed. STL is building the launch command correctly:

```
/usr/bin/gamemoderun /usr/bin/gamescope -w 1920 -h 1080 -W 3840 -H 2160 -r 144 -f -e --force-grab-cursor --adaptive-sync --prefer-vk-device -- /home/mart/.local/share/Steam/steamapps/common/SteamLinuxRuntime_sniper/_v2-entry-point --verb=waitforexitandrun -- /usr/share/steam/compatibilitytools.d/proton-ge-custom/proton waitforexitandrun /home/mart/.local/share/Steam/steamapps/common/Palworld/Palworld.exe
```

The GameMode error may be a false positive coming from a Proton log, unless you can confirm the GameMode daemon does not start. However, if it does not start, that is likely because of an incorrect path as GameMode **should be** at `/usr/bin/gamemode`.

Outside of this, you're using a heavily outdated version of SteamTinkerLaunch, it is not a good idea to use "stable" versions as you end up missing out on features and fixes. You could try updating SteamTinkerLaunch and to match, your version of GameScope (ensuring you're using the latest git version of both ofc).

The GameScope error is really strange and sounds like something is being built incorrectly in the launch command, as `--verb=waitforexitandrun` should get passed to Proton and not GameScope - which, in the command in your log, it is and this looks like a correct, runnable start command. In fact, you should be able to paste that into terminal manually I think and the game should start.

[...]

PR #1026, opened 2024-01-27
GameScope: Add Mouse Sensitivity

sonic2kk:
GameScope has recently added support for configuring mouse sensitivity with ValveSoftware/gamescope@87834648e61964889de7971a370b084b00de83fb. This is a decimal value that the base mouse sensitivity is multiplied by. This PR exposes it on our GameScope GUI.

image

This option didn't seem to do anything in my tests, both from the SteamTinkerLaunch GUI and manually with games from the commandline. I am using the right version as the option shows up in `--help`.

<hr>

TODO:
- [ ] Further testing
- [ ] Update langfiles
- [ ] Update wiki screenshot


sonic2kk:
It will go in a seperate PR, but I am going to refactor how we index the arrays. It's getting way too annoying to change each index when we bump an option.

sonic2kk:
Something strange is going on. When `--adaptive-sync` is enabled, `GSMOUSESENSITIVITY` is set to `--`. Otherwise, it works fine.

Something weird is going on, potentially related to #1024.

sonic2kk:
Weirdness should be fixed with #1027.

In that PR I used the example of `-r` and `--rt`, but here the issue is likely related to the `-s` in `--adaptive-sync` (`-sync` -> `-s`).

sonic2kk:
Rebased on master after #1027, fixed an issue where we were trying to compare decimals, and the weirdness is gone.

Game launch works OK in my tests too with this option, and when the option is `1`, we correctly ignore it and don't append it to the GameScope arguments. It also seems to have an effect with Steam games, unlike my earlier tests.

I will update the langfiles and bump the version and then merge.

sonic2kk:
Langfiles updated, version bumped, shellcheck is green, we are ready to merge!

PR #1028, opened 2024-01-27
GameScope: Refactor GSARR indexing

sonic2kk:
This PR refactors how we parse and index GSARR in `GameScopeGui`. Previously we were hardcoding each index, and this was brittle. When we added a new element, everything had to get shifted down by one. This PR uses a "relative positioning" system, where everything indexed relative to the index of the heading above it. We calculate heading positions using a hardcoded number of elements per section.

If we want to add a new element to, say, the HDR section, we can do it much more simply now. All we have to do is bump `GSHDROPTSLEN` by `1`, and then add the element in. To position the element we can calculate its index based on its number in the section (i.e. if it's the 5th section, then it'll be `GSFILTERINGHEADING + 5`). If a new element is inserted into the middle of a section, we only need to re-organise the indexes in that section and not the whole list. Instead of changing all 60+ lines for how we parse the GameScope variables out of `GSARR`, we only need to change a few lines. At worst if something went at the top of the section, we would only need to update the indexes for the variables in that section (as well as the variable tracking the length of the section, since this is used to position the other headings).

It is much easier to format things this way, and having the variable tracking the length of each section is also a good way for us to check our work and make sure each section is the correct and expected length.

<hr>

In testing, this seems to work with existing and fresh GameScope configurations. Everything is parsed correctly from existing configs, even complex ones like `GAMESCOPE_ARGS="-w 1920 -h 1080 -W 3840 -H 2160 -r 144 -f --force-grab-cursor -s 7 -F fsr -S stretch -m 3 --fsr-sharpness 8 -C 5 --rt --"`. I will do more testing though and if this works fine, I will bump the version and merge.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240127-2"
+PROGVERS="v14.0.20240127-3 (gamescope-gsarr-refactor)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
This seems to work fine. I think we can merge this as-is after a version bump. Tried a couple of games (Cookie Clicker, NieR:Automata) and they work. The new Mouse Sensitivity option also still works.

sonic2kk:
ShellCheck is green and version is bumped, works in testing as noted, this is ready to merge!

Let us hope it doesn't result in disaster :sweat_smile:

PR #1029, opened 2024-01-27
Remove Depressurizer Support

sonic2kk:
This PR removes support for [Depressurizer](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Depressurizer). I do not use this tool or know anything about it, I'm also not sure if anyone uses it with STL. I heard a couple of times in passing in unrelated issues that it doesn't work anymore.

This PR is pretty sweeping and will need testing before it is merged. If someone is using it and/or if someone is vocal about keeping it, then this PR can be closed.

**I am not planning to merge this PR until after v14.0**. There have been significant changes since the last release, and someone may want to restore Depressurizer support on top of v14.0. Not merging until after release will give a more functional base release to work with restoring and fixing. Of course if anyone does this they're free to merge it back upstream if they are interested in keeping Depressurizer in STL (some people may not be vocal about a change until it is already made).

Opening as a draft as unsure yet if it should be removed. It has also not been tested outside of running a quick ShellCheck.


sonic2kk:
The current plan is to go forward with this after v14.0. Should probably rebase this regularly to make sure it's ready for merging come the time, just in case.

If someone wants to fork and re-add Depressurizer for their own purposes, they should be able to use the diff of this PR to do so. This also needs some more testing to make sure nothing unrelated is broken, so this will stay in draft for a while longer.

PR #1031, opened 2024-01-27
Commandline: Flatten getGameExe checks

sonic2kk:
More recent codebase additions have been using the (in my opinion, more readable) return-early paradigm. This flattens the checks instead of containing everything in a big `else` block. This was recently done for `getTitleFromID` in #1011.

In testing, this seems to work just as well, but I will do more testing before merging.


sonic2kk:
Bumped version with a rebase, works in testing. ShellCheck is green. This should be ready to merge.

PR #1032, opened 2024-01-27
Commandline: Flatten and refactor getCompatData

sonic2kk:
Along the lines of #1031 and #1011, but also a slightly more significant refactor.

This PR flattens the structure of `getCompatData` , making it return early instead of having a bunch of nesting. It also removes an unnecessary codepath. We had some weird check for `STLCOMPDAT` existing and we would error out if it didn't. This was probably a holdover from the time when this function relied *entirely* on `STLCOMPDAT` to parse symlinks. We no longer do this. Instead, this check now wraps our symlink check. So we only check `STLCOMPDAT` for symlinks if it exists, and if we find a path to the compatdata we're looking for, we return it here and exit early. This makes much more logical sense to me; if we didn't find the path in `STLCOMPDAT` we were moving on to check for the compatdata in the library folder manually *anyway*, so there's no need to error out if the STL folder simply doesn't exist (which is almost guaranteed to never be the case as STL verifies its folder structure).

`STLCOMPDAT` just points to, for example, `~/.config/steamtinkerlaunch/compatdata`. It doesn't represent any specific compdatdata, so this check was a pointless check on the STL folder structure. There is no need to error out. Honestly, the `-d` check is kind of pointless to begin with, but it's for safety in case `find` gives us unexpected results if `-d` is not a valid directory.

Apart from that, it's just flattening the logic to remove nesting.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240128-2"
+PROGVERS="v14.0.20240128-3"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Needs a rebase and some more testing, but ShellCheck is green.

sonic2kk:
Appears to work in testing, can probably rebase and merge.

sonic2kk:
ShellCheck is green still after a rebase and this seems to work fine, can merge.

PR #1033, opened 2024-01-27
Commandline: Flatten getGameDir

sonic2kk:
Part of some ongoing work along the lines of #1031, and #1032 (and part of #1011).

This PR flattens out the nested logic in `getGameDir`, to make it more maintainable and easier to read. There wasn't really any refactors here like in #1032, this was more like #1031 which just got flattened.

PR #1034, opened 2024-01-27
x64dbg: Flatten dlX64Dbg function

sonic2kk:
When working on #1031, #1032, #1033, this function happened to catch my eye as another that could use a refactor. There's going to be a lot of these around the codebase, I'll tackle them as I spot them. These PRs can also serve as reference for others that spot these changes.

Refactors like this are good to do every once in a while :-)


sonic2kk:
Verified that the behaviour is the same on master and this branch using a hardcoded `debug` call to `dlX64Dbg`. ShellCheck is also green on this.

Just needs a version bump and we can merge.

PR #1035, opened 2024-01-27
SBS: Flatten some SBSVR-related functions

sonic2kk:
Building on the trend of recent PRs.

These functions were fairly ugly and difficult to read. They are also closely related as they're all SBSVR-related.

ShellCheck is green on this, but I don't have a headset to check on this with. So I'll give it a review and merge when I feel like the logic is sound.


sonic2kk:
This has been stale for a while but I think it's still good. Will do a ShellCheck and merge.

PR #1036, opened 2024-01-28
Commandline: Flatten getIDFromTitle

sonic2kk:
More work related to #1031, #1032, #1033, #1034, #1035.

Removes the nesting from `getIDFromTitle`.


sonic2kk:
Tested this out and it returns the same results as `master`, and also still works for Non-Steam Games. ShellCheck is also green. Should be good to merge.

PR #1041, opened 2024-02-16
Zink: Properly enable if used with Primerun

### LINKED ISSUE #1040: Zink and Prime-Run Compatibility
To ensure compatibility between prime-run and zink, the zink variables must be loaded after the prime-run ones (this is because both of them set  `__GLX_VENDOR_LIBRARY_NAME` and for Zink to activate this must be set to mesa.

Currently it seems the prime-run variables are loaded after the zink ones thus leading to no zink if prime-run is also enabled.

To illustrate:

`zinkrun prime-run mangohud glxgears`

`prime-run zinkrun mangohud glxgears`

sonic2kk:
Fixes #1040.

Load Zink after Primerun in order to allow Zink to set __GLX_VENDOR_LIBRARY_NAME environment variable, as Primerun will override this and not load Zink.

Thanks to @Rabcor for reporting and detailing the required fix! This can be merged once the fix is confirmed on their end.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240131-1"
+PROGVERS="v14.0.20240216-1 (primerun-zink-change)"
 PROGCMD="${0##*/}"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Confirmed to fix the issue in #1040, didn't break regular Zink usage in my testing, version bumped, and ShellCheck is green. This is ready to merge!

PR #1050, opened 2024-03-06
GameScope: Append '--' to end of GAMESCOPEARGSARR if it is missing

### LINKED ISSUE #1048: Enableing gamescope crashes games
## System Information
- SteamTinkerLaunch version: 14.0.20240218-1
- Distribution: Garuda Linux
- Installation Method:AUR

## Issue Description
Games crash wen gamescope is enabled in stl settings. running the game with the same launch commands (tools) WITHOUT stl works fine with gamescope

Tested this by running a hat in time with stl and no extra tools/options enabled except gamescope (with no arguments). Game crash with the following in the log `ERROR - startGame - Could not determine pid of '/home/zany130/.local/share/Steam/compatibilitytools.d/GE-Proton8-31/files/bin/wineserver'`

Then I tried running a hat in time with stl with gamescope disabled. Game ran fine

Then running the same game with `gamescope -- %command%` in steam with proton-GE set as the compat tool instead of stl. Game ran fine

I tried repeating the tests after a reboot as I have seen that pid error go away before with a reboot (maybe a lingering wine processes or something) and I got the same results, enabling gamescope gives `ERROR - startGame - Could not determine pid of '/home/zany130/.local/share/Steam/compatibilitytools.d/GE-Proton8-31/files/bin/wineserver'` disabling gamescope runs the game as expected

## Logs
[noGamescope.log](https://github.com/sonic2kk/steamtinkerlaunch/files/14489775/noGamescope.log)
[Gamescope.log](https://github.com/sonic2kk/steamtinkerlaunch/files/14489855/Gamescope.log)

sonic2kk:
Fixes #1048.
Follow-up to #1049.

Very similar to #1049, except this handles the case where the given GameScope arguments string that we want to break up into `GAMESCOPEARGSARR` (which ultimately gets passed to the launch command) does **not** end with `--`. This causes the same problem as if the arguments are blank: the GameScope command is not properly terminated.

ShellCheck is good on this at time of writing but I want to do some more testing (such as with whitespaces and such after the args string to see how we handle that).

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -6,7 +6,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240307-1"
+PROGVERS="v14.0.20240307-2 (gamescope-args-incorrect-ending-fix)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Tested and this works with the following GameScope arg strings:
- `-w 1280 -h 720 -W 1280 -H 720 -f --` - valid args with `--` at the end as it should normally be by default, and also when edited from the GameScope GUI.
- `-w 1280 -h 720 -W 1280 -H 720 -f` - valid args, but missing `--` at the end, so invalid syntax. This kind of case would only arise if the args were manually entered/edited. This PR catches this case and appends `--` at the end.
- `` - Empty string, correctly handled.
- `none` - This is what actually gets saved to `GAMESCOPE_ARGS` config var if the args are blank after saving, but this was tested with the above by going to the Game Menu -> Deleting all GameScope args so the textbox was blank -> Save and Play (success) -> Close and relaunch game -> Go to Game Menu -> GameScope args were `none` -> Save and Play (success).

I tested the arguments with a ton of whitespaces and they do break, but I think it's up to the user to make sure they format the args correctly. Appending `--` is just a convenience, trying to handle formatting the spacing between flags would add a lot of complexity especially when it comes to paths. So for now, we will leave this.

<hr>

Now that testing is done, this is ready to merge after a version bump.

PR #1054, opened 2024-03-14
feat(dep): check integrity

### LINKED ISSUE #859: Steam Deck Experience Improvements [Community Contributions Wanted]

Mte90:
Based on https://github.com/sonic2kk/steamtinkerlaunch/issues/859

This check if innoextract and cabextract are fine.

Check the size is not something that can confirm as you suggested in the ticket, the only one is to run the command and see if crashes.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -25790,9 +25790,10 @@ function checkSteamDeckDependencies {
 		local DEPFILENAME="$2"
 		local DEPDIR="$3"
 		local REPOURL="$4"
-		local CHECKCMD="$($DEPCMD &> /dev/null --version && echo "OK" || echo "NOK")"
+		local CHECKCMD
+		CHECKCMD="$($DEPCMD &> /dev/null --version && echo "OK" || echo "NOK")"
 
-		if [ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" -eq "OK" ]; then
+		if [ -f "$(command -v "$DEPCMD")" ] && [ "$CHECKCMD" = "OK" ]; then
 			writelog "INFO" "${FUNCNAME[0]} - Using '$DEPCMD' binary found in path: '$(command -v "$DEPCMD")'"
 			echo "Dependency '$DEPCMD' already installed, nothing to do."
 		else
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -25790,8 +25790,9 @@ function checkSteamDeckDependencies {
 		local DEPFILENAME="$2"
 		local DEPDIR="$3"
 		local REPOURL="$4"
+		local CHECKCMD="$($DEPCMD &> /dev/null --version && echo "OK" || echo "NOK")"
 
-		if [ -f "$(command -v "$DEPCMD")" ]; then
+		if [ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" -eq "OK" ]; then
 			writelog "INFO" "${FUNCNAME[0]} - Using '$DEPCMD' binary found in path: '$(command -v "$DEPCMD")'"
 			echo "Dependency '$DEPCMD' already installed, nothing to do."
 		else
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240317-1"
+PROGVERS="v14.0.20240318-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"


sonic2kk:
This looks pretty great, thanks! I like this approach a lot, I think it makes sense. I had some comments but ShellCheck summarises them better, I'll leave them as comments on the specific lines themselves.

I realise you had issues with the Bash LSP so it's understandable. As a workaround you can instead download ShellCheck 0.8.0 manually and add it to your path as an alias, that's what I do for ShellCheck 0.8.0. Then from the project folder, you can run `shellcheck steamtinkerlaunch` to check the script. This will hopefully be unnecessary once a newer ShellCheck version releases, fixing the issue for various scripts.

Apart from that, I was wondering if we should make it a dedicated function, in case we can use this elsewhere, but I don't think we really use many native binaries this way. If needs be in future, we can split it out.

Thanks for your contribution! Once the ShellCheck comments are addressed, I'll merge this :-)

steamtinkerlaunch
```diff
@@ -25776,8 +25776,9 @@ function checkSteamDeckDependencies {
 		local DEPFILENAME="$2"
 		local DEPDIR="$3"
 		local REPOURL="$4"
+		local CHECKCMD="$($DEPCMD &> /dev/null --version && echo "OK" || echo "NOK")"
```
sonic2kk: ShellCheck gives this warning:

> `SC2155 (warning): Declare and assign separately to avoid masking return values.`

ShellCheck gives a warning if you try to do this because it can potentially miss error codes, see more on the [SC2155 wiki page](https://www.shellcheck.net/wiki/SC2155).

I believe I ran into this a while back as well (honestly not sure why I went with locals for the Steam Deck stuff...) and just defined it separately. Something like this will work (even if it's a bit ugly):

```
local CHECKCMD
CHECKCMD="$(...)"  # replacing '...' with the actual command, although see below comment for ShellCheck complaints)
```

One more universal solution. might be to break this out into a separate function (although note the ShellCheck syntax comments. But if you want to ignore this and justify why, we can add a ShellCheck ignore in a worst case (there are a couple of cases of this around the code, but I keep it to a minimum).

steamtinkerlaunch
```diff
@@ -25776,8 +25776,9 @@ function checkSteamDeckDependencies {
...
 
-		if [ -f "$(command -v "$DEPCMD")" ]; then
+		if [ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" -eq "OK" ]; then
```
sonic2kk: The `if` check isn't valid, it needs to be separated into two `[ ]` conditions. Also, you should use `=` instead of `-eq` when comparing strings, this `-eq` does actually flag up with `integer expression expected`.

ShellCheck is flagging them up as follows:

> `SC2107 (error): Instead of [ a && b ], use [ a ] && [ b ]`
> `SC2170 (error): Invalid number for -eq. Use = to compare as string (or use $var to expand as a variable).`

Both of these should be fixable with the following (splits the checks into two `[ ]` tests, and uses `=` for string comparison:

```
if [ -f "$( command -v "$DEPCMD" )" ] && [ "$CHECKCMD" = "OK" ]; then
```

Mte90:
Yeah I am using Scite that doesn't use LSP and doesn't lag with huge files (I use that for mysql dump that often are bigger than this).

sonic2kk:
ShellCheck v0.10.0 is out, and once I add the directive to not use extended analysis, as long as you're using an up-to-date LSP things should be fine.

sonic2kk:
Sorry, I just realised I forgot to mention (and document) that the version should be bumped. There's no exact rule on this, basically just the date for you + 1, and then bump the revision number. So for example you might set `v14.0.20240316-1`.

This isn't enforced, this is just my own workflow that you might find useful when working on PRs: typically what I do is set the version during development, and in brackets, add the branch name. This means if another PR uses the same date, the version string is still unique. So in this case it might be `v14.0.20240316-1 (valid-dependency)`. Then, once the PR is ready to be merged, I change the version. If you submit a large PR it might be good to do this, and once all feedback is addressed and ready to merge, I can leave a comment to bump the version and then it'll be ready to merge. :slightly_smiling_face: 

This is used to know when and how to update config files, and they are assigned a version. It's not massively important for afaik any of your PRs but still just good practice (I forget to do it sometimes myself!)

Once this is done, the rest of the changes look good to me, and this will be ready to merge. ShellCheck v0.10.0 is very angry about a lot of this script, but v.0.8.0 gives no warnings after the changes. At a glance though I think v0.10.0 isn't complaining about anything specific to this PR.

<hr>

Thanks a whole lot! :smile:

Mte90:
As you asked version bumped.

sonic2kk:
Thanks! Although the string isn't formatted quite right.

Your version string is `v14.0.20240316-(valid-dependency)`.
- For development, it can be `v14.0.20240316-1 (valid-dependency)` (note the `-1` and the space between `(valid-dependency)`.
- Before merging, it should be `v14.0.20240318-1` (or `-2` and so on, depending on the version string on master).

This PR has conflicts because of my mergings, but I'll see if I can rebase this PR myself and merge it if that's ok. :smile: You can also use the diff here as reference for future PRs.

<hr>

Thanks!

Mte90:
Fixed the version and for conflicts, yeah it is better that you do so :-)

sonic2kk:
Heh, I took a quick look at how to contribute to someone else's PR using [this gist](https://gist.github.com/wtbarnes/56b942641d314522094d312bbaf33a81). I will try my best not to mess anything up on you, but if I do, I can probably just pick your PR into another branch and then squash-merge it (you _might_ end up as a co-author), but you'll be credited correctly in the changelog :smile:)

Mte90:
To me is perfectly fine :-)

sonic2kk:
I got it in the end I think. Just doing one last ShellCheck run and then I will merge this.

Thank you for your patience, continued discussions and contributions! It is much appreciated.

sonic2kk (approved):

sonic2kk:
Changelog has also been updated to include your contribution (might be easiest to see with Ctrl+F for `@Mte90`): https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog - This changelog gets copied and pasted into the release notes, so you will be credited in the same way there too.

Also, it looks like despite the commits in the PR being co-authored (because of the rebase), the squash-merge properly only included you, which I am happy about!

I am very keen on making sure all contributors get properly recognised, which is also why I bring it up here too. Thanks again for all your work!

PR #1055, opened 2024-03-15
feat(steamdeck): custom variable to detect

Mte90:
ref #859

Right now just the variable as it isn't for me understand when needed to be used.


sonic2kk:
I think we already have this, it's called `ONSTEAMDECK`:
- Defined here: https://github.com/sonic2kk/steamtinkerlaunch/blob/8bcc2af6492c8feccea3ce249f62829e7ff8ece7/steamtinkerlaunch#L19
- Set here in `steamdedeckt`: https://github.com/sonic2kk/steamtinkerlaunch/blob/8bcc2af6492c8feccea3ce249f62829e7ff8ece7/steamtinkerlaunch#L25911-L25913

Is there a way this is different?

Mte90:
No it is the same thing but the ticket about steam deck mentioned that it was needed to check if the device was steam deck. Check all the code it isn't easy so I didn't noticed.

sonic2kk:
I don't see this being mentioned as something that needs to be implemented. Did I make a mistake somewhere? We already use this variable in a lot of places to do conditional Steam Deck things, like disable various options that don't make sense in Game Mode (such as GameScope, MangoHud, GameMode, etc) and when performing the Steam Deck installation.

This wasn't implemented as part of #859 either, it has been around since Steam Deck skeleton code was added (commit from probably around this time 2 years ago, so I couldn't find it :sweat_smile:), so if I gave the impression it has to be added, very sorry and I'll happily correct it!

Mte90:
So #859 mention to do some stuff about Steam deck but doesn't mention that there is a variable for that so I thought that a detection was needed.

Anyway it isn't a problem, I have VM now for steam deck :-)

PR #1059, opened 2024-03-15
feat(detect): if steam and yad not found stop the program

### LINKED ISSUE #1058: Check if steam is installed
So right now executing STL without steam you get this:

```
ven 15 mar 2024, 11:56:37, CET WARN - setSteamPaths - Steam 'userdata' directory not found, other variables depend on it - Expect problems
grep: : File o directory non esistente
grep: : File o directory non esistente
ven 15 mar 2024, 11:56:38, CET ERROR - checkIntDeps - '' was not found! Check 'steamtinkerlaunch --help' for alternatives and/or read 'https://github.com/sonic2kk/steamtinkerlaunch/wiki/Yad'
ven 15 mar 2024, 11:56:38, CET ERROR - checkIntDeps - Yad version '0' is too old. You need to update to at least '7.2'
```

So I think that the tool should check if the steam binary exists, also in my case I don't have yad but is still reporting a version.

Mte90:
Fix #1058

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240307-3-(steam-check)"
+PROGVERS="v14.0.20240320-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -619,8 +619,8 @@ function setSteamPath {
 }
 
 function setSteamPaths {
-	if [ ! -x "$(command -v "steam")" ]; then
-		echo "ERROR" "${FUNCNAME[0]} - Steam not found"
+	if [ ! -x "$(command -v "steam")" ] && [ "$INFLATPAK" -eq 0 ]; then
+		echo "ERROR" "Steam not found"
 		writelog "ERROR" "${FUNCNAME[0]} - Steam not found"
 		exit 0
 	fi
@@ -18262,7 +18262,7 @@ function checkIntDeps {
 		fi
 		
 		if [ ! "$YAD" ]; then
-			writelog "ERROR" "${FUNCNAME[0]} - Could not find yad" "E"
+			writelog "ERROR" "${FUNCNAME[0]} - Could not find yad"
 			exit 0
 		fi
 
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240319-1"
+PROGVERS="v14.0.20240307-3-(steam-check)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -618,13 +618,7 @@ function setSteamPath {
 	fi
 }
 
-function setSteamPaths {
-	if [ ! -x "$(command -v "steam")" ] && [ "$INFLATPAK" -eq 0 ]; then
-		echo "ERROR" "Steam not found"
-		writelog "ERROR" "${FUNCNAME[0]} - Steam not found"
-		exit 0
-	fi
-	
+function setSteamPaths {	
 	if [ -f "$STPAVARS" ] && grep -q "^SUSDA" "$STPAVARS" ; then
 		writelog "INFO" "${FUNCNAME[0]} - Reading Steam Path variables from '$STPAVARS'"
 		loadCfg "$STPAVARS" X
@@ -18260,11 +18254,6 @@ function checkIntDeps {
 		if [ -z "$YAD" ]; then
 			YAD="$(command -v "yad")"
 		fi
-		
-		if [ ! "$YAD" ]; then
-			writelog "ERROR" "${FUNCNAME[0]} - Could not find yad"
-			exit 0
-		fi
 
 		if [ -n "$YAD" ] && [ ! -f "$YAD" ]; then
 			OYAD="$YAD"
@@ -18282,8 +18271,15 @@ function checkIntDeps {
 
 		if [ ! -x "$(command -v "$YAD")" ]; then
 			DEPSMISSING=1
-			writelog "ERROR" "${FUNCNAME[0]} - '$YAD' was not found! Check '${PROGCMD} --help' for alternatives and/or read '$PROJECTPAGE/wiki/Yad'" "E"
+			writelog "ERROR" "${FUNCNAME[0]} - Yad dependency ('$YAD') was not found! Check '${PROGCMD} --help' for alternatives and/or read '$PROJECTPAGE/wiki/Yad'" "E"
 			notiShow "$(strFix "$NOTY_NOTFOUND" "$YAD")"
+		fi
+		
+		
+		if [ ! -x "$(command -v "steam")" ] && [ "$INFLATPAK" -eq 0 ]; then
+			DEPSMISSING=1
+			echo "ERROR" "Steam not found"
+			writelog "ERROR" "${FUNCNAME[0]} - Steam not found"
 		fi
 
 		setAwkBin
@@ -24518,7 +24514,7 @@ function addNonSteamGame {
 }
 
 function setCloseVars {
-# yes, ugly... ¯\_(ツ)_/¯
+# yes, ugly... Â¯\_(ã)_/Â¯
 	if [ ! -f "$CLOSEVARS" ]; then
 		writelog "INFO" "${FUNCNAME[0]} - Storing all variables used in closeSTL into '$CLOSEVARS'"
 		{
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -618,7 +618,7 @@ function setSteamPath {
 	fi
 }
 
-function setSteamPaths {	
+function setSteamPaths {
 	if [ -f "$STPAVARS" ] && grep -q "^SUSDA" "$STPAVARS" ; then
 		writelog "INFO" "${FUNCNAME[0]} - Reading Steam Path variables from '$STPAVARS'"
 		loadCfg "$STPAVARS" X
@@ -24514,7 +24514,7 @@ function addNonSteamGame {
 }
 
 function setCloseVars {
-# yes, ugly... ¯¯\_(ツ)_/¯¯
+# yes, ugly... ¯\_(ツ)_/¯
 	if [ ! -f "$CLOSEVARS" ]; then
 		writelog "INFO" "${FUNCNAME[0]} - Storing all variables used in closeSTL into '$CLOSEVARS'"
 		{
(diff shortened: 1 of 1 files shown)


sonic2kk:
Thanks!

I don't think we should exit if Steam is missing, I think we should just more explicitly warn.

In its current state as well, I have a few concerns:
- If we're going to check for Steam this way, we need to also make sure we're not on Flatpak.
- The `echo` isn't correct for Steam being missing, it should just say `echo "ERROR: Steam not found`
- Exiting if we can't find Yad is a bit excessive I think, since there is perfectly valid commandline usage. We already log if Yad is missing, but I think we can improve it to more clearly state that a Yad binary is missing in `checkIntDeps`.
    - The logging you showed in #1058 isn't great.
    - Also, users should already know to install Yad based on reading the installation documentation.
    - If the commandline usage *on Desktop* still complains about a missing Yad, that ideally should be fixed, but that's a larger effort for something that I think can be fixed by making the logs clearer.

I'll leave some of these as review comments where appropriate for easier tracking :-)

steamtinkerlaunch
```diff
@@ -618,6 +618,12 @@ function setSteamPath {
 }
 
 function setSteamPaths {
+	if [ ! -x "$(command -v "steam")" ]; then
```
sonic2kk:
```suggestion
	if [ ! -x "$(command -v "steam")" ] && [ "$INFLATPAK" -eq 0 ]; then
```

Ensures we don't fail with missing Steam when running in Flatpak, where `steam` may still work, but I'd rather not do this kind of check (Flatpak STL can't really be used outside of Flatpak Steam to begin with, so we wouldn't need to check for Steam there anyway).

steamtinkerlaunch
```diff
@@ -618,6 +618,12 @@ function setSteamPath {
...
+		echo "ERROR" "${FUNCNAME[0]} - Steam not found"
```
sonic2kk:
```suggestion
		echo "ERROR - Steam not found"
```

steamtinkerlaunch
```diff
@@ -18243,6 +18249,11 @@ function checkIntDeps {
 		if [ -z "$YAD" ]; then
 			YAD="$(command -v "yad")"
 		fi
+		
+		if [ ! "$YAD" ]; then
```
sonic2kk: Instead of doing this check, we can probably just improve the missing Yad logging in `checkIntDeps`.

steamtinkerlaunch
```diff
@@ -18243,6 +18249,11 @@ function checkIntDeps {
...
+			writelog "ERROR" "${FUNCNAME[0]} - Could not find yad" "E"
```
sonic2kk: I know we're probably going to remove this but just for future note, `writelog` usage is like this:

```suggestion
			writelog "ERROR" "${FUNCNAME[0]} - Could not find yad"
```

sonic2kk (changes requested):

sonic2kk:
This is my first time doing the "request changes" thing on GitHub so let me know if I did anything wrong :-)

I ran ShellCheck v0.10.0 against your branch, there were general warnings not present in v0.8.0, and none specific to your PR, so no worries there.

steamtinkerlaunch
Mte90: The problem as you saw if yad is not found I get also some notification and fails the version detection.
So maybe we can improve to just exclude the CLI case when a parameter is used.

Mte90:
You did in the right way but I missed those notifications the other day.

Anyway I think that check if it is running in CLI mode for yad detection can be a way.

steamtinkerlaunch
sonic2kk: > So maybe we can improve to just exclude the CLI case when a parameter is used.

I don't know if this will work as SteamTinkerLaunch will call itself from the UI too, for example it might called `mo2 start` from the UI.

I think some loggings and (I assume harmless?) errors are fine. Although Yad really *should* be installed to use SteamTinkerLaunch fully, since a lot of it still uses Yad. For example, the `ansg` may use Yad if called with some or none of the required parameters, but will not if all the parameters are supplied.

I don't know that quitting if Yad is not installed is necessary though, same with the other dependencies (except Steam I guess). We can just let STL fall over if dependencies are missing.

sonic2kk:
No problem, easy to miss :-) Before I learned how to do this properly, I left comments in review without actually submitting them for a day or two :smile:

steamtinkerlaunch
Mte90: I think that the only thing that you need to decide is: Yad is mandatory?

In this way also for pure cli it should fail if not found it. Maybe as it is now it is better to do that, maybe when the code will be reorganized and this kind of check about UI will be improved we can say that yad is not mandatory on CLI.

sonic2kk: I think for now, Yad being mandatory is the path of least resistence. I don't know if the UI will ever be reorganised in the way to avoid Yad being a dependency, since STL calling itself is by design, but for sure this is something to look at in the future.

Yad is already mentioned as a [hard dependency](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Installation#hard-dependencies) on the wiki.

Mte90: Well it is an hard dependency so we decided, you can just approve the PR if there isn't anything else to do.

sonic2kk: I think we can remove the Yad check in this way. We already have logging for a missing Yad and I don't think we need to exit on Yad being missing :-)

Mte90: In my opinion we should exit if not found other wise you can get errors in console or notifications that makes no sense.
In this way the tool doesn't keep following do the rest of the program.

sonic2kk: Reading through `checkIntDeps` agaiin, I think your check is still unnecessary.
1. If Yad is missing, we catch it with `DEPSMISSING=1` in the `if [ ! -x "$(command -v "$YAD")" ]; then` block.
2. At the bottom of `checkIntDeps`, we will exit if any dependencies are missing using `if [ "$DEPSMISSING" -eq 1 ]; then`.
3. We will continue to correctly skip this check if `SKIPINTDEPCHECK -eq 1`.

So actually, perhaps your Steam check should also go here, and if Steam is missing, set `DEPSMISSING=1`. So you can copy your Steam check from `setSteamPaths` - that is, everything in `if [ ! -x "$(command -v "steam")" ] && [ "$INFLATPAK" -eq 0 ]; then` - and move it to `checkIntDeps`, inside of the `else` block. Then, instead of `exit 0`, you can set `DEPSMISSING=1`, and then we'll exit at the end of the function.

In doing this, we get a full log if an execution fails, that should list all missing dependencies.

The wiki should also be updated to note Steam as a hard dependency after this PR.

sonic2kk: I forgot to address your other point, the error message for Yad currently indeed does not make a whole lot of sense. I'll leave a review comment in case you want to tackle that here too, I think the fix is to replace `'$YAD'` with `Yad` (since `$YAD` is just an empty string if we get this far)

sonic2kk:
Apparently I cannot leave a comment on it, but if you want to fix up the Yad error messaging, you can change this line:

```bash
writelog "ERROR" "${FUNCNAME[0]} - '$YAD' was not found! Check '${PROGCMD} --help' for alternatives and/or read '$PROJECTPAGE/wiki/Yad'" "E"
```

To

```bash
writelog "ERROR" "${FUNCNAME[0]} - Yad dependency ('$YAD') was not found! Check '${PROGCMD} --help' for alternatives and/or read '$PROJECTPAGE/wiki/Yad'" "E"
````

There's no obligation to, but as you pointed out, the error doesn't make much sense, so this would be a nice win to include in this PR :-)

sonic2kk:
Sorry for the triple-commenting, but just wanted to mention: This PR doesn't have any conflicts, and I don't think it will as I'm hoping to merge it soon, but if it does end up with conflicts (or any if your future PRs) and you want assistance rebasing, feel free to ask :-) I'm not an expert but I'll do my best to help out anyone generous enough to help out.

Rebasing is still kinda magic to me, and I don't think we'll be in a situation with a large difficult-to-rebase PR (if there was ever a large PR to come in, I would prioritise merging it first over my own commits to make life easier for contributors).

sonic2kk:
Sorry for the labelling noise, was testing out the ShellCheck CI (adjusted it a little with #1068).

I expected it to fail as it wasn't rebased against master, but it didn't fail, which was a bit odd... Locally it's eating up tons of ram still, and the fixes aren't in place from #1064. I have no idea why the workflow is passing, because it should be failing for this PR...

EDIT: The Action *is* working, see #1069. Not sure why it's passing here but I'll take it I suppose.

Mte90:
So there is a conflict as I can see, this happens with huge files but we already talked about this.

Anyway I did the changes you suggested.

sonic2kk:
> this happens with huge files

No, conflicts happen regardless, it is nothing to do with the size of the file (lines, not bytes ofc :wink:). This is simply the nature of version control, happens often in other projects I contribute to (namely ProtonUp-Qt), happens in work all the time as well.

I haven't checked yet, but the issue here is almost definitely a result of needing rebased from the version bump in other commits (most likely a7e06e320cf5d1c5d79ffabb0299c0104a34860a), this is typically what happens when one feature is merged before another. Normally each PR needs a rebase after a feature merge. I'm not sure why the GitHub Web Editor says the conflicts are "too complex" but this happens often now, and the Web Editor is not great to use to resolve conflicts anyway.

But either way, I'm happy to do the rebasing. If you haven't done a rebase before, it is basically lifting all of the commits and moving them on top of the most recent commit, as if you were applying your changes to begin with at the most recent commit.

For cases like this it is normally just a case of:
1. Make sure your master branch is synced with this master branch (you can typically do this from GitHub on your fork's master branch page, it'll fast forward the commits - There is a command, but I forget it now)
2. Pull these changes into your local master, so it is up-to-date also, i.e. `git checkout master && git fetch && git pull master && git checkout steam-check`
3. Rebase with `git rebase master`
4. Resolve the conflicts, normally this is just the version bump. Some editors have UIs for resolving conflicts, but usually you can just accept your current version. After this you can `git add steamtinkerlaunch` (or any other files that have conflicts) and force-push the rebased commits (rebasing rewrites history since it "moves" commits and the hash changes).

You can read more about rebasing on [the Git docs](https://git-scm.com/book/en/v2/Git-Branching-Rebasing), since it can do a bit more than what is needed here (for example, squashing certain commits together).

But I'm always happy to help out with rebasing or other related things with PRs :-)

sonic2kk:
Rebase done, it was indeed just the version that needed a conflict resolved. I rebased keeping the version the same, and then pushed a commit that bumped the version to be ready for merging.

Running ShellCheck locally and on CI, and once verified, I will merge (the CI is so freaking cool by the way, I love it :smile:)

sonic2kk (approved):

sonic2kk:
Updated the wiki to note that Steam is required: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Installation/_compare/1355066310e31cc7b00d872d2c3f67ce8e79b54d

~~Not sure if it's worth updating the note on ProtonUp-Qt. I think I'll open a discussion over there and ask.~~ Nevermind, after DavidoTek/ProtonUp-Qt#356, it shouldn't be possible for it to install any Steam compatibility tool without a valid installation; in other words, Steam will always be available when installing SteamTinkerLaunch from ProtonUp-Qt, since the option to install any Steam compatibility tool won't appear if there is no Steam installation to begin with.

sonic2kk:
I forgot to update the changelog to note this contribution until just now! I'm sorry about that, the changelog is now updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog/_compare/95260902f974ece48d77509e0362eab14b207832

PR #1061, opened 2024-03-15
ShellCheck: Set directive 'extended-analysis=false'

sonic2kk:
*(See also: koalaman/shellcheck#2652)*
Prerequisite to #1056.

New directive available in ShellCheck v0.10.0. Fixes a memory issue with the DFA introduced in ShellCheck v0.9.0, which is why STL was stuck using v0.8.0. Now we can use newer ShellCheck version again!

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -1,12 +1,13 @@
 #!/usr/bin/env bash
 # shellcheck source=/dev/null
+# shellcheck extended-analysis=false
 
 ### project ###
 
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240307-3"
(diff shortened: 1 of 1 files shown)


sonic2kk:
ShellCheck v0.10.0 is very upset about a *lot* of things in SteamTinkerLaunch, but I'll open a separate PR to fix those.

We will also need to update `CONTRIBUTING.md` to note that ShellCheck v0.8.0 is no longer the maximum version. It could do with some general updates though, so I'll probably handle all the updates in the same PR.

sonic2kk:
Tested a few times, ShellCheck completes a little faster (granted, there are quite a few warnings) but it won't eat up tons of RAM now. ShellCheck v0.10.0 won't work without this line, but works fine with it. I tested without the directive and it went up to 24gb ram used, I killed the process at that point.

This is ready to merge I think.

PR #1062, opened 2024-03-15
Update CONTRIBUTING.md with More Up-To-Date Contributing Information

sonic2kk:
Updates some outdated information and clarifies some things I realised weren't really clearly documented when helping out a new contributor.

* Update PR commit workflow, no longer care about squashed commits
* Update ShellCheck note now that v0.10.0 does work as long as extended-analysis is disabled (off in the script with a directive by default)
* Add information about when and how to adjust script PROGVERS


sonic2kk:
Clarified a couple things in 9903708. Think this is good to merge now.

PR #1065, opened 2024-03-16
Misc: Warn that WINDECO option does nothing on Wayland

sonic2kk:
Yad allows you to specify a `--decorated` or `--undecorated` option for dialogs, which we control with `WINDECO`. However this flag has no effect on Wayland. This may be because Mutter, GNOME's Wayland compositor (Yad is a GTK3 tool), does not implement the [`xdg-decoration` protocol](https://wayland.app/protocols/xdg-decoration-unstable-v1#compositor-support) at time of writing. KDE does, and window decorations can be toggled with a KWin rule and with a keyboard shortcut to "Toggle Window Titlebar and Frame" (it also disables shadows, like it does on X11). This seems to be a rabbit hole debate around server-side vs client-side decorations!

This PR adds a warning for this and updates the langfile strings to note this, as I don't see this being resolved anytime soon.


sonic2kk:
Apparently, toggling window decorations requires a restart, because we only set it in `createDefaultCfgs` when we call `prepareGUI`. We could probably fix this to be honest, but for now, we'll note it requires a restart as well.

PR #1069, opened 2024-03-18
Purposely break to test ShellCheck CI

sonic2kk:
Testing ShellCheck CI with a purposely broken string. This should produce an [SC2086](https://www.shellcheck.net/wiki/SC2086) warning.


sonic2kk:
Huh, it did actually fail!

PR #1074, opened 2024-03-23
Menus: Display paths with backslashes properly

### LINKED ISSUE #1072: Unable to pass windows paths to custom commands

sonic2kk:
Partial fix for #1072.
See also: #1073.

## Overview
On various places on the UI, we may display paths with backslashes. This PR is concerned with those paths coming from the Steam Game Launch Options, and the Custom Command arguments, and fixes them by using `printf` in various ways.

## Problem
The problems are as follows:
- **Game Menu**
    - Entering a path with backslashes in a custom command argument, such as `Z:\this\is\my\path`, and then saving, will strip this path of its backslashes, turning it into `Z:\thisismypath`.
    - This can be worked around by double escaping, but the path would then turn out like this: `Z:\\this\\is\\my\\path` -> Save -> Path becomes `Z:\this\is\my\path` -> Save again -> `Z:thisismypath` - This is because the path is getting stripped of backslashes from the escaping on each save
    - This means each time the Game Menu is opened, paths with backslashes need to be quadruple-escaped before saving.
- **Main Menu**
    - Paths coming from Steam with backslashes (with and without quotes) will not display their backslashes properly.
    - Ex: This launch option `Z:\this\is\my\path` will be displayed on the SteamTinkerLaunch Main Menu as `Z:thisismypath`.

## Solution
This PR fixes these problems in two different ways, both using `printf`:
- For the Game Menu, we can use `printf "%s"` in a subshell when displaying the field for `CUSTOMCMD_ARGS`. This, for some reason I don't fully understand, fixes displaying and thus saving the paths with backslashes correctly. They are no longer mangled.
- For the Main Menu, we can use `printf "%q"`, which is an option that preserves strings to be used as part of a command. This handles our double-escaping for display purposes
    - We cannot use `%q` for the Game Menu because then the double-escapes (and the escape around any other parts of the launch command) would be displayed, which would also result in the commands not launching correctly.

Note that paths going as arguments to custom commands may need to be double-escaped, as they don't have quotes. This PR corrects the display and saving of these paths, and #1073 fixes the paths so that they are double-escaped.

Fwiw, Proton doesn't seem to get quotes for launch arguments either though (https://github.com/sonic2kk/steamtinkerlaunch/issues/1072#issuecomment-2016276706) so I think this is acceptable.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240322-1 (customcmd-args-backslash-fix)"
+PROGVERS="v14.0.20240323-1 (customcmd-args-backslash-fix)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Uh oh, we're running the CI twice... We can probably remove the PR one...

sonic2kk:
Tested by me and @zany130 in #1072, appears to work as expected, so merging!

PR #1076, opened 2024-03-24
updateConfigEntry: Pass `sed` an escaped `CFGVALUE`

### LINKED ISSUE #1072: Unable to pass windows paths to custom commands

sonic2kk:
Work for #1072.

`sed` will expand escape sequences in strings, so give it an escaped version of the `CFGVALUE` string to make updating existing config value items with backslashes work without expanding them. For example, `sed` would escape the following string, removing all backslashes and expanding the `\t` to be a tab character: `Z:\home\gaben\testgames`.

Should fix updating blank config entry items to have new values with backslashes in them. For example, updating a blank `CUSTOMCMD_ARGS` to have `CUSTOMCMD_ARGS="Z:\home\gaben\testgames"`.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240324-1 (updateConfigEntry-escpae-incoming-cfgvalue)"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Tested by me and by @Zany130 in #1072, works as expected and does not appear to introduce any regressions to saving with existing config files as well as with fresh config files. This should be ready to merge.

PR #1077, opened 2024-03-24
Misc: Bump Yad AppImage to use new release

### LINKED ISSUE #1060: Bump Yad to a new release
## Feature Description

The repository used it as fallback, if on the machine yad is not installed, is on readonly since latest december https://github.com/frostworx/steamtinkerlaunch-tweaks/releases
The latest release is from 2021 so is time to find a more viable solution to get yad as fallback.

On https://github.com/v1cont/yad/issues/177 no one proposed another binary and steamtinkerlaunch is an appimage so need that all the dependence are there.

I am getting yad compiled statically without appimage with all the dependencies, in my debian sid and the same executable on a steamos instance without gtk installed.

* Downloading https://github.com/v1cont/yad
* `autoreconf -ivf && intltoolize`
* `./configure --enable-standalone && make`
* Downloading https://github.com/oufm/packelf
* `./packelf.sh ../yad/build/src/yad ./yad`

I get a 90~ mb without webgtk and gtkspell support.

image

I think that just a binary to download in case is not available is more simple then a appimage.

I am thinking to do a CI, that do the binary and the release etc in a personal repository.
In this way it is more easy to maintain and merge this CI to the original yad project.

sonic2kk:
Implements #1060.
Related #859.

Bump Yad AppImage to use new release over at: https://github.com/sonic2kk/steamtinkerlaunch-tweaks/releases/tag/Yad-fbfb0fa-x86_64.AppImage

We pin to the latest AppImage release on GitHub by assuming the release name and AppImage will have the same string. The new release should follow this convention. This means existing code and older releases will still use the old AppImage, but newer versions will download the new AppImage.

This new release updates the Yad version. As always on SteamOS, dependencies will have to be cleared to test this (the `steamtinkerlaunch cleardeckdeps` command).

This PR needs significant amounts of testing before it can be merged. Once merged, the AppImage release will no longer be marked as a draft.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240526-3"
+PROGVERS="v14.0.20240324-1 (bump-yad-appimage)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -18197,7 +18197,8 @@ function setYadBin {
 		fi
 
 	elif [ "$1" == "ai" ] || [ "$1" == "appimage" ]; then
-		YADSTLIMAGE="Yad-8418e37-x86_64.AppImage"
+		# YADSTLIMAGE="Yad-8418e37-x86_64.AppImage"
+		YADSTLIMAGE="Yad-fbfb0fa-x86_64.AppImage"
 		YAIDL="$YAIURL/$YADSTLIMAGE/$YADSTLIMAGE"
 		YADAPPIMAGE="$YADSTLIMAGE"
 		DLCHK="sha512sum"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -18198,7 +18198,7 @@ function setYadBin {
 
 	elif [ "$1" == "ai" ] || [ "$1" == "appimage" ]; then
 		# YADSTLIMAGE="Yad-8418e37-x86_64.AppImage"
-		YADSTLIMAGE="Yad-fbfb0fa-x86_64.AppImage"
+		YADSTLIMAGE="Yad-13.0-x86_64.AppImage"
 		YAIDL="$YAIURL/$YADSTLIMAGE/$YADSTLIMAGE"
 		YADAPPIMAGE="$YADSTLIMAGE"
 		DLCHK="sha512sum"
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240324-1 (bump-yad-appimage)"
+PROGVERS="v14.0.20240325-3 (bump-yad-appimage)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -18223,7 +18223,6 @@ function setYadBin {
 				export ONSTEAMDECK=1
 				YADAIDLDIR="$STLSDPATH"
 				YAIDST="$YADAIDLDIR/$YADAPPIMAGE"
-				YAIURL="$GHURL/sonic2kk/steamtinkerlaunch-tweaks/releases/download"  # TODO isn't this already defined in URLs with the same namd and path?
 				YAIDL="$YAIURL/$YADSTLIMAGE/$YADSTLIMAGE"
 				YADAPPIMAGE="$YADSTLIMAGE"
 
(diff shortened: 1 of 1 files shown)


sonic2kk:
We can probably remove the duplicate `YAIURL` in `setYadBin` in this PR as well. There was a TODO left here about this since #1063, as it appears to be identical to the one found in `uris.conf`.

sonic2kk:
The AppImage appears to work fine on my Linux PC. I don't have SteamTinkerLaunch on my Steam Deck, but I should be able to download and test the Yad AppImage standalone there sometime next week.

Ideally before this gets merged, someone will test it out on SteamOS.

Mte90:
I am trying on SteamOS but doesn't happen anything.
In console I am getting nothing so I can't help..

immagine

sonic2kk:
Are you sure the AppImage downloaded correctly? It doesn't look like it did based on your terminal output. Double-check that the filesize is ~80mb.

I was able to run it on my Steam Deck and it errored out around the glibc version. I'm guessing there's some glibc incompatibility between my Arch PC and the Steam Deck.

<hr>

SteamOS `glibc 2.38-3.2`, that explains why it doesn't work as the AppImage on my PC was built against a newer glibc version.

Sigh, I'll probably have to get a Debian VM or something to build against, assuming it has an older glibc version. Alternatively, this could wait until SteamOS updates to a newer glibc version. Hopefully with the next snapshot update this happens.

Mte90:
You were right the file downloaded with curl was 0 bytes...
I got the same error about glibc 2.38 and 2.34 not found.

Maybe we can compile it in a VM for Steamos or my idea of a CI will simplify this situation.

sonic2kk:
Once again, I don't think CI is necessary. This also isn't an urgent change, Yad is not updated frequently.

If needs be we can manage building it on an Ubuntu VM once every two years :-)

sonic2kk:
Updated AppImage, needs testing. @AtomHare feel free to check it out on SteamOS if you like :-)

AtomHare:
I just tried it and it worked well ! Both in desktop and in gamemode !

However I uncountered an issue, I did a `cleardeckdeps` to download the latest `Yad`, however when "downloading" no progress bar appear and after looking in the deps folder the yad appimage was of 0 Bytes, thus steamtinkerlaunch didn't work as expected.
After replacing manually the appimage everything works as expected though.

sonic2kk:
Network errors can occur and cause this. I'm not sure we can do much about this without some logic to retry downloads, but failure to download/connect to the internet is quite common on SteamOS (iirc it has a lot to do with the Steam Client downloading in the background hogging the connection and causing timeouts).

Thanks for testing though! This can be merged then I think.

AtomHare:
The download didn't fail because of a "steam deck unreliability", it's completely unrelated indeed, it looks like the crafted URL by STL is : `YAIDL="$YAIURL/$YADSTLIMAGE/$YADSTLIMAGE"` which worked before because the artifact was at `https://github.com/sonic2kk/steamtinkerlaunch-tweaks/releases/download/Yad-8418e37-x86_64.AppImage/Yad-8418e37-x86_64.AppImage`

but now it's at `https://github.com/sonic2kk/steamtinkerlaunch-tweaks/releases/download/Yad-13.0/Yad-13.0-x86_64.AppImage` renaming the release tag from `Yad-13.0` to `Yad-13.0-x86_64.AppImage` should fix it ;)

I'm going to create a little PR for this !

sonic2kk:
Good to know that this issue is separate, thanks for catching this!

Although it's still important to note (for the few Steam Deck users that actually do things properly and check previous issues and PRs) that dependency download failures are ***incredibly common on SteamOS*** for various projects, including SteamTinkerLaunch.

However this case is separate indeed, but, the download may still fail.

<hr>

New release has been published here following sonic2kk/steamtinkerlaunch-tweaks#4: https://github.com/sonic2kk/steamtinkerlaunch-tweaks/releases/tag/Yad-13.0-x86_64.AppImage

AtomHare:
> Good to know that this issue is separate, thanks for catching this! 

No issue at all, glad to help

> Although it's still important to note (for the few Steam Deck users that actually do things properly and check previous issues and PRs) that dependency download failures are incredibly common on SteamOS for various projects, including SteamTinkerLaunch.

Honestly, I never got download issues with my Steam Deck, maybe it's more related to people having bad Wi-Fi connections at home

SilentException:
Hi, I updated STL on my Steam Deck, updated ~/.config/steamtinkerlaunch/uris.conf and removed yad executable to force redownload of new version. Couldn't thing of another procedure. Then i ran steamtinkerlaunch in terminal and it proceeded to download yad but because this:
https://github.com/sonic2kk/steamtinkerlaunch/pull/1077/files#diff-77c6d4bb1bf7c1988ff5a068c856b6c24e113663f2e61cb3a2c98723c0fafc04L18225
was removed, the download link wasn't populated (after debugging the YAIDL was "/Yad-13.0-x86_64.AppImage/Yad-13.0-x86_64.AppImage") and nothing was downloaded. I termporary reverted that one line and everything downloaded correctly.

sonic2kk:
Huh, that's interesting. Thanks for pointing that out. I'll push a fix for that now. I guess it was needed after all, but I'm not sure why. I guess maybe the values in `url.conf` aren't read yet at this point when `steamdedeckt` is called.

sonic2kk:
Should be fixed with 2d0b0ec.

Issue #848, opened 2023-07-03
It is impossible to download mods (MO2)


RyanHakurei:
## System Information
- SteamTinkerLaunch version: 12.12
- Distribution: Arch
- Installation Method: Protonup-qt

## Issue Description
No matter what it is impossible to actually download mods via nxm links. When clicking a link SteamTinkerLaunch appears to open for a few seconds, but nothing shows up in MO2. Upon trying to download a mod via the CLI this error is presented

```
/home/ryan/stl/prefix/steamtinkerlaunch mo2 u "nxm://fallout4/mods/11058/files/48494?linkspam"
Mon Jul  3 05:51:54 PM EDT 2023 INFO - dlMod2nexurl - Attempted to download Url 'nxm://fallout4/mods/11058/files/48494?linkspam' for game 'fallout4', but the source script '/home/ryan/.config/steamtinkerlaunch/mo2/dldata/fallout4.conf' for portable mo2 is missing - trying to start a global ModOrganizer
Mon Jul  3 05:51:54 PM EDT 2023 ERROR - dlMod2globnexurl - Attempted to download Url 'nxm://fallout4/mods/11058/files/48494?linkspam' for game 'fallout4', but the source script '/home/ryan/.config/steamtinkerlaunch/mo2/dldata/global.conf' for global ModOrganizer is missing - giving up
```

So it seems a file that is needed is not being created. Well, I found `global.conf` in another directory, (`.config/steamtinkerlaunch`) BUT
```
/home/ryan/stl/prefix/steamtinkerlaunch mo2 u "nxm://fallout4/mods/11058/files/48494?linkspam"
Mon Jul  3 05:52:23 PM EDT 2023 INFO - dlMod2nexurl - Attempted to download Url 'nxm://fallout4/mods/11058/files/48494?linkspam' for game 'fallout4', but the source script '/home/ryan/.config/steamtinkerlaunch/mo2/dldata/fallout4.conf' for portable mo2 is missing - trying to start a global ModOrganizer
Mon Jul  3 05:52:23 PM EDT 2023 ERROR - dlMod2globnexurl - Attempted to download Url 'nxm://fallout4/mods/11058/files/48494?linkspam' for game 'fallout4', but seems like global '/home/ryan/.config/steamtinkerlaunch/mo2/dldata/global.conf' has incomplete data - giving up
```

So it seems that a file that MO2 needs is not being created, and the directory that file is normally in, `dldata`, isn't even being created.

[...]

Issue #851, opened 2023-07-07
Mod Manager 2 Web URL Portal does not work


jarrard:
Can't open any MO2 link commands or commands into default browser. 

EDIT

As it turns out I was trying to use a preconfigured instance of MO2 which was pre setup. 

It seems things ONLY work with STL if you start a BRAND NEW MO2 install, no option or functionally to hook into pre-installed MO2 such as those from Wabbajack.

jarrard:
I'll close this for now since you can swap in your own MO2 as a symlink and all is dandy. 

Sure it would be nice to just have some option to point STL to your preferred MO2 instance and have it symlink and hotwire it all in, but this works fine.

sonic2kk:
There is no support for wabbajack in STL. See #305.

jarrard:
Yeah I used a VM to install the package.  Someone has managed to get WJ working under Linux but its got problems.

sonic2kk:
I'm not sure if it's what you were referring to in your OP, but NXM links and such are not supported anymore. They break very often and I'm not interested in trying to fix them.

It should also be noted that for any chance of NXM links to work, you need to use MO2 Game Mode. But the usage you have described, while nifty, is not supported by STL in general. Maybe one day if someone makes a PR, but STL isn't a modding tool :-)

jarrard:
I was not specifically talking about NXM but opening links within the MO2 app.  
However installing MO2 vis STL then mapping over my standalone solves that.

PR #1082, opened 2024-04-14
Fix handling of nxm:// URLs for MO2

zerodogg:
It now writes the MO2*Conf files on each startup of MO2, and in addition to the STEAM*/PROTON variables it will now also write MO2PFX, MO2WINE and MO2INST, so that it can launch URLs in an MO2 instance launched by startMO2.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -17564,6 +17564,12 @@ function startMO2 {
 		mkProjDir "$PFXSUTEMP"
 		cd "${MO2EXE%/*}" >/dev/null || return
 
+		if [ -n "$MOINST" ] && [ "$MOINST" == "portable" ]; then
+			updateMO2PortConf
+		else
+			updateMO2GlobConf
+		fi
+
 		if [ -n "$MO2INST" ]; then
 			writelog "INFO" "${FUNCNAME[0]} - WINEDEBUG=\"-all\" WINEPREFIX=\"$MO2PFX\" \"$MO2WINE\" \"$MO2EXE\" -i \"$MO2INST\"" E
 			WINEDEBUG="-all" WINEPREFIX="$MO2PFX" "$MO2WINE" "$MO2EXE" -i "$MO2INST" 2>&1 | tee "$STLSHM/${FUNCNAME[0]}_${IFILE##*/}.log"
@@ -17619,6 +17625,16 @@ function dlMod2nexurl {
 			cd "${GMO2EXE%/*}" >/dev/null || return
 			writelog "INFO" "${FUNCNAME[0]} - Starting portable '$RUNPROTON run ${MO}.exe $1'"
 			STEAM_COMPAT_CLIENT_INSTALL_PATH="$STEAM_COMPAT_CLIENT_INSTALL_PATH" STEAM_COMPAT_DATA_PATH="$STEAM_COMPAT_DATA_PATH" "$RUNPROTON" run "${MO}.exe" "$1" 2>&1 | tee /tmp/RUNMO2DL.log
+			cd - >/dev/null || return
+		elif [ -d "${GMO2EXE%/*}" ] && [ -n "$MO2PFX" ] && [ -n "$MO2WINE" ]; then
+			writelog "INFO" "${FUNCNAME[0]} - Download Url '$1' for game '$MONGURL' using data from portable '$MYPORTDLDAT'"
+			cd "${GMO2EXE%/*}" >/dev/null || return
+			writelog "INFO" "${FUNCNAME[0]} - Starting portable MO2 in prefix '$MO2PFX' using '$MO2WINE'"
+			if [ "$MO2INST" ]; then
+				WINEDEBUG="-all" WINEPREFIX="$MO2PFX" "$MO2WINE" "$GMO2EXE" -i "$MO2INST" "$1" 2>&1 | tee /tmp/RUNMO2DL.log
+			else
+				WINEDEBUG="-all" WINEPREFIX="$MO2PFX" "$MO2WINE" "$GMO2EXE" "$1" 2>&1 | tee /tmp/RUNMO2DL.log
+			fi
 			cd - >/dev/null || return
 		else
 			writelog "ERROR" "${FUNCNAME[0]} - Attempted to download Url '$1' for game '$MONGURL', but seems like portable '$MYPORTDLDAT' has incomplete data - trying global $MO" "E"
@@ -19744,8 +19760,11 @@ function updateMO2PortConf {
 	writelog "INFO" "${FUNCNAME[0]} - Updating '$MYPORTDLDAT' with up to date data"
 
 	{
-		echo "GMO2EXE=\"$GMO2EXE\""
+		echo "GMO2EXE=\"$MO2EXE\""
 		echo "RUNPROTON=\"$RUNPROTON\""
+		echo "MO2PFX=\"$MO2PFX\""
+		echo "MO2WINE=\"$MO2WINE\""
+		echo "MO2INST=\"$MO2INST\""
 		echo "STEAM_COMPAT_CLIENT_INSTALL_PATH=\"$STEAM_COMPAT_CLIENT_INSTALL_PATH\""
 		echo "STEAM_COMPAT_DATA_PATH=\"$STEAM_COMPAT_DATA_PATH\""
 	} > "$MYPORTDLDAT"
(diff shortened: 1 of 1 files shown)


zerodogg:
Resolves #848 (improves #851)

sonic2kk:
Thanks for the PR! I'll do some testing to confirm.

I'd like to know though, why does this NXM link handling work for some people and not others? It works for me already.

If you don't have an answer that's fine. I'll test when I have some time and confirm that at the very least this doesn't break anything.

Really nice to see the community stepping up to help here!

sonic2kk:
For some reason, ShellCheck CI didn't run against this PR. Odd. I'll see if I can manually kick it off!

zerodogg:
It seems like it only wrote the MO2Conf file in certain instances when running the actual game, but never when MO2 was just started on its own. So the file wouldn't exist (or could be wrong, referencing an older game). This patch adds writing that file when starting MO2 as well, and I had to just add some additional vars since starting MO2 references WINE and WINEPREFIX rather than a PROTON.

steamtinkerlaunch
```diff
@@ -17620,6 +17626,16 @@ function dlMod2nexurl {
 			writelog "INFO" "${FUNCNAME[0]} - Starting portable '$RUNPROTON run ${MO}.exe $1'"
 			STEAM_COMPAT_CLIENT_INSTALL_PATH="$STEAM_COMPAT_CLIENT_INSTALL_PATH" STEAM_COMPAT_DATA_PATH="$STEAM_COMPAT_DATA_PATH" "$RUNPROTON" run "${MO}.exe" "$1" 2>&1 | tee /tmp/RUNMO2DL.log
 			cd - >/dev/null || return
+		elif [ -d "${GMO2EXE%/*}" ] && [ -n "$MO2PFX" ] && [ -n "$MO2WINE" ]; then
+			writelog "INFO" "${FUNCNAME[0]} - Download Url '$1' for game '$MONGURL' using data from portable '$MYPORTDLDAT'"
+			cd "${GMO2EXE%/*}" >/dev/null || return
+			writelog "INFO" "${FUNCNAME[0]} - Starting portable MO2 in prefix '$MO2PFX' using '$MO2WINE'"
+			if [ "$MO2INST" ]; then
```
sonic2kk, line 17633: I think this check has a mistake, although ShellCheck interestingly isn't flagging it up.

```suggestion
			if [ -n "$MO2INST" ]; then
```

Let me know if this sounds reasonable :-) This would also mirror what is in `startMO2`, it does a similar check.

steamtinkerlaunch
```diff
@@ -17620,6 +17626,16 @@ function dlMod2nexurl {
 			writelog "INFO" "${FUNCNAME[0]} - Starting portable '$RUNPROTON run ${MO}.exe $1'"
 			STEAM_COMPAT_CLIENT_INSTALL_PATH="$STEAM_COMPAT_CLIENT_INSTALL_PATH" STEAM_COMPAT_DATA_PATH="$STEAM_COMPAT_DATA_PATH" "$RUNPROTON" run "${MO}.exe" "$1" 2>&1 | tee /tmp/RUNMO2DL.log
 			cd - >/dev/null || return
+		elif [ -d "${GMO2EXE%/*}" ] && [ -n "$MO2PFX" ] && [ -n "$MO2WINE" ]; then
```
sonic2kk, line 17629: Just going to "think out loud" about what this section of the code is doing. Let me know if my understanding is on the right track. For reference, I didn't write *any* of the MO2 code in SteamTinkerLaunch.

To me it seems like this controls how ModOrganizer 2 is started, and it looks to essentially mirror how `startMO2` works? That is, "if instance defined, launch with instance selected, otherwise launch global ModOrganizer 2"

In the case of the URL handler though, it is always using the Game MO2 exe, but if no instance is defined it will just launch without an instance selected. Which probably won't happen when using a URL handler, unless the instance was never set up I guess.

My only "concern" here is in case ModOrganizer 2 could get into a bad startup state where it could never launch, but I don't think that can happen here from reviewing the code :-) This looks pretty safe. GMO2EXE is likely what we want to launch anyway, since that should be what gets started when you launch MO2 in Game Mode to play a game.

sonic2kk (approved):

sonic2kk:
Oh, right, can you bump the version at the top of the file to `PROGVERS="v14.0.2040415-1"`? :-)

I can't leave a review comment about that.

sonic2kk:
Changes make sense overall and don't break anything. ShellCheck is also green locally. I'm not sure why it didn't run on CI but I'll look into that separately, that isn't an issue for your PR.

MO2 still starts up in Standalone Mode and Game Mode correctly. NXM URLs still work for me with no change in behaviour, and code makes sense. I did request a slight change but since ShellCheck isn't complain it's not a huge deal. I've also asked a question just for my own clarity, since I'm not overly familiar with this part of the code. Version needs bumped but at worst I can bump it myself.

Thank you a ton for taking the time to investigate and fix this issue! It comes up _all_ the time, despite me highly disapproving of this feature in MO2 to begin with, I appreciate that you spent the time to fix it.

steamtinkerlaunch
zerodogg: What it does here is check if the GMO2EXE directory exists, and it has the required MO2PFX and MO2WINE variables, all there of which is coming from the config file it just sourced (the logic is: the steamtinkerlaunch instance that start MO2 dumps its variables into a file, which 'steamtinkerlaunch mo2 u' reads and uses to launch MO2 with the URL specified). So what this does is just check that it has the vars it needs., and if it doesn't it'll fall back to the codepath that's already there (which appears to try a global one, which also fails here, and then it just errors out without this patch).

steamtinkerlaunch
zerodogg: Yup. I think both work, but -n is definitely the proper way. Pushed a fix.

zerodogg:
> Oh, right, can you bump the version at the top of the file to `PROGVERS="v14.0.2040415-1"`? :-)
> 
> I can't leave a review comment about that.

Done

steamtinkerlaunch
zerodogg: ...the trouble with the code that was there is both that it doesn't actually write the file when just starting MO2 from the GUI, but also that the file when written doesn't contain the variables that it needs (namely MO2PFX and MO2WINE). I tried just making sure the file was written initially, but that failed since it didn't have RUNPROTON, which is why I added the new vars and this additional code path which uses them, leaving the old one intact as well since, as you said, it works in some cases - which I'd rather not break (which is also why it does this check *after* checking for the RUNPROTON variant, to make sure the new one can't break the old one).

zerodogg: Here's an example of the config file it's exchanging information through, when using the 'start MO2 through the GUI, then use an nxm:// link':
```
[0 zerodogg@browncoats ~]$ cat .config/steamtinkerlaunch/mo2/dldata/newvegas.conf
GMO2EXE="/mnt/ssd4t/SteamLibrary/steamapps/compatdata/22380/pfx/drive_c/Modding/MO2/ModOrganizer.exe"
RUNPROTON=""
MO2PFX="/mnt/ssd4t/SteamLibrary/steamapps/compatdata/22380/pfx"
MO2WINE="/mnt/ssd4t/SteamLibrary/steamapps/common/Proton 9.0 (Beta)/files/bin/wine"
MO2INST=""
STEAM_COMPAT_CLIENT_INSTALL_PATH="/home/zerodogg/.local/share/Steam"
STEAM_COMPAT_DATA_PATH="/mnt/ssd4t/SteamLibrary/steamapps/compatdata/22380"
```

So it's getting all of those vars it's using from there. RUNPROTON isn't used to start mo2, so it's empty, and MO2INST is only used in certain cases, hence the test to see if that too is empty.

sonic2kk: Makes sense based on my understanding of what the NXM link handlers need and how the existing MO2 logic is in place. Thank you!

sonic2kk:
Awesome change, thank you! Will merge now.

zerodogg:
Thanks ! :)

sonic2kk:
Changelog has been updated with credit to you :slightly_smiling_face:  https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog#fixes

frostworx:
> Really nice to see the community stepping up to help here!

full ack :heart:

EDP2289 (approved):

PR #1086, opened 2024-04-18
Steam Linux Runtime: Re-enable parsing Steam Linux Runtime from Compatibility Tool start command

### LINKED ISSUE #1084: CS2 crashes saying is not launched with Steam for Linux
## System Information
- Distribution: Garuda Linux (Arch Linux)
- Installation Method: Package Manager

## Issue Description

Hey guys. So, I tried to search as much as I could because I thought I was just doing something stupid. But, I found nothing (which makes me more suspicious of being stupid hehe), so, I came here for ask for help haha.
Basically, when I try to run CS2 using `launch options` (since it is a native game), it crashes. And crashes saying that CS2 was not launched using Steam for Linux runtime. I even tried a clean installation for both Steam and Steamtinkerlaunch. I'm attaching the logs and the screenshot that shows that CS2 is running through Sniper (Steam for Linux runtime). As I said, I'm using a fresh install, so default settings for STL.

From the things I read in STL docs, I think that this is because CS2 command is coming with `--verb=waitforexitandrun`. You guys can see in the logs as well.

## Logs

image
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/15028093/steamtinkerlaunch.log)
[730.log](https://github.com/sonic2kk/steamtinkerlaunch/files/15028097/730.log)
[730.log](https://github.com/sonic2kk/steamtinkerlaunch/files/15028102/730.log)
image

sonic2kk:
Fixes #1084

Primarily (but not exclusively) allows native games that have a Steam Linux Runtime selected (either by the user, or by Valve as sometimes they force SLR 1.0 or 3.0 for games) to launch with the SLR selected.

Previously, the SLR would be ignored entirely if it came from the compatibility tool, but this causes incompatibilities with games that REQUIRE a specific SLR version (ex: CS2 requires SLR 3.0 and is selected by Valve, the user doesn't have a choice)

This change allows STL to read the SLR coming from the compat tool launch, so if a game is launched with a specified SLR version, we will use this instead of ignoring or building the start command ourselves.

TODO:
- [x] Test with native games forced from Steam to use SLR 1.0 and make sure 1.0 is being used correctly
- [x] Test with Proton games and inspect log to ensure the SLR it needs is used correctly (i.e. check that the path in this PR is not followed for Proton games) - In other words, make sure this PR doesn't cause any regressions
- [x] Test with native games that don't have any compat tool selected and check that the SLR 1.0 is appended properly and NOT from this new code path -- Again, makes sure this PR doesn't cause any regressions

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240415-1"
+PROGVERS="v14.0.20240419-1 (reenable-get-slr-from-compattool-launch-cmd)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -159,7 +159,11 @@ FIXGAMESCOPE=0
 SMALLDESK=0
 VTX_DOTNET_ROOT="c:\\Program Files\\dotnet\\\\"
 STLQUIET=0
-SLRAID="1070560"  # Hardcoded SLR AppID for native games -- See `setSLRReap`
+
+# Hardcoded SLR AppID for native games -- See `setSLRReap`
+# TODO refactor to be array for native AIDs? Some games (CS2) can use SLR 3.0
+# We should allow the user to choose whether to use SLE 1.0 or 3.0 for native games
+SLRAID="1070560"
 DEFWINEDPI="96"
 WINEDPIVALUES="${DEFWINEDPI}!120!150!240"
 SGDBTIMEOUT="15"
@@ -2271,8 +2275,10 @@ function setGameVars {
 	fi
 
 	if grep -q "$SLR" <<< "$@"; then
+		writelog "INFO" "${FUNCNAME[0]} - Found SLR is launch option"
 		HAVESLR=1
 	else
+		writelog "INFO" "${FUNCNAME[0]} - No SLR is in launch option"
 		HAVESLR=0
 	fi
 
@@ -2380,6 +2386,7 @@ function setGameVars {
 	fi
 
 	if grep -q "$SLR" <<< "${WIPDGCMD[@]}"; then
+
 		HAVESLRCT=1
 	else
 		HAVESLRCT=0
@@ -20145,7 +20152,7 @@ function fixCustomMeta {
 }
 
 function initPlay {
-	HAVESLR=0
+	# HAVESLR=0
 	HAVESLRCT=0
 	HAVEREAP=0
 	HAVESCTP=0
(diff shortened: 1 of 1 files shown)


sonic2kk:
Copied from https://github.com/sonic2kk/steamtinkerlaunch/issues/1084#issuecomment-2065480765

Tested and the following scenarios work:
- CS2 works out of the box, with Valve Testing's enforce SLR 3.0 (verified in log that the new code path is followed)
- CS2 crashes as expected with Steam Linux Runtime disabled from Steam (this verifies the SLR is disabled as expected), and works again if the SLR is enabled (verified in code that SLR is seen as disabled)
    - Start command was `STL LAUNCH COMMAND: '/home/emma/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=730 -- /run/media/emma/2B/Games/steamapps/common/Counter-Strike Global Offensive/game/cs2.sh -steam'`
- Proton games work as expected and still follow the code path to pull the SLR from the compat tool `toolmanifest.vdf` (The so-called "Pressure Vessel Funtime 2nd Edition Ver. 2.31" code path is used)
    - Tested PowerWash Simulator and Sonic Adventure 2
- Proton games are hit-and-miss with the SLR disabled, sometimes crashing on launch, as expected, so no change in behaviour
- Native games with NO COMPATIBILITY TOOL ENFORCED (none from Valve and none by the user) will still use the native Linux SLR 1.0, so no change in functionality
- Native games with the SLR disabled will correctly ignore it, so no change in functionality 
    - This can be verified by checking the processes in system monitor: no additional `srt-bwrap` and `pressure-vessel-adverb` processes are spawned on game launch with the SLR disabled, but are spawned with the SLR enabled
- Native games WITH SLR 1.0 FORCED FROM PROPERTIES -> COMPATIBILITY will use the SLR specified here using the new code path that pulls it from the launch command and not the `toolmanifest.vdf` (verified using log)
- Native games WITH SLR 1.0 FORCED FROM PROPERTIES -> COMPATIBILITY but which have SLR disabled from SteamTinkerLaunch Game Menu will correctly ignore it
    - This can once again be verified from the processes running
    - The launch command still includes some reference to the Steam Linux Runtime, but as it is missing various arguments, it is likely harmless. We could probably grep and remove this, but I think we can just ignore it as it doesn't have any effect on whether the SLR is used or not. This also only happens in the case of a native game using SLR 1.0 with the SLR disabled from STL settings by the user.

This PR is ready to merge pending feedback from OP of original issue. In my testing this works as expected and fixes the issue, without causing regressions, and actually adds the benefit of STL being able to pick up the SLR for native games if enforced by the user, which can be handy! This means users that manually enable SLR 3.0 can use it too (STL cannot yet allow SLR 3.0 for native titles).

sonic2kk:
In future, we should have a dropdown to select the Native Linux Steam Linux Runtime, and let the user decide  to use 1.0 or 3.0. We should make it clear that this is for native titles only.

This value can override `SLRAID`, which should still default to `1070560` (AID for SLR 1.0 Scout) but we should have a dropdown to choose if it should be SLR 3.0 Sniper. Then the Pressure Vessel Funtime code can pick up and check for SLR 3.0. If the syntax is the same for the launch command for 1.0 and 3.0, then no code change should be needed, since Funtime will work with the ID for 3.0.

sonic2kk:
Confirmed to fix the issue by OP, this needs a little bit of comment cleanup, version bump, and it's good to merge.

Issue #1087, opened 2024-04-20
[Feature Request] Add dropdown to select Steam Linux Runtime version for Native Games


sonic2kk:
## System Information
- SteamTinkerLaunch version: v14.0.20240420-1
- Distribution: Arch Linux

## Feature Description
*(See also: #1084)*

Valve now have two Steam Linux Runtime versions that can be used for native games: "Steam Linux Runtime 1.0 (Scout)" and "Steam Linux Runtime 3.0 (Sniper)". If no Steam Linux Runtime is specified in the start command for a native title (i.e. when a native game is not started with SLR 1.0 or 3.0 selected as a compatibility tool), and if the Steam Linux Runtime is enabled (it is by default), then SteamTinkerLaunch will always attempt to use Steam Linux Runtime 1.0 with native games.

However, native games can also use Steam Linux Runtime 3.0. We should default to SLR 1.0 (as it is currently the most widely used and reliable for native titles), but add a dropdown to the Game Menu to allow the user to select whether a title should use SLR 1.0 or 3.0. We should make a new config variable for the per-game config file that tracks this, and default to `SLRAID`. We should then update `setSLRReap` to use this value instead of the hardcoded `SLRAID`.

We would need some way to map the value on the dropdown with the AppID for SLR 1.0 and 3.0 (we can probably hardcode these, and maybe update `SLRAID` to specifically say `SLRAIDSCOUT`, and then create `SLRAIDSNIPER`. Implementation on this part specifically will be the most complex.

**This will only impact native games, for Proton games we will always check the required SLR in `toolmanifest.vdf`!**

If the selected SLR is not installed, `setSLRReap` will handle logging and notifier usage, so no extra work should be needed here.

**NOTE**: We need to check if the syntax for SLR 1.0 and 3.0 are the same. If they are, we should be in the clear, but otherwise we will need specific logic to build a different launch command if SLR 3.0 is selected.

sonic2kk:
The syntax between SLR 1.0 and SLR 3.0 is different. Sniper uses `SteamLinuxRuntime_sniper/_v2-entry-point` (very similar to what Proton does) whereas Scout uses `SteamLinuxRuntime_soldier/_v2-entry-point --verb=waitforexitandrun -- SteamLinuxRuntime/scout-on-soldier-entry-point-v2 ` -- it appends an extra `SteamLinuxRuntime/scout-on-soldier-entry-point-v2`

[...]

PR #1090, opened 2024-04-21
Game Menu: Add option to disable Native Linux Steam Linux Runtime coming from Compatibility Tool

### LINKED ISSUE #1088: [Feature Request] Allow Overriding Steam Linux Runtime Compatibility Tool Version for Native Games
## System Information
- SteamTinkerLaunch version: v14.0.20240420-1
- Distribution: Arch

## Feature Description
Sometimes, a native game may be launched with a Steam Linux Runtime Compatibility Tool selected. This can be set by the user in the Properties dialog, or it may be selected by Valve Testing such as in the case of Counter-Strike 2 (#1084). If a game is launched with a Steam Linux Runtime specified in the incoming start command, this means the SLR version is specified by the compatibility tool, and so STL will use this instead of finding the SLR to use for the native game itself (#1086). It basically trusts whatever the incoming SLR is and uses that.

In 99% of cases, this is fine, because:
1. If the user selected the SLR compatibility tool, they probably *want* to use this specifically and *not* the one STL will look for
2. If this was selected by Valve Testing, it was likely for compatibility, and we would want to always use this by default (in the case of CS2, it will fail of SLR Sniper is not used, so it is a requirement to use the SLR specified by the launch command, which comes from the forced SLR 3.0 compatibility tool)
    a. If Valve Testing selects a compatibility tool, there is no way to really "disable" it -- In the case of CS2, you can't easily specify to *not* use SLR 3.0, as it is not forced under the Properties dialog. The compat tool selected by Valve Testing is never shown as forced under the Properties dialog, so it is always the default.
    b. STL still lets you disable the SLR though, so if you launch CS2 and untick "Use Steam Linux Runtime" from the STL Game Menu, it won't be used.

But a user may want to disable the forced SLR, and then they may want to override the SLR coming from the selected SLR compatibility tool launch command, and use whatever STL is using. We should offer a checkbox with the other SLR options to "Ignore Steam Linux Runtime from Launch Command", which basically means if the launch command contains an SLR (i.e. if the game was started with a Steam Linux Runtime compatibility tool selected) then we should ignore this and let SteamTinkerLaunch fetch the compatibility too.

A use case for this: If a title is selected by Valve Testing to use SLR 1.0, but a user wants to use SLR 3.0, then once #1087 is implemented they could select to use SLR 3.0, and then with this feature they can tick the box to ignore the Valve-forced SLR 1.0 for that game, and tell STL to instead fetch SLR 3.0 and use that to run the game.

sonic2kk:
Implements #1088.
Soft prerequisite for #1087.

If a native game is launched with a Steam Linux Runtime Compatibility Tool selected, either by the user or Valve Testing, the Steam Linux Runtime will come in the game launch command. However a user may want to ignore this for any reason, so this adds a checkbox to allow users to ignore the incoming Steam Linux Runtime and let SteamTinkerLaunch find it.

This has only been minimally tested, and needs regression testing. Also, langfiles need to be updated. I noticed in testing `DESC_USESLR`, a string unrelated to this PR, has a typo, so I fixed it in the `english.txt`. This will also need to be fixed in other langfiles where appropriate alongside adding the new strings for the checkbox in this PR to the other langfiles.

This PR is a soft prerequisite for #1087 because it makes that feature more useful. By allowing to disable the incoming SLR in this PR, it means choosing a different SLR is much easier. Once #1087 is implemented it means a user could use this PR to disable a Valve Testing-selected SLR 1.0, and then use #1087 to select SLR 3.0 to let STL find that SLR for the game.

<hr>

Following this PR and any work for related issue #1087, we need to update the Steam Linux Runtime wiki page to have more information about what it is, how SteamTinkerLaunch uses it (will attempt to find the required Steam Linux Runtime stipulated by a Proton version's `toolmanifest.vdf`, otherwise will default to SLR 1.0 for native games, and if an SLR is incoming for a native game from a selected SLR compatibility tool then STL will always use this unless this PR's checkbox is enabled).


sonic2kk:
CI is running twice, god dammit...

sonic2kk:
Okay, CI seems fixed now. It should only run on push for branches, and will run for pull request events if the branch is from a fork.

sonic2kk:
Tested the following games, eveything works as expected:
- Tested CS2 (native)
    - Default, it launches with SLR 3.0 selected by Valve
    - SLR disabled, it crashes and no SLR in start command in log
    - SLR enabled and Ignore incoming SLR enabled, it crashes on launch and the log shows it's using SLR 1.0 and logging that it correctly ignored the SLR from the incoming start command
- Tested To the Moon (native)
    - Default, it launches with no SLR from Steam but gets the SLR from SteamTinkerLaunch because  the option is enabled, it correctly logs that it is getting it from Pressure Vessel Funtime and the SLR 1.0 is appended to the start command
    - SLR disabled, this works, although previously To the Moon required the SLR (I guess there was an update to fix that? All Freebird Games' games got an update not long ago to improve Linux compatibility on SteamOS). Launch command doesn't show SLR and the log shows it skipped fetching the SLR
    - With SLR 1.0 set from Compatibility settings, log shows that SLR is gotten from incoming launch command and appended correctly.
    - With SLR 1.0 set from Compatibility settings and the option to ignore incoming SLR enabled, log shows that incoming SLR got ignored and that SteamTinkerLaunch fetched and appended SLR 1.0 to the launch command
- Tested PowerWash Simulator (Proton, should be zero change in functionality as this option should do nothing for Proton)
    - Tested default settings (SLR enabled, ignore incoming SLR disabled), game launches with Pressure Vessel Funtime being used to fetch the SLR and append it to the start command
    - Tested Ignore incoming SLR enabled, no change in functionality and nothing logged as `HAVESLR` should always be `0` for Proton since it never comes in the launch command.

This PR seems to be ready to merge pending a version bump.

EDIT: Oops, didn't push the change I had locally to make a change in logic to when we check `HAVESLR -eq 1`. Will push with the version bump and merge.

PR #1092, opened 2024-04-21
ReShade: Use `ALTEXEPATH` to install ReShade DLLs for Custom Commands with `ONLY_CUSTOMCMD=1`

### LINKED ISSUE #1089: Reshade always installs into custom command directory when using ONLY_CUSTOMCMD with no way to override
## System Information
- SteamTinkerLaunch version: steamtinkerlaunch-v14.0.20240420-1
- Distribution: Garuda Linux
- Installation Method: AUR

## Issue Description

this is the same exact issue as https://github.com/sonic2kk/steamtinkerlaunch/issues/669 except its when the user sets `ONLY_CUSTOMCMD` and wants reshade installed into the game's path 

in this case reshade will always be installed into the custom commands directory regardless of `CUSTOMCMDRESHADE`

When using a custom command INSTEAD of the actual game exe stl installs all files into the path of the custom command.

as you can see [here](https://github.com/sonic2kk/steamtinkerlaunch/blob/04ae6e9ea8733d980fe5d78fa93327230a9c34e2/steamtinkerlaunch#L9583) the reshade destination directory is set to be the same as the shader destination directory

`SHADDESTDIR` gets set to[ path to the executable which is getting launched](https://github.com/sonic2kk/steamtinkerlaunch/blob/04ae6e9ea8733d980fe5d78fa93327230a9c34e2/steamtinkerlaunch#L9217C20-L9217C34) which in this case is our custom command.

We actually have a workaround for this, though
if we have an `ALTEXEPATH` then that gets set to `SHADDESTDIR` (strangely in two places dup?)
https://github.com/sonic2kk/steamtinkerlaunch/blob/04ae6e9ea8733d980fe5d78fa93327230a9c34e2/steamtinkerlaunch#L5314

https://github.com/sonic2kk/steamtinkerlaunch/blob/master/steamtinkerlaunch#L9208

The strange thing is, as you can see in my log, I am setting `ALTEXEPATH` to be the game directory
```
setShaderDest - Overriding SHADDESTDIR to '/home/zany130/.local/share/Steam/steamapps/common/P3R/P3R/Binaries/Win64' because ALTEXEPATH is set to '/home/zany130/.local/share/Steam/steamapps/common/P3R/P3R/Binaries/Win64'
```
It even installing the shaders into the directory (without this option set it will also install the shaders it the directory of my custom command)

my guess is that `SHADDESTDIR` is getting set to `ALTEXEPATH` it installs the shaders and then somehow `SHADDESTDIR` gets overwritten again to the custom commands path?

## Logs
[steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/15049250/steamtinkerlaunch.log)

## P.S

Sorry for the long issue lol I was trying to debug it myself since I still mostly remember how all the reshade stuff worked, but got stuck

sonic2kk:
Fixes #1089.

This PR adds an extra check to `setFullGameExePath`, where if an alternative EXE path is defined, this should be used in place of the custom command EXE directory. This makes sense to me, as the order of priority should be to use `ALTEXEPATH` above all other paths when deciding where to put things, and otherwise either the custom command folder (if custom commands are enabled) or the game EXE folder should be used.

This should address an issue where when installing ReShade with `ONLY_CUSTOMCMD` enabled, ReShade would install shaders to the ALTEXEPATH (as this is handled by `setShadDest`) but it would not do this for the ReShade DLLs - handled by a call in `installReshade` to `setShadDestDir`, which in turn calls `setFullGameExePath` to get the path where the DLLs should be installed.

This will need some testing to ensure this doesn't cause any regressions. The following cases should be testing:
- [x] Regular game launches should still work fine
- [x] Regular game launches with an Alt EXE dir should respect the chosen alt EXE path
- [x] Custom commands with alt EXE paths chosen should respect it with ReShade
- [x] Custom commands with alt EXE paths chosen should respect it without ReShade

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240421-1"
+PROGVERS="v14.0.20240421-1 (fix-reshade-dlls-custcmd)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -9582,7 +9582,9 @@ function installRSdll {
 function installReshade {
 	if [ "$USERESHADE" -eq 1 ]; then
 		prepareReshadeFiles
-		setShadDestDir
+		# setShadDestDir
+		# Use setShaderDest to avoid overwriting ALTEXEPATH if defined
+		setShaderDest
 
 		INSTDESTDIR="$SHADDESTDIR"
 
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -9132,8 +9132,10 @@ function ShaderRepoDialog {
 
 function setFullGameExePath {
 	if [[ ( "$USECUSTOMCMD" -eq 1 && -f "$CUSTOMCMD" && "$CUSTOMCMDRESHADE" -eq 1 ) || "$ONLY_CUSTOMCMD" -eq 1 ]]; then
-		FGEP="${CUSTOMCMD%/*}"
-
+		# Use Alternative EXE Path if defined instead of custom command path
+		# We should only use the custom command directory if no alternatiive EXE path is defined, and
+		# we should prioritise the alt path if it is defined
+		FGEP="${DEFINEDALTEXEPATH:-${CUSTOMCMD%/*}}"
 		writelog "INFO" "${FUNCNAME[0]} - Using the directory '$FGEP' of the used custom command as absolute game exe path"
 
 		if [ "$CUSTOMCMDRESHADE" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


zany130:
sorry for the late test. I got sidetracked with other things, but....

it seems its still installing into the custom command dir  [steamtinkerlaunch.log](https://github.com/sonic2kk/steamtinkerlaunch/files/15050550/steamtinkerlaunch.log)

sonic2kk:
Oops, that's because I'm an idiot and removed a line that is supposed to be there... `DEFINEDALTEXEPATH` is never set so it never uses it. Really sorry about that...

sonic2kk:
Should work now with 3b6ffb0.

zany130:
Works as expected for my case of using persona 3 reload  with the reloaded mod manager. 

Ill also test to see if everything works as expected when there is no custom command. 

Unfortunately, I don't think I have any games that have launchers in a different directory than the main executable (Star Wars, maybe?).

sonic2kk:
I'll do some testing tomorrow of that case as well then keeping an extra eye on it, but sounds like this is in a good position! Thanks!

sonic2kk:
Didn't get a chance to test this again today, but once I have time I will and will merge this hopefully soon into the week.

zany130:
no worries I haven't been able to test extensively but I did fire up a few different games and mad sure it installs Reshade games no regressions there

sonic2kk:
Seems safe in my tests as well. Pending CI this will be merged :-)

PR #1097, opened 2024-04-28
Implemented MO2 silent mode

zerodogg:
This successfully launches Fallout: New Vegas with mo2 in silent mode for me

PR #1100, opened 2024-04-29
ModOrganizer 2: Finalize Silent Mode

sonic2kk:
Finalizes the ModOrganizer 2 Silent Mode implementation. While #1097 implemented the logic to make Silent Mode work, this adds some UX touches to make it easier for users to enable.
- Adds back the "Silent" button on the ModOrganizer 2 Wait Requester, removed a long time ago.
- Adds `silent` option to MO2MODE dropdown on Game Menu.
- Updates the actions on the Wait Requester to actually respect `gui`/`silent`/`disabled`
    - Newly added "Silent" button will start in silent mode
    - Wait Requester timeout will now default to the selected MO2 mode. Before, we only had two; `disabled` and `gui`, so we always defaulted to `gui` as the sensible choice for the Wait Requester timing out. But now it will actually respect whatever the user selected in `disabled`. Having said that, the Wait Requester should only appear if `MO2MODE` is not `disabled`, so the user won't see it if the mode isn't `gui` or `silent` anyway.
- Minor code refactoring

This needs more testing from my side, and once completed this can be merged to fully finalize the MO2 Silent implementation!

Note: I did some work to update the ModOrganizer 2 wiki page, but another pass over it might be good to make sure all the intricacies are documented.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240612-1"
+PROGVERS="v14.0.20240330-2 (finalize-mo2-silent)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
I've been thinking about the implementation of this, and I may rework it before merging.

sonic2kk:
Eh, I don't think we need to make any changes here. I was concerned that showing the Wait Requester when Silent Mode was enabled wouldn't make sense but there are cases where the Wait Requester timeout may be 0, so in that case Mo2 can be started entirely silently.

I think this is good to merge after a version bump and langfile check.

PR #1104, opened 2024-05-12
GameScope: Force DXVK_HDR=1 if GameScope HDR is enabled

### LINKED ISSUE #1103: [Bug] Troubleshooting HDR Activation in Steamtinkerlaunch
## System Information
- Distribution: Arch Linux
- Installation Method: AUR
- Gamescope-git

## Issue Description
When using the normal version of Steam with Proton: Adding `gamescope -W 5120 -H 1440 --hdr-enabled --fullscreen %command%` to the command parameter of a game allows turning on HDR within it. This method aligns with the instructions outlined in this article : [https://zamundaaa.github.io/wayland/2024/05/11/more-hdr-and-color.html](https://zamundaaa.github.io/wayland/2024/05/11/more-hdr-and-color.html)

With Steamtinkerlaunch set as a Compatibility Tool: Enabling gamescope usage in the gamescope menu of Steamtinkerlaunch does make the game run inside gamescope, but enabling the HDR support option (which is supposed to only add --hdr-enabled to the launch command of gamescope) is not working, or at least it is impossible to enable HDR within games running with it.

sonic2kk:
Fixes #1103.

If DXVK_HDR is disabled, GameScope will still be unable to use HDR. We should assume if GameScope HDR is enabled that a user also wants HDR enabled from the DXVK side. There should be no reason to pass --hdr-enabled to GameScope if you don't actually want to use HDR.

Still need to update langfiles but this should be good to go otherwise.


sonic2kk:
Noticed while updating the langfiles that the string for HDR nits (`DESC_GSHDRSCNITS`) was not updated following its bump in #855. Can't believe that went unnoticed for over a year, but I guess HDR in general has only really picked up in the last few months.

PR #1111, opened 2024-05-20
Steam Deck dependency enhancement

### LINKED ISSUE #859: Steam Deck Experience Improvements [Community Contributions Wanted]

AtomHare:
This draft implements the base discussed in https://github.com/sonic2kk/steamtinkerlaunch/issues/859#issuecomment-2067536075.

Here what's implemented in this draft and some hurdles I noticed that we will need to fix before marking it ready.

- [X] 1. STL starts up on Steam Deck.
- [x] 2. If the Automatic Update checkbox is off, don't do any checks and skip.
- [X] 3. Otherwise, we check if SteamTinkerLaunch has updated by comparing the value in `lastvers` with that of the incoming version (we can get the incoming version from `PROGVERS`) - managed by `checkSteamDeckSTLUpdated`
- [x] 4. If the version stored in `lastvers` is the same as the incoming version, do nothing (this will prevent the constant echo spam on each launch stating that STL needs to be updated) 
- [X] 5. If the version in `lastvers` does not match the incoming version string, we should try to update our dependencies, as the newer STL version may specify a newer version of a given dependency
- [X] 6. Once everything is complete, update the version string in `lastvers` so that we know we already checked for dependency updates for this version. - See in issues, I think we could force the overwrite each time the libs are updated

Issues:
- ~~If coming from an already existing install, we will create the `lastvers` file ⇒ leading to not update the libs on the first stl update after this PR, maybe we need to create the lastvers after the installation of the libs each time (overwrite the `lastvers` or creating it for the first time, either way it's a simple bash pipe), and make `checkSteamDeckSTLUpdated` return 0 if the file doesn't exist either.~~
- Is this PR also related with the notification spam of STL at the launch of a game ? (Only in desktop mode)
- Already downloaded archives and installed libs may cause issues. Maybe we should clean up the downloaded archives after their successful extraction ? The handling probably needs to be updated too, I'm not very familiar with the codebase.

Note: Also while editing I saw that sometimes there are 4 spaces while in the script it's mainly tabs. Should we make some reformat to fix this later on ?
Note 2: **This is currently untested**


sonic2kk:
Woah! Thank you for making this PR!

> If the Automatic Update checkbox is off, don't do any checks and skip. - not sure where it belongs

We can put this on the Global Menu, but indeed, it is tricky to know the best spot for this... Maybe under "Misc" settings?

https://github.com/sonic2kk/steamtinkerlaunch/blob/master/steamtinkerlaunch#L5664

> If the version stored in lastvers is the same as the incoming version, do nothing (this will prevent the constant echo spam on each launch stating that STL needs to be updated) - Not sure if checkSteamDeckSTLUpdated already does it.

From a quick glance at the logic here (https://github.com/sonic2kk/steamtinkerlaunch/pull/1111/files#diff-77c6d4bb1bf7c1988ff5a068c856b6c24e113663f2e61cb3a2c98723c0fafc04R25928) it seems like it would already handle it. 

The `if` check is `if [ -f "$(command -v "$DEPCMD")" ] && [ "$CHECKCMD" = "OK" ] && [ "$( checkSteamDeckSTLUpdated )" -eq 0 ]; then` - And if all of these are true, then we assume we don't need to install any dependencies.
- `-f "$(command -v "$DEPCMD")"`` - If the dependency file exists.
- ` "$CHECKCMD" = "OK"` - If the dependency can actually be ran.
- `"$( checkSteamDeckSTLUpdated )" -eq 0` - If the SteamTinkerLaunch version has not changed since the last run.

Then skip the dependency update logic. That should prevent STL constantly trying to update itself.

> If coming from an already existing install, we will create the lastvers file ⇒ leading to not update the libs on the first stl update after this PR, maybe we need to create the lastvers after the installation of the libs each time (overwrite the lastvers or creating it for the first time, either way it's a simple bash pipe), and make checkSteamDeckSTLUpdated return 0 if the file doesn't exist either.

I think the easiest solution here is to run `clearDeckDeps` if there is no `lastvers`; assume all dependencies are bad and need wiped, and pull the fresh ones. Then everytime a user updates they'll get fresh dependencies.

If I'm missing something here let me know :-)

> Is this PR also related with the notification spam of STL at the launch of a game ? (Only in desktop mode)

Yes. The reason it's only in Desktop Mode is because notifiers don't work in a GameScope session.

> Already downloaded archives and installed libs may cause issues. Maybe we should clean up the downloaded archives after their successful extraction ? The handling probably needs to be updated too, I'm not very familiar with the codebase.

Yeah, there's going to be a conflict here.

SteamTinkerLaunch explicitly does NOT clean up archives because I want the ability to use it entirely offline eventually. So the idea was to allow users to manually place the archives and have them be extracted.

But with automatically updating the dependencies, I'm not sure how to serve both needs. The complexity around solving this is very off-putting. 

>  Also while editing I saw that sometimes there are 4 spaces while in the script it's mainly tabs. Should we make some reformat to fix this later on ?

Some consistency around this would be good, but making sure it doesn't interrupt places where spaces are used deliberately is important. Unless we go with using spaces instead of tabs rather than the other way around.

I guess it's something I can look into at another time :-)

AtomHare:
This draft is still **UNTESTED**.

What's new :
- Clear deps on the first run if the `lastvers` file doesn't exist, fix any weird issues in the case of an existing install
- Added the autoupdater settings - if you have any naming recommendation feel free to tell me I will correct it
- Set the PROGVERS to differentiate it from the master branch

Does everything look good to you ? Before jumping on the ship in order to test it, I wanted to know if "theoretically" all the logic is in place.

steamtinkerlaunch
```diff
@@ -25923,7 +25928,13 @@ function checkSteamDeckDependencies {
...
 		CHECKCMD="$($DEPCMD &> /dev/null --version && echo "OK" || echo "NOK")"
 
-		if [ -f "$(command -v "$DEPCMD")" ] && [ "$CHECKCMD" = "OK" ]; then
+
+		# The check now says, don't update dependencies if all of these conditions are true:
+		#  1. The dependency file exists
+		#  2. The dependency can actually be ran, confirming it is a valid file
+		#  3. SteamTinkerLaunch has not updated or autoupdater isn't enabled, so there would be no change in dependency version and thus no need to update
+		# If any of these are false, we need to check our dependencies (if a file is missing we would need to update, or if it cannot be used we need to update, and also if STL updated we may need a newer version, so update).
+		if [ -f "$(command -v "$DEPCMD")" ] && [ "$CHECKCMD" = "OK" ] && { [ "$( checkSteamDeckSTLUpdated )" -eq 0 ] || [ "$( STEAMDECK_AUTOUP )" -eq 0 ] }; then
```
sonic2kk: ShellCheck is angry about [SC1083](https://www.shellcheck.net/wiki/SC1083), but I think it's slightly misleading. I think the way this check is done simply needs changed.

This is the `if` check for checking if we need to update a dependency. `checkSteamDeckSTLUpdated` checks if the last version and current version match.

I guess what we're trying to do here is a "compound" check, right? At least in my head the check is supposed to be like this in other languages that have much saner syntax than Bash: `if -f DEPCMD and CHECKCMD == "OK" and ( checkSteamDeckSTLUpdated returns 0 and STEAMDECK_AUTOUP == 0 )`. The last two "terms" so to speak are treated as one conditional, since we only care if the STL version changed when auto-updates are enabled.

Compound checking like this in Bash is something that constantly escapes me and I have to keep looking it up in the code. But the following should work: `if [[ ( "$( checkSteamDeckSTLUpdated )" -eq 0 && "$STEAMDECK_AUTOUP" -eq 1 ) && -f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK" ]]; then`. I validated that ShellCheck at least doesn't get angry anymore :-)

Note that I changed the `STEAMDECK_AUTOUP` check.
- Before, it was like this: `"$( STEAMDECK_AUTOUP )" -eq 0`
    - This may work, ShellCheck didn't complain about it at least.
- However using `"$( FOOBAR )"` is typically to capture function returns. When you just want to check a variable, you can do `[ "$FOOBAR" -eq 0 ]`.
    - I used the `[]` here because that's typically how you do a test, but when doing compound checks you have to use the funky `[[]]` and put your grouped terms in `()`. For example `[[ ( foo && bar ) || ( fizz && buzz ) && foobar && fizzbuzz ]]`.
- I also changed it to an `-eq 1` check, because we want to make sure the value is enabled.
- EDIT: I just noticed I swapped the order, I don't think this matters much but the part of the code I was referencing on how to do this had the "compound" terms at the beginning. But I think logically you can put the terms in parentheses anywhere.

At least, that is my understanding, and I'm certainly not a Bash expert! But I hope I helped clarify a little bit about the suggestions and where they came from :smile:

steamtinkerlaunch
```diff
@@ -3362,6 +3363,7 @@ function setDefaultCfgValues {
 		if [ -z "$USESPEKD3D47" ]						; then  USESPEKD3D47="1"; fi
 		if [ -z "$SDLUSEWAYLAND" ]						; then  SDLUSEWAYLAND="0"; fi
 		if [ -z "$STLRAD_PFTST" ]                       ; then  STLRAD_PFTST="none"; fi
+		if [ -z "$STEAMDECK_AUTOUP" ]                   ; then  STEAMDECK_AUTOUP="1"; fi
```
sonic2kk: Should we set this to `1` (enabled) by default? I don't remember if we wanted that behaviour or not, I'm not _super_ bothered either way but to me I would probably expect this to be off by default. However if you think differently please shout!

Note that if you test this in future and want to change this locally, you will need to remove the value from your `~/.config/steamtinkerlaunch/global.conf` and then bump the version. Values are only added when the version changes, and values are not automatically changed on version changes either; if we need to change a value after a release, we have `migrateCfgOption` to do so.

steamtinkerlaunch
```diff
@@ -5709,6 +5713,7 @@ function AllSettingsEntriesDummyFunction {
 --field="     $GUI_RUNSBS!$DESC_RUNSBS ('RUNSBS')":CHK "${RUNSBS/#-/ -}" `#CAT_Misc` `#SUB_Checkbox` `#MENU_GAME` \
 --field="     $GUI_SDLUSEWAYLAND!$DESC_SDLUSEWAYLAND ('SDLUSEWAYLAND')":CHK "${SDLUSEWAYLAND/#-/ -}" `#CAT_Misc` `#SUB_Checkbox` `#MENU_GAME` \
 --field="     $GUI_STLRAD_PFTST!$DESC_STLRAD_PFTST ('STLRAD_PFTST')":CBE "$(cleanDropDown "${STLRAD_PFTST/#-/ -}" "none!gpl!sam!rt!emulate_rt!rtwave64!video_decode")"  `#CAT_Misc` `#MENU_GAME` \
+--field="     $GUI_STEAMDECK_AUTOUP!$DESC_STEAMDECK_AUTOUP ('STEAMDECK_AUTOUP')":CHK "${STEAMDECK_AUTOUP/#-/ -}" `#CAT_Misc` `#SUB_Checkbox` `#MENU_GLOBAL` \
```
sonic2kk: The actual code here is absolutely fine, it's just the `GUI_STEAMDECK_AUTOUP` and `DESC_STEAMDECK_AUTOUP` need to be added to the langfiles. For initial testing you can just put it in `lang/english.txt`, until the wording is finalised, and then copypaste it into the other files afterwards (I want to make a script for it at some point, to sync langfiles, I've just been lazy.)

Don't see any issues apart from that though. It won't crash either, but the checkboxes would show up empty.

steamtinkerlaunch
```diff
@@ -26190,6 +26204,22 @@ function restoreGtkCss {
 	fi
 }
 
+function checkSteamDeckLastVers {
+	# This function just makes sure that the 'lastvers' file exists at the defined path
+	# If it does not we set it and clear the deps to ensure it starts downloading the libs at the right version
+	if ! [ -f "$STLSTEAMDECKLASTVERS" ]; then
+		clearDeckDeps
```
sonic2kk: Could this potentially cause problems? Just thinking out loud.

If we clear the dependencies and then set `STLSTEAMDECKLASTVERS` to the current version, wouldn't this cause the `checkSteamDeckSTLUpdated` check in `checkSteamDeckDependencies` to fail, and so we'd end up with no dependencies?

I _think_ the `-f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK"` would catch this, and since the files don't exist we would fall through to downloading the dependencies. But it's just something I wanted to call out, it might be good to validate this in testing!

sonic2kk:
I left a couple of comments, but overall this makes sense to me. This is very well implemented from reading and following the code. If might see just how difficult it would be to set up a SteamOS VM to test this, because I'd like to give this the proper testing and review it deserves.

I appreciate being able to follow the commits for the new changes too. The only one I left a comment about really was the `clearDeckDeps` addition. I think it's fine but I wanted to call out a small concern I had with it, because that might be good to double-check in testing. While writing I kinda went back and forth if it was worth noting but in the end I thought, better to say it than not :-)

Overall, freaking excellent job :+1: I think once we resolve the open comments this is good to start testing. I don't really remember what the problems were with a SteamOS VM that others reported but I guess there's only one way to find out the problem and that's to tinker around with it for myself :smile:

steamtinkerlaunch
AtomHare: Since this autoupdate feature is only for the Steam Deck it should be enabled by default because I see it like that :
we add a new feature that **requires** the new libs update, we want the users to automatically get the right lib if they update STL altogether. I think enabling it by default will prevent breakages. And it's possible to disable it for *power users* because it's not the expected behavior of STL.

steamtinkerlaunch
AtomHare: You're right, thanks for pointing it out ! I was looking at the code and didn't know where all those variables were actually being set !

steamtinkerlaunch
AtomHare: Normally it should be covered by the `-f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK"` check since it will fail because no libs will be available. It's currently how it already behaves.

I also thought about internet connection issues. However, since the clearDeps should **ONLY** happen after a fresh installation or an STL update internet should already be there to download the STL update in the first place.

steamtinkerlaunch
AtomHare: I just fixed it, thanks a lot for the whole explanation about bash scripting. I'm very new to this, it's way more clear now ! 

> I also changed it to an -eq 1 check, because we want to make sure the value is enabled.

It still should be `( "$( checkSteamDeckSTLUpdated )" -eq 0 || "$STEAMDECK_AUTOUP" -eq 0 )`, we don't want to update if either `lastver==current_ver` **OR** the autoupdater isn't enabled (since it's an OR equation it will bypass the `lastver==current_ver` when it mismatches)

AtomHare:
Thank you a lot for all your explanations on the different part of the code. I will start testing on an SteamOS VM very soon. I found [this guide](https://i12bretro.github.io/tutorials/0857.html) which I will use as a base to try the installation !

steamtinkerlaunch
sonic2kk: Ah, yeah, makes sense. When a user updates we want those libs to get pulled by default.

> And it's possible to disable it for _power users_ because it's not the expected behavior of STL.

Agreed, although, **SteamTinkerLaunch is for power users only**, so everyone using it should know what they're doing, or they should be prepared to mess something up :-)

sonic2kk:
Resolved all the comments, thanks for your explanations.

Also thanks for the guide. I wonder if it works without VirtualBox. It doesn't seem like any of the steps here are specific to it (the Guest Additions can probably be replaced with Spice). Most of it seems like general Arch installation steps with a little extra flavour to set the default boot session and such :-)

lang/english.txt
```diff
@@ -1125,6 +1125,8 @@ GUI_SGASETACTION="Artwork Set Method"
 DESC_SGASETACTION="Specifies how to set the artwork, either by copying the source files, symlinking them, or moving them - If unsure, leave at default 'copy'"
 GUI_STLRAD_PFTST="RADV Perfest options"
 DESC_STLRAD_PFTST="A comma-separated list of named flags, which can be used to enable experimental driver features and performance enhancements for RADV Vulkan driver"
+GUI_STEAMDECK_AUTOUP="Autoupdate STL libraries [SteamDeck ONLY]"
```
sonic2kk: Real nitpick here: Should probably be `Steam Deck` (with the space).

steamtinkerlaunch
sonic2kk: ShellCheck is happy now, and your clarification makes sense (resolved and forgot to leave the reply, apologies).

steamtinkerlaunch
AtomHare: > Agreed, although, **SteamTinkerLaunch is for power users only**, so everyone using it should know what they're doing, or they should be prepared to mess something up :-)

Yea that's why I put it in italic, but if we can avoid some bug reports and issues it's always a welcomed default !

lolcatplays:
> Thank you a lot for all your explanations on the different part of the code. I will start testing on an SteamOS VM very soon. I found [this guide](https://i12bretro.github.io/tutorials/0857.html) which I will use as a base to try the installation !

I have tested this on my steam deck but so far seems to not be working. I added the yad URL as listed in the master commit as it downloaded a 0B file but I'm still getting the update spam. On the "Main" (bleeding-edge branch) of SteamOS. 

```
steamtinkerlaunch vortex winetricks
Downloading latest SteamTinkerLaunch

Installing dependency 'innoextract'
Downloading dependency 'innoextract-1.9-8-x86_64.pkg.tar.zst'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  214k  100  214k    0     0   564k      0 --:--:-- --:--:-- --:--:--  563k
Successfully downloaded innoextract-1.9-8-x86_64.pkg.tar.zst!
Extracting dependency 'innoextract-1.9-8-x86_64.pkg.tar.zst'...
Successfully extracted 'innoextract-1.9-8-x86_64.pkg.tar.zst' to '/home/deck/stl/deps'
Dependency 'cabextract' already installed, nothing to do.
/home/deck/stl/prefix/steamtinkerlaunch: line 26219: [: : integer expression expected
Dependency 'yad' already installed, nothing to do.

Finished updating SteamTinkerLaunch ('v14.0.20240601-1 (steamdeck-dependency-handling)')!


```

AtomHare:
Thank you a lot for your testing. I rebased to have the Yad url fix in this PR. 
it looks like it failed the `checkSteamDeckSTLUpdated` check there https://github.com/sonic2kk/steamtinkerlaunch/blob/f699a630d47fd96bdf299dec47d011ba65b6b6bf/steamtinkerlaunch#L26217. 

Not sure what went wrong, I will test it on my steam deck tonight too and see how I can prevent notification spam.

sonic2kk:
Since we're checking if it's equal to one, perhaps we can simply do this:

```bash
function updateSteamDeckLastVers {
	# This function updates the 'lastvers' file after a dependency update if there was a version change
	if checkSteamDeckSTLUpdated ; then
		echo "$PROGVERS" > "$STLSTEAMDECKLASTVERS"
	fi
}
```

This is how we use, for example, [`checkSGDbApi`](https://github.com/sonic2kk/steamtinkerlaunch/blob/2d0b0ec464f5deb3c4f36a92fd6ad7334710b367/steamtinkerlaunch#L1402-L1410) and, more recently, [`haveAnySteamShortcuts`](https://github.com/sonic2kk/steamtinkerlaunch/blob/2d0b0ec464f5deb3c4f36a92fd6ad7334710b367/steamtinkerlaunch#L8316-L8322). We'd need to rework how `updateSteamDeckLastVers` is checked in any other places we use it. Also, we'd need to check the return for `checkSteamDeckSTLUpdated` as well. Returning `0` is a success status, so if STL updated, we'd want `checkSteamDeckSTLUpdated` to return `0`.

The `return` didn't work because using a subshell with `$( func_name )` will only capture the echo/printf/etc, I guess. Which I'm a bit surprised about to be honest.

I'll do a scan of the PR later to see if we need to fix this up in other places. This is my bad for not realising, I'm too used to convenience in other languages 😅 Both in not catching it in this PR and not capturing it when I prototyped this logic before.

(I didn't run the workflow again because afaict the last force-push was just a rebase, so I'll run the workflow again when this is marked ready for review.)

steamtinkerlaunch
```diff
@@ -26343,6 +26373,21 @@ function getSteamDeckCompatInfo {
...
 
+# This function will check if the version in `$STLSTEAMDECKLASTVERS` does not match the current version
+# For example, if we had v14.0 installed before, 'lastvers' would have v14.0
+# Then, if we were updating to v15.0, 'lastvers' would be 'v14.0' but 'PROGVERS' would be 'v15.0'
+function checkSteamDeckSTLUpdated {
+	# This is how updateCfgFile gets the config file version, so re-use it
+	CHECKLASTVERSSTEAMDECK="$( cat "$STLSTEAMDECKLASTVERS" )"
+	if [ "$CHECKLASTVERSSTEAMDECK" = "$PROGVERS" ]; then
+		writelog "INFO" "${FUNCNAME[0]} - Last known SteamTinkerLaunch install version ('$CHECKLASTVERSSTEAMDECK') and the current version ('$PROGVERS') match -- There has been no update since last launch"
+		return 0
```
sonic2kk:
```suggestion
		return 1
```

steamtinkerlaunch
```diff
@@ -26343,6 +26373,21 @@ function getSteamDeckCompatInfo {
...
+	else
+		writelog "INFO" "${FUNCNAME[0]} - Last known SteamTinkerLaunch install version ('$CHECKLASTVERSSTEAMDECK') and the current version ('$PROGVERS') do NOT match -- It seems there has been an update!"
+		return 1
```
sonic2kk: We need to `return 0` because `0` is a success code.

```suggestion
		return 0
```

steamtinkerlaunch
```diff
@@ -25924,7 +25929,13 @@ function checkSteamDeckDependencies {
...
 		CHECKCMD="$($DEPCMD &> /dev/null --version && echo "OK" || echo "NOK")"
 
-		if [ -f "$(command -v "$DEPCMD")" ] && [ "$CHECKCMD" = "OK" ]; then
+
+		# The check now says, don't update dependencies if all of these conditions are true:
+		#  1. The dependency file exists
+		#  2. The dependency can actually be ran, confirming it is a valid file
+		#  3. SteamTinkerLaunch has not updated or autoupdater isn't enabled, so there would be no change in dependency version and thus no need to update
+		# If any of these are false, we need to check our dependencies (if a file is missing we would need to update, or if it cannot be used we need to update, and also if STL updated we may need a newer version, so update).
+		if [[ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK" && ( "$( checkSteamDeckSTLUpdated )" -eq 0 || "$STEAMDECK_AUTOUP" -eq 0 ) ]]; then
```
sonic2kk: With the changes  to how `checkSteamDeckSTLUpdated` works, it'll return `0` (positive, meaning an update occurred), so we can use it like a true boolean value. Note how it can be used without square brackets, because it's not a `test` condition (Bash is weird).

```suggestion
		if [[ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK" && ( checkSteamDeckSTLUpdated || "$STEAMDECK_AUTOUP" -eq 0 ) ]]; then
```

ShellCheck also shouldn't complain about this, with the code below at least it was happy.

<hr>

I verified this sort of logic works with a little test script, hopefully it helps illustrate the logic a little better :-)

```bash
#!/usr/bin/bash

PROGVERS="1.2.3"

function hasUpdated {
    oldvers="1..3"
    
    if [ "${oldvers}" = "${PROGVERS}" ]; then
        # no update
        return 1
    else
        return 0
    fi
}

boolvar=1  # Set this to 0 to get "no update"
if ( hasUpdated && [ "${boolvar}" -eq 1 ] ); then
    echo "Has updated"
else
    echo "no update"
fi
```

steamtinkerlaunch
```diff
@@ -26191,6 +26205,22 @@ function restoreGtkCss {
...
+	# If it does not we set it and clear the deps to ensure it starts downloading the libs at the right version
+	if ! [ -f "$STLSTEAMDECKLASTVERS" ]; then
+		clearDeckDeps
+		echo "$PROGVERS" > "$STLSTEAMDECKLASTVERS"
+	fi
+}
+
+function updateSteamDeckLastVers {
+	# This function updates the 'lastvers' file after a dependency update if there was a version change
+	if [ "$( checkSteamDeckSTLUpdated )" -eq 1 ]; then
```
sonic2kk: Same idea here with how `checkSteamDeckSTLUpdated` is used:

```suggestion
	if checkSteamDeckSTLUpdated; then
```

sonic2kk (commented):
Should've caught this before, very sorry I only realised these now :sweat: 

Give these a try locally and see if they pan out. If so push them up and the update checks should be fine :-)

steamtinkerlaunch
sonic2kk: Was about to fall asleep and it just dawned on me that this should probably be a `! checkSteamDeckSTLUpdated` check, since before we were checking `-eq 0`. 

I think there is some weirdness around how Bash treats the `!` negation, so please test if you can to make sure the negation doesn't affect the whole statement and only that `checkSteamDeckSTLUpdated` check.

I'll try do some testing in a test script tomorrow to see how Bash handles this. I don't know if there's any examples in the codebase of this kind of check in a compound `if` check.

AtomHare: I just tried this solution and came across an issue, it doesn't look like the return value of a function is supported inside a test `[[`. See https://stackoverflow.com/a/42971383 
I'm not sure how we should tweak it, one idea could be to "preload" the function response before the answer output

sonic2kk: > I just tried this solution and came across an issue, it doesn't look like the return value of a function is supported inside a test

That's a bit strange, is the code snippet I provided invalid then? It seemed to work in my testing...

Plus, we use `if funcThatReturnsZero ; then` syntax around the codebase already.

AtomHare:
I just tried your suggestions, we're close to making it work but it looks like we need to make some tweaking !

And I noticed other things:
log spam still occurs: "preparing to install SteamTinkerLaunch", "Downloading latest SteamTinkerLaunch", "Add steamtinkerlaunch as compatibility tool", "Finished updating SteamTinkerLaunch". Maybe if the `lastvers` file is detected, we could remove all of those notifications ?

Recap of the remaining issues :

- [x] Tweak the `if` formula
- [x] ~~Prevent the log spam~~ -> in a separate PR
- [ ] Avoid redownloading the libs on each update if they don't change, (it looks like it redownloads the libs, even though they were downloaded before?)

steamtinkerlaunch
AtomHare: It's not invalid, in the current `if` it's more like `if [[ hasUpdated && ( "${boolvar}" -eq 1 ) ]]` and I don't know how to properly wrap the `or` to override the update if they're disabled

sonic2kk:
> Maybe if the `lastvers` file is detected, we could remove all of those notifications ?

Yes, we should update notifier logic I think, but I think all the remaining notifier-related "problems" can be done in a separate PR.
- Notifier for installing SteamTinkerLaunch, including finishing installation
    - We should keep this, because when a user double-clicks the script to install STL we will want to show this visual feedback. If it is not running only during installation, we should fix that, but it can go in a separate PR, nothing to worry about here :-)
- Notifier for updating SteamTinkerLaunch, including finishing updating
    - We can remove the notifiers around preparing to install SteamTinkerLaunch, as the auto-update logic (which doesn't really work) will be removed shortly. Users can just manually update STL, and when they do, this PR will update the dependencies automatically!
- Notifier for adding SteamTinkerLaunch as a compatibility tool
    - This should be reworked to only add SteamTinkerLaunch as a compatibility tool if it isn't already there. We should move the notifier logic into this function, wrapped in Steam Deck-specific checks: https://github.com/sonic2kk/steamtinkerlaunch/blob/bddc3b36369632ef6bcaeba36eddf05c552dae65/steamtinkerlaunch#L21800 -- So we notify if the symlink is created/updated only, and skipped when the symlink already exists.

So we can remove two notifiers related to automatic updates once that logic is stripped out. If the notifier logic for installing and finished installing is incorrect, that should be fixed to only show on initial installation (i.e. when the script is executed to install STL initially, or when installed from ProtonUp-Qt which just runs `./steamtinkerlaunch` as well). Finally, the notifier for adding SteamTinkerLaunch as a compatibility tool should be updated to be inside of the `CompatTool` function and should only run when the compat tool symlink is added/updated. And all of these changes can go in separate PRs, so no need to stress about them here. You're already working on enough in this PR. If you'd like though you can contribute to improving the notifier in a future PR!

Users that are bothered by the notification spam are probably not users who STL is aimed at, they tend to be the non-developers that bought Steam Decks in my experience. And STL on Steam Deck is mainly an enthusiast/developer curiosity, not something to be used seriously in its current form, hence the wiki's emphasis that SteamTinkerLaunch on SteamOS is in early stages, yet you are one of the few devs who actually stood up to contribute, there aren't many contributors on the desktop side and virtually none for SteamOS, so it's nice to see someone that cares enough to actually improve the situation. So to that end feel free to pick up the notifier work in a future PR, but there is absolutely no obligation on you to do so in this PR (it is probably cleaner in a separate PR anyway).

<hr>

~~I will take a look at the `if` check and your comment above when I have some time, I am curious as to why it isn't working properly myself :-)~~

**EDIT**: I left a comment about it, we can discuss the `if` formula further there. I probably shouldn't have even mentioned it here.

>Avoid redownloading the libs on each update if they don't change, (it looks like it redownloads the libs, even though they were downloaded before?)

Hmm, something is going wrong in `installDependencyVersionFromURL` then? I'm not sure why from looking at the code...

It might be worth checking the `steamtinkerlaunch.log` file to see if there's any indication of where things are going sideways.

steamtinkerlaunch
sonic2kk: Would this work?

```bash
if [[ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK" ]] && ( checkSteamDeckSTLUpdated || [[ "$STEAMDECK_AUTOUP" -eq 0 ]] ); then
    # ...
fi
```

I'm playing around with this in a test script. The logic is a little complex here, but I think this should work, since `( checkSteamDeckSTLUpdated )` should capture the return, and we wrap the check on `$STEAMDECK_AUTOUP` in a separate `[[` test.

<hr>

In my tests with the above:
- If `DEPCMD` fails, we will need to update.
- If `CHECKCMD` fails even if `DEPCMD` passes, we need to update.
- If `checkSteamDeckSTLUpdated` returns `1` (meaning no update was found), we will only need to update if
    - `DEPCMD` fails
    - `CHECKCMD` fails
    - Both `DEPCMD` and `CHECKCMD` fails
    - Or if `STEAMDECK_AUTOUP` is `1` (meaning we want to always check for updates).
- If `STEAMDECK_AUTOUP` is `1`, we will always need to update regardless of `DEPCMD`, `CHECKCMD`, and `checkSteamDeckSTLUpdated`

I think that logic is correct, but I might be missing something obvious :sweat_smile:

AtomHare:
Thank you for your comments, sadly I won't be able to look and tinker with the code until Monday but will as soon as I'm available!

I will also make a separate PR for the steam deck notifications.

And I may look into a debug var that enables the steam deck behavior to make it easier to test. I don't need it to run the libs but mainly to test the codes/logic! (I'll probably add a dev check to prevent people wanting the steam deck behavior in desktop mode)

sonic2kk:
No worries, you can work on this whenever you have time. All your work is appreciated thus far as well! :-)

sonic2kk:
I went ahead and linked this PR to #859, as this is the last thing discussed in that issue related to Steam Deck improvements.

You've expressed interest in doing more work, and you are *absolutely welcome* to contribute any improvements in the future. But as far as that issue goes, this PR would implement all that was left specific to that issue. So once this is merged, that issue would be done, even if there is more work you or anyone else may want to do!

Of course that doesn't mean there is any rush on this either. I've just been doing a little "issue cleanup" and this was part of it. You can work on this whenever you have time and motivation. If we hit any major blockers in a worst-case and we can't merge this (which I don't foresee happening, imo this is already most of the way there, once again fantastic work on your part), linking the issues gives a good paper-trail on what went on during development. Development transparency is very important to me with this project and giving everyone insight into what decisions were made, where discussion took place, etc is something I strive for (even if I don't always live up to that in the way I want to).

So yeah, please don't feel any additional pressure here. Just some repo management :smile:

steamtinkerlaunch
AtomHare: Let me write a truth table, I think it will be more clear for the `If STEAMDECK_AUTOUP is 1, we will always need to update regardless of DEPCMD, CHECKCMD, and checkSteamDeckSTLUpdated` part.

Variables | `DEPCMD` | `CHECKCMD` | `checkSteamDeckSTLUpdated (TRUE=0 if update)` | `STEAMDECK_AUTOUP` | Do we update ? | boolean value
--- | --- | --- | --- |--- |--- |---
| 0 | 0 | 0 | 0 | 0 | YES | 0 |
| 1 | 0 | 0 | 0 | 1 | YES | 0 |
| 2 | 0 | 0 | 1 | 0 | YES | 0 |
| 3 | 0 | 0 | 1 | 1 | YES | 0 |
| 4 | 0 | 1 | 0 | 0 | YES | 0 |
| 5 | 0 | 1 | 0 | 1 | YES | 0 |
| 6 | 0 | 1 | 1 | 0 | YES | 0 |
| 7 | 0 | 1 | 1 | 1 | YES | 0 |
| 8 | 1 | 0 | 0 | 0 | YES | 0 |
| 9 | 1 | 0 | 0 | 1 | YES | 0 |
| 10 | 1 | 0 | 1 | 0 | YES | 0 |
| 11 | 1 | 0 | 1 | 1 | YES | 0 |
| 12 | 1 | 1 | 0 | 0 | NO (no update and autoupdater not enabled) | 1 |
| 13 | 1 | 1 | 0 | 1 | NO (no update and autoupdater enabled) | 1 |
| 14 | 1 | 1 | 1 | 0 | NO (there's an update but autoupdater isn't enabled) | 1 |
| 15 | 1 | 1 | 1 | 1 | YES (there's an update and autoupdater enabled)  | 0 |

Notice how the 12 first lines basically describe the initial `if`, we don't want to tamper those because tampering those would mean we could be in a case where the user has no libs anymore or non-working ones.
The autoupdate strategy appears at the end with the new additional `AND`. 

I think I made a mistake initially, the right thing to add is `AND not ( checkSteamDeckSTLUpdated && STEAMDECK_AUTOUP -eq 1)` instead which looks equivalent to `( ! checkSteamDeckSTLUpdated ||  STEAMDECK_AUTOUP -eq 0)` . Bash logical operators and 0 doesn't help

Feel free to correct me if I made a mistake

AtomHare: It looks like the error we got before :  `steamtinkerlaunch: line 25938: [: : integer expression expected` comes from the fact that the new env variable `STEAMDECK_AUTOUP` is not populated inside `global.conf` after an update. ~~How would I fix that ?~~ It looks like i edited `setDefaultCfgValuesdefault_template` instead of `setDefaultCfgValuesglobal` :man_facepalming: 

The if line seems to work with `if [[ -f "$(command -v "$DEPCMD")" && "$CHECKCMD" = "OK" ]] && ! ( checkSteamDeckSTLUpdated && [ "$STEAMDECK_AUTOUP" -eq 1 ] ); then` :tada:

AtomHare:
No issue at all, thanks a lot for your help ! 

I've tested the PR and it looks ready for review !

steamtinkerlaunch
sonic2kk: The config file values get updated each time there is a version bump, so if you make a new field the version needs changed. It doesn't need to be exact, the string just needs to change to something unique I believe :-)

sonic2kk:
Thank you! I will look into this PR. I am not sure how much time I will have next week as it is unusually busy but I'm grateful for all your work.

steamtinkerlaunch
sonic2kk: That truth table is really awesome by the way. Makes the reasoning for the conditional you came up with a lot more sense. I think this comment is resolved!

AtomHare:
You're welcome, thank you for your continuous help with the PR ! Yeah, no problem, there's no hurry !
I'll try to make some followup PRs to continue improving the Steam Deck and overall experience too

steamtinkerlaunch
```diff
@@ -4096,6 +4098,8 @@ function saveCfg {
 			echo "SDLUSEWAYLAND=\"$SDLUSEWAYLAND\""
 			echo "## $DESC_STLRAD_PFTST"
 			echo "STLRAD_PFTST=\"$STLRAD_PFTST\""
+			echo "## $DESC_STEAMDECK_AUTOUP"
```
sonic2kk: This puts the option in the Per-Game Config File and not the Global Config. It should be moved to `saveCfgglobal`. Then the version will need bumped for the change to take effect I believe.

[This PR](https://github.com/sonic2kk/steamtinkerlaunch/pull/935/files#diff-77c6d4bb1bf7c1988ff5a068c856b6c24e113663f2e61cb3a2c98723c0fafc04R3637-R3640) might serve as a good reference for adding a global config value. I should really add comments for this too now that I've thought about it :-)

I didn't test this on SteamOS yet btw, just on my PC.

sonic2kk:
~~Just a heads up; As per #1135, it seems some dependencies are now broken on SteamOS. That issue is specifically where Game Mode can't find extract archives properly. If you can replicate the issue it may be a bug in SteamOS that could block this PR for a bit. 😦~~

That issue turned out to be the innoextract version needing bumped again for SteamOS 3.6 Preview which updates the Arch snapshot. False alarm! 😄

sonic2kk:
Hello, it has been a while since the last commit. Did you have a chance to investigate the suggested change, and test this out on SteamOS? From what I remember aside from the PR comment I left this PR was in good shape and really just needed that change and testing.

To make testing easier, I had a thought of introducing a way to force the SteamOS code path in SteamTinkerLaunch in a debug mode of sorts. This would be intended to run on, say, a general Linux VM (so as not to disturb any existing SteamTinkerLaunch installation configuration). I have not gotten around to this yet and there is no ETA but if you have not gotten around to testing and don't wish to test on your own device (which is understandable) then we can park this PR for a little bit pending that change.

I am not in any hurry for a v14.0 release, there's still some loose ends on my side to tie up that will take a while, but this PR is important enough that I would wait for it to be merged anyway. :slightly_smiling_face:

AtomHare:
Hey, sorry, I've been quite busy lately and wasn't much on GitHub. I don't mind testing on my own device, that's no issue. 

About this PR, everything works fine (and is tested by editing manually the conf).  I only need to edit the config save to point to the global conf instead of the game one. 

I also thought of having a special env variable that we could use to force the STL steam deck behavior to make tests easier. I might take a look at it soon !

sonic2kk:
No worries, real life always comes first! As I said there is no major hurry or anything.

> About this PR, everything works fine (and is tested by editing manually the conf). I only need to edit the config save to point to the global conf instead of the game one.

Awesome news! This will be a very welcome improvement I am sure by all Steam Deck users.

> I also thought of having a special env variable that we could use to force the STL steam deck behavior to make tests easier. I might take a look at it soon !

I haven't gotten around to it yet so by all means take a look! If nothing materializes, it's no problem either :-)

AtomHare:
Finally fixed the place where the config was saved, sorry for the delay !
The PR is rebased too, it's now ready to test !

The only untested part (on my end) of this PR is not the behavior of the autoupdater logic but mainly if the config gets updated or not and if the global settings works too.

steamtinkerlaunch
AtomHare: Nice catch ! Fixed it in my last commit :)

sonic2kk:
Checked out the PR locally on my laptop and `STEAMDECK_AUTOUP` is created in the right place with the expected default value of `1`. As far as I'm concerned, the global setting is being created as intended.

I will investigate testing this on SteamOS, on hardware or otherwise, in particular for testing the global setting, but feel free to test if you have time and are willing as well!

<hr>

Thanks a ton for your hard work on this. I don't envision any major problems from looking at the code, the remaining work for this PR is really just a testing effort to confirm that things are fine, and updating the rest of the langfiles (and before merging we can bump the version to exclude the branch name :smile:). I will also test this on my laptop and PC just to make 100% sure this doesn't impact anything outside of SteamOS.

If you're happy with the wording of the new strings in `english.txt`, feel free to copy and paste those into the other langfiles.

I did notice that for some reason with this PR, Yad is using XWayland, but I don't think that's related to this PR and is probably just some Yad weirdness, I'll double-check to see what is up there though just in case.

<hr>

Once again, super excellent work, I'm really looking forward to seeing this merged!! I don't see any remaining blockers other than testing and making any fixes needed, and if any fixes are needed they're probably not going to require major refactors, I think the changes here is sound.

sonic2kk:
As a brief follow-up, I confirmed the XWayland weirdness I was seeing is entirely unrelated to this PR, so no worries there :-)

sonic2kk:
If you have tested this on your own Steam Deck and are happy that it works, I think this can be merged. It does not interrupt any of the Linux Desktop usage, and the code looks good to me. I don't have any way to test SteamTinkerLaunch on Steam Deck right now (I would really prefer to keep SteamTinkerLaunch off my OLED, and I don't have access to the LCD Steam Deck that I had gotten permission to test on right now).

I have left this in limbo for far longer than I would've liked and I feel bad about it, so if you're happy that it works, I can merge this. I am planning to note that Steam Deck is not actively supported prior to v14.0, so there won't be any responsibility or expectation here. The work here looks solid and I appreciate all the effort, and if you are happy that it works, I'll get this merged :-)

AtomHare:
Sorry for the late reply, the past month has been quite busy.

I consider this PR finished and everything should work as expected (Considering the only thing untested was the wrongly placed configuration).

However, I should be able to test in two weeks to verify the PR state !

P.S: Don't worry about the time scale of this PR, I lost a bit of momentum + life work, no issue at all regarding this !!

AtomHare:
I have left this PR hanging on longer that I'd have liked to. However, I consider it ready and tested ! 
The previously untested part (global.conf) works and successfully migrated my config and updated the libs on my steamdeck !!

sonic2kk:
Apologies! I see you left that comment 3 days ago. From my side, I am happy with your testing and am glad to merge this. I will take a final once-over and try to make time to merge before the weekend :smile: 

Absolutely stellar work here, I cannot thank you enough :-)

AtomHare:
Thank you a lot for your guidance during the PR. And no issues take your time, this PR is in no rush !

sonic2kk:
[Changelog has been updated to give credit](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog/_compare/bfacc0497379039893515e960407e65660aa1a4f). Please let me know if you think more should be described or added. Your work here will be appreciated by many :-) Heck, maybe I'll even use SteamTinkerLaunch on my Steam Deck again thanks to this!

AtomHare:
Thank you a lot! I don't know if it should be considered as an auto updater as is. This "autoupdater" basically nukes the libs if there was an STL update to avoid any issues from happening on the libs side. Ideally a lib vers checker would be needed and we would need to log the libs version inside a file.

PR #1119, opened 2024-06-05
Add Non-Steam Game: Set AllowOverlay and OpenVR in `localconfig.vdf`

### LINKED ISSUE #960: [Feature Request] Non-Steam Game Improvements

sonic2kk:
Work for #960.

## Overview
This PR fixes setting the AllowOverlay and OpenVR flags when adding Non-Steam Games. Previously these values were set and read by the Steam Client from `shortcuts.vdf`, but now it reads them from `localconfig.vdf`, under the `"Apps"` block, in a nested block within that based on the **32bit signed AppID** for the Non-Steam Game.

This PR implements functionality to write these two values to this section. Some logic for other VDF functions had to be touched to make this PR possible. I have tested that the logic for adding to `config.vdf` still works, and for getting the current global compatibility tool from `config.vdf` also still works. But we should do extra testing to confirm no breakages.

## Implementation
We write these values at the end of `addNonSteamGame`. To write these values properly, we need to make sure that the `"Apps"` section exists first, and if this section has the section for the Non-Steam Game (identified by its 32bit unsigned AppID). Then we have to write the values for AllowOverlay and OpenVR into this file if they aren't already there. If they are already there, we need to update their values.

### `"Apps"` Block
This `"Apps"` block is always written to the end of `localconfig.vdf`, although the Steam Client may move them. There is at least one other block in this `localconfig.vdf` with the name `"Apps"`, so we explicitly check for the block with one indentation level. This block holds information for any and all Steam games, including Non-Steam Games, that set some properties, but the only relevant one for STL adding a Non-Steam Game are toggling the Steam Overlay and OpenVR support.

### Per-Steam Game Block
The AllowOverlay and OpenVR keys are `OverlayAppEnable` and `DisableLaunchInVR` respectively. It's worth noting the negative `DisableLaunchInVR` key. On the STL side we check if OpenVR is _enabled_, but Steam does a negative check and checks if it is _disabled_. When writing our value to `DisableLaunchInVR`, we have to invert it.

These keys are stored under a block in the `"Apps"` section. For regular Steam Games, this is just the AppID. However for Non-Steam Games, for some reason this explicitly uses the 32bit Signed AppID, i.e. `"-12345678"`, as opposed to the 32bit Unsigned AppID (used for Steam Artwork names in the `grids` folder, in `CompatToolMapping` in `config.vdf`, etc).

This section could exist under a different block, so we make sure that it exists nested inside of the `"Apps"` section using `getNestedVdfSection`.

**Note**: Block names at the same indentation level **must be unique** in the VDF format.

### Walkthrough of Logic Example
We insert the values into this block in `localconfig.vdf` using a new function, `updateLocalConfigAppsValue`. This handles creating the sections we want if they don't exist, and _creating OR updating_ the values we want to insert into the AppID block.

Here is a written example of how `updateLocalConfigAppsValue` works. We will assume the 32bit signed AppID is `"-12345678"` in this example.
1. Does the `"Apps"` section exist in the file at the given indentation level?
    a. If yes, continue.
    b. If no, create it.
2. Does the `"-12345678"` section exist nested inside of this `"Apps"` section?
    a. If yes, continue.
    b. If no, create it inside of this `"Apps"` block.
3. Does the given key/value pair we want to insert into this file already exist?
    a. If yes, update it with the new value.
    b. If no, create it.

<hr>

This functionality was tested extensively using `steamtinkerlaunch debug` and a lot of temporary logic to do so. It was then transferred into `updateLocalConfigAppsValue`. It took quite a long time to figure out but I have tested various scenarios, including testing `config.vdf` to check for breakages, and this seems to all still work fine.

However, extra testing should be done, so I will not merge this for a while.

Figuring out how to do this was the most complex part. I had the steps figured out, but working out how to implement it with Bash and our current VDF functions, and what changes were needed to fix some incorrect behaviours with these functions, took months. But I am reasonably confident that this should work.

<hr>

TODO:
- [x] ShellCheck
- [x] More code cleanup
- [x] Version bump


sonic2kk:
There are some warning comments, such as around `getVdfSection`. I will remove these before merging.

sonic2kk:
Tested adding a few games from the GUI and commandline. Setting the compatibility tool still worked, as did downloading fresh artwork from SteamGridDB, and of course the AllowOverlay and OpenVR options now work!

This PR just needs a little bit of cleanup and it should be good to go. I'm not finding any breakages in testing.

sonic2kk:
Pending ShellCheck passing, this is ready to merge.

PR #1121, opened 2024-06-07
Non-Steam Game: Add `non-steam|nsg` filter to `list` Command

### LINKED ISSUE #960: [Feature Request] Non-Steam Game Improvements

sonic2kk:
Work for #960.

## Overview
This PR adds a `non-steam` filter (with `nsg` shorthand) to the `steamtinkerlaunch list` command. Now all Non-Steam Games in the library can be listed with the `list` command as AppID, name, path, count, or all of the above.

The usage is exactly the same as it is for using `owned` (`o`) or `installed` (`i`), but now we can use `non-steam` (`nsg`) with this command too.

Non-Steam Games are NOT printed alongside owned/installed AppIDs, they are ONLY printed when the `non-steam`/`nsg` filter is used.

As an aside, this also fixes a small formatting bug where `printf` wasn't taking a newline properly, which resulted in some ugly output, and the function has been flattened slightly.

## Implementation
We follow the same pattern as we do for `installed` and `owned`, and start by getting all of the Non-Steam Game AppIDs and then storing this in an `LISTAIDSARR` array. To do this, we have a new function to get all Non-Steam AppIDs, named `listNonSteamGameIDs`. This prints each AppID parsed from `shortcuts.vdf` in a `while` read loop the same way we do for `getOwnedAids` and `listInstalledGameIDs`.

From there the rest of the program just falls through the rest of the logic that it already does for `installed` and `owned`.

The regular logic here calls `getTitleFromID` and `getGameDir`, however these two functions have separate parameters to know whether they should search for Non-Steam Games or not. This was implemented because these functions are used in the codebase in situations where Non-Steam Games should not be returned (such as for Vortex game detection checks). Therefore searching on Non-Steam Games is **disabled** by default.

Likewise, we don't want `installed`/`owned` to return Non-Steam Games, so we only want to have these functions return Non-Steam Games when we search for them.

In order to get these functions to only return the Non-Steam Game AppIDs when we search for Non-Steam Games, we have a conditional variable to control this. We only call `listNonSteamGameIDs` if we pass the `non-steam` and `nsg` filter, so if we do this, we set `SEARCHSTEAMSHORTCUTS` equal to `1`. It is `0` by default. This value is given to `getTitleFromID` and `getGameDir`, as they expect `0`/`1` values. This allows us to control when these calls return Non-Steam Games, allowing us to only return Non-Steam Games when we explicitly want to find them.

## Concerns
This implementation is not very efficient in particular for Non-Steam Games. We fetch the list of AppIDs by parsing `shortcuts.vdf`, and then we have to do this all over again for each call to `getTitleFromID` and `getGameDir`.

This is the most inefficient for the default, where we want to get name *and* path for Non-Steam Games. With how the function is currently implemented, I don't see a clean way to refactor this. We are probably best off making separate functions for listing owned, installed, and Non-Steam games at some point in the future, but it's overkill for now.

The performance hit is undesirable but acceptable _for now_ until I find time to re-evaluate `listSteamGames`. If people complain about it being slow, and something downstream depends on it being faster, refactoring will be prioritized.

<hr>

This PR has been tested and it seems to still work fine in my testing with `owned`, `installed`, and now `non-steam` with no regressions or issues.

This will probably be ready to merge soon, pending a version bump. This is one of the last remaining tasks for the Non-Steam Game integration overhaul!

TODO:
- [x] Check if wiki needs updated
- [x] Version bump

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7779,8 +7779,8 @@ function pickGameWindowNameMeta {
 
 # General function for the "steamtinkerlaunch list <arg> function"
 # Can take two types of commands:
-# - `steamtinkerlaunch list owned/installed` - Returns "Game Name (AppID)"
-# - `steamtinkerlaunch list owned/installed id/name/path/full` - Returns either AppID, Game Name, Game Paths, or all in the format "Game Name (AppID) -> /path/to/game"
+# - `steamtinkerlaunch list owned/installed/non-steam` - Returns "Game Name (AppID)"
+# - `steamtinkerlaunch list owned/installed/non-steam id/name/path/full` - Returns either AppID, Game Name, Game Paths, or all in the format "Game Name (AppID) -> /path/to/game"
 #
 # This function is not very efficient, Non-Steam Games in particular are inefficient because we read shortcuts.vdf each time we want to parse info
 # We parse it once to get the IDs, then again in each `getTitleFromID` and `getGameDir` call. This makes it pretty slow
@@ -7849,7 +7849,7 @@ function listSteamGames {
 			done
 		fi
 	elif [ "$LSTYPE" == "count" ]; then
-		printf "\n%s" "$( getGameCount )"
+		printf "\n%s\n" "$( getGameCount )"
 	elif [ "$LSTYPE" == "full" ] || [ -z "$LSTYPE" ]; then  # This is the default if id/name/path/full is not passed
 		for AID in "${LISTAIDSARR[@]}"; do
 			GAMDIR="$( getGameDir "$AID" "" "${SEARCHSTEAMSHORTCUTS}" )"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Confirmed again that this works and doesn't cause any regressions. Commands that may depend on existing functionality like `vortex list-installed` seem to be fine and haven't regressed since adding `${SEARCHSTEAMSHORTCUTS}`.

I'll double-check the wiki and make any needed updates, then bump the version and merge this.

sonic2kk:
Added a small section about this command on the ["Add Non-Steam Game" wiki page](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Add-Non-Steam-Game#listing-all-non-steam-games). I think this plus the help screen update to note this usage is enough.

With a version bump this will be ready to merge.

PR #1122, opened 2024-06-09
misc: Flatten getGameWindowPID

sonic2kk:
Remove a bunch of nesting for readibility.

Needs tested to make sure it still works.


sonic2kk:
Tested reasonably extensively between `master` and here, and compared the logs and meta files. Don't see any regressions. This should be good to merge once the conflicts are resolved and the version is bumped.

Issue #1118, opened 2024-06-04
Vortex doesn't open


Bread-Ch4n:
## System Information
- SteamTinkerLaunch version: 12.12-1
- Distribution: Arch Linux, Kernel: 6.9.3-zen1-1-zen
- Installation Method: aur/yay

## Issue Description
Tried starting vortex but it never opens.
Using the console to launch it all it outputs before opening:
```
steamtinkerlaunch vortex start
sed: -e expression #1, char 0: no previous regular expression
Tue Jun  4 06:23:16 AM EEST 2024 INFO - startVortex - Starting Vortex without options

```

## Logs
steamtinkerlaunch logs when opening vortex
[31337.log](https://github.com/user-attachments/files/15542130/31337.log)

sonic2kk:
All issues from users ignoring the issue template dictating to use the latest commit from master will be closed.

> Feel free to search through currently open issues to see if someone else has already reported your issue, or to see if your issue has already been resolved. If at all possible, try testing the **latest version of SteamTinkerLaunch from the master branch** to see if the issue has already been fixed. You can also check the changelog on the wiki to see if its been fixed since the latest release: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog

The changelog notes extensive changes to Vortex behaviour. Also, checking the GitHub releases would show v12.12 is over a year old. There will eventually be a v14.0 but you should never, never, never, ever use anything except master, pulling and re-installing or rolling back to last known working commits in the event of issue.

Uninstall and re-install Vortex from master. If the issue still persists, feel free to investigate and submit a PR. Vortex works for me on master but I don't recommend using that software and soon hope to remove it from STL in favour of their Vortex app.

Bread-Ch4n:
It weirdly started working today xD

Bread-Ch4n:
Also thanks for making me notice nexus deprecated vortex and made a new c# app

rarelambo21:
what did you do to fix it?

sonic2kk:
They said it just started working in their reply, however, make sure to use SteamTinkerLaunch from master. Users who do not and insist on ignoring all guidance and common sense and using a version from over a year ago despite hundreds of commits and changes since then, should expect problems.

My advice to that user still stands; remove Vortex entirely, use STL from master (you should never use a release unless you have a very good reason), and if the issue persists, dig into the code and logs and see if you can find the issue. If you are on SteamOS, do not use mods, it is a waste of time.

[...]

PR #1123, opened 2024-06-12
Logging: Improve access to log files

sonic2kk:
Based on discussion in #1118.

This PR does a few things:
- Add logging for the Vortex start command, so it can more easily be copied and ran manually for troubleshooting
- Add some SteamTinkerLaunch log paths to the "Game Files" menu:
    - Path to `/dev/shm/steamtinkerlaunch/steamtinkerlaunch.log` file, so that it can be opened graphically.
    - Path to `/dev/shm/steamtinkerlaunch` folder, which may contain other logs specific to that session.
    - Path to the SteamTinkerLaunch log directory, which is a configurable path, but defaults to `${STLCFGDIR}/logs/steamtinkerlaunch`.

As per usual, if `xdg-open` is not available on the system, these files cannot be opened. If the paths don't exist in the first place when the menu is being opened, they won't be displayed on the menu (following the existing pattern).

TODO:
- [x] Update remaining langfiles

### DIFF
--- a/lang/chinese.txt
+++ b/lang/chinese.txt
@@ -1284,3 +1284,6 @@ GUI_IGNORECOMPATSLR="Ignore Native Linux Steam Linux Runtime from Compatibility 
 DESC_IGNORECOMPATSLR="ignore Steam Linux Runtime for native games selected either by you or by Valve Testing as a Compatibility Tool and let SteamTinkerLaunch find it instead. For example, ignore Steam Linux Runtime 3.0 selected by Valve Testing and let SteamTinkerLaunch find and use Steam Linux Runtime 1.0 instead. This can have MAJOR impacts on compatibility, and has NO EFFECT for games using Proton"
 GUI_VORTEXDEVICESCALEFACTOR="Vortex GUI Scale Factor"
 DESC_VORTEXDEVICESCALEFACTOR="uses the --device-scale-factor flag to set the Vortex GUI scale. Defaults to 1.0, but if you want a larger GUI you can set it to match your display's scale factor (or the largest scale factor, if using multiple across displays)"
+GUI_STLSHMLOG="Current SteamTinkerLaunch log file"
+GUI_STLSHMDIR="Current session logging directory"
+GUI_STLPERGAMELOGSDIR="Per-game log files"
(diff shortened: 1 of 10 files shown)


sonic2kk:
@OrderedSet86 This is based off of our discussion in https://github.com/sonic2kk/steamtinkerlaunch/issues/1118#issuecomment-2161904545. Feel free to take a look, if you like.

The logging changes in Game Files were primarily made around your point here:

> Yes, I appreciate this! This would be a large change but you might consider the Lutris style of logging, where a window pops during game installation / setup for logs. Since STL is a tinkering tool it could just always have a log window open. Additionally, if logs are sometimes hidden, a "logs" button would be helpful - Lutris lets you right click on the game and click "Show Logs", but for STL you could include it in the settings / mainmenu grid. It doesn't have to be fancy, just xdg-open the user's editor pointing at `/dev/shm/steamtinkerlaunch/vwrun.txt`. Somewhere in here perhaps: [image of Main Menu]

### Logging Window
Because all SteamTinkerLaunch windows are Yad dialogs, persistent windows like you have described are not really feasible. When you click any of the buttons along the bottom row of a window, the reason it closes is because those are dialog action buttons, a row of glorified "Ok/Cancel" buttons. The Main Menu buttons spawn separate processes that spawn further Yad dialogs. Dynamic windows like the one in Lutris is also not possible, I don't think you can continually update dialog content with Yad.

In theory, there could be a button that would open a dialog with log text, but it wouldn't be possible to my knowledge to dynamically update this content with Yad, so it wouldn't do anything. On top of this, it would be tied to the Steam Game process. If it was started as, say, a separate disowned SteamTinkerLaunch process, then `steamtinkerlaunch.log` would contain logging for _that_ process (we may be able to hack in a way to force-disable logging for some processes, but it would be ugly. But to begin with there would be no way to my knowledge to achieve that kind of dynamic window behaviour with Yad.

This is why I went with the  `xdg-open` approach. Game Files already has this integration, so including the files here in this way made the most sense to me. The files here are primarily for per-game configuration as noted on the wiki, but I think having logs here as well is fine too.

### Include `vwrun.txt` Link
The `vwrun.txt` log is too specific in my opinion to include, so as a compromise, I went with including a way to open `/dev/shm/steamtinkerlaunch`, where the log is stored. Plus, the Vortex log may not exist when this dialog is opened, which is why it is not included.

In addition to this, I also included a new log line that states where the Vortex run log (`vwrun.txt`) is stored, so that users can more easily find it.

<hr>

I aim to strike a balance here of incorporating your suggestion, which I understand to essentially be to make the logs easier to access within the limitations here.

You don't have to test out the PR or anything, I'm pinging for awareness. I try to include users who make suggestions where possible in discussion. Normally, this would've been discussed in a dedicated issue, but to avoid further cluttering the earlier issue, I thought pinging here about this specific change made more sense.

sonic2kk:
I think this is good to merge, please let me know if you have any comments based on what I couldn't add exactly in your original ask. Anything you want to reply to in my above message is still welcome :smile:

OrderedSet86:
Thanks, I really appreciate these changes! I also will be putting in a PR to update the README shortly. Hopefully that will lead to less confusion over using the correct version.

OrderedSet86:
https://github.com/sonic2kk/steamtinkerlaunch/pull/1127

PR #1128, opened 2024-06-17
Custom Command: Add logging for native custom commands

sonic2kk:
Adds logging for native custom commands, where the output (including any errors) will be stored under `STLSHMDIR` as `launchCustomProg.log`. Proton titles already had this loggiing in `extProtonRun`.

I would also like to find a way to log the `PROTON_LOG` for Proton custom commands. I'm not sure if it's possible, I think it is from looking at the Proton script - it should default to `steam-proton.log` if there is no `SteamGameId` in the environment, and I'm not sure if it will pick up the game AppID by default. If it does, we can temporarily unset this var before launching, and re-set it after launching as a hacky workaround if we really must.

We could even temporarily manipulate the value of `SteamGameId` to allow us to set the log file to something like `steam-<gameid>-customcmd.log` so that it is obvious the log is for a custom command pertaining to the given AppID.

Allowing a Proton log has the added benefit of being able to see some Proton-specific environment variables being set. The Proton script has a section where it tries to, for example, log information about the current Steam Linux Runtime. We _could_ re-implement this ourselves, but I'd rather hook into what Proton is currently doing.

All this also hinges on whether or not custom commands running with Proton can even use `PROTON_LOG`. When using `ONLY_CUSTOMCMD` in particular I'm not sure how many of our set environment variables (Proton-related or otherwise) get passed to custom commands, for example the ones in the per-game environment variable file, and the global one.

<hr>

This PR does accomplish one logging improvement, but further improvement to be able to give a Proton log since we run custom commands with Proton would be pretty nice.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240618-1 (improve-customcmd-logging)"
+PROGVERS="v14.0.20240619-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
It seems we _can_ use `PROTON_LOG=1`, but it won't actually output anything unless we have `SteamGameId`. This is a problem for two reasons:
- When running from Steam with a game (if we load STL with a game by pressing "Play" and then try to use One-Time Run), we will have `SteamGameId` but it'll be the same as the game's AppID. So it wouldn't be possible to use both `PROTON_LOG` for the game and custom command.
- When running outside of Steam (such as by using `steamtinkerlaunch otr`) then we won't have `SteamGameId` so the Proton log won't actually get generated.

The Proton Python script [does actually have logic](https://github.com/ValveSoftware/Proton/blob/bb76da532ee936dcb33e6c3b6e944c7dcfbd2f64/proton#L1294-L1295) that tries to account for generating a Proton log without a `SteamGameId`, but it only runs if `append_forever` is `True`. This seems to be controlled elsewhere, and is not set up by default when running the logic to generate a log.

To get around this, we can probably temporarily set the `SteamGameId` when running a custom command. The value of the environment variable shouldn't be permanently changed if we set it inline when running the command, but we will need to test this.

So we can actually `PROTON_LOG` for custom commands. And we could control what the logfile name will be by controlling how we set `SteamGameId`. But the last problem is cleanly implementing this functionality optionally...

We will probably have to pass an argument to `extProtonRun` to control if we want Proton logging, off by default. We will then have to add it to each custom command launch case that runs Proton. Finally, we will need to set `SteamGameId` for the Proton logging.

Adding the ability to `PROTON_LOG` is probably overkill, I think we are fine for now, but I will add a TODO in case this comes up later or if someone wants to implement it. It would be a nice feature but not one I think is worth delving into right now, as we'd need a lot of custom logic to handle optionally setting `SteamGameId` and setting it to a custom value (or default if no value is given, such as `extProtonRun`) *if* the Proton log is enabled.

I may revisit it in future or someone else might, so a TODO it is. After that, this PR is ready to merge.

PR #1129, opened 2024-06-17
One-Time Run: Default to per-game Proton version if `--proton` is not passed

sonic2kk:
This PR allows for running One-Time Run without passing a Proton version. If the Proton version given is blank, or if `--proton=` is omitted entirely, we will fall back to the per-game Proton version, "`USEPROTON`".

If the per-game Proton version `USEPROTON` does not exist in `ProtonCSV.txt` (perhaps there have been Proton updates, or the game has not been launched in a long time and the value is a very old GE-Proton version) then we will resolve the mismatch using the `USEPROTON`.

This PR also fixes an issue with commandline One-Time Run where blank values were not being resolved correctly, because `fixProtonVersionMismatch` does not run unless `ISGAME -eq 2`, which is not guaranteed to be the case with One-Time Run if launched outside of Steam. Therefore, if we get far enough into the `commandlineOneTimeRun` function that we think we need to use Proton, we will force `fixProtonVersionMismatch` by passing a new argument `FORCEPROTONMISMATCHRESOLVE` which defaults to `0` if not passed to preserve existing functionality.

With that `fixProtonVersionMismatch` alone, it meant that it would've returned the first item in ProtonCSV if the mismatch version to resolve was blank (i.e. if `--proton` was not passed), which is not ideal. Instead this PR has extra functionality to make `fixProtonVersionMismatch` fall back to the per-game Proton version if `RUNOTPROTON` is blank (i.e. if `--proton` is not passed).

As a result, this PR means the following lines work:
- `steamtinkerlaunch otr --exe="/path/to/valid/exe"` - This previously would not have worked because `fixProtonVersionMismatch` would not have ran, so the EXE would not run
    - With this PR, not only is this fixed, but we will fall back to the per-game Proton version instead of relying on `fixProtonVersionMismatch` to return whatever the first Proton value is in ProtonCSV
- `steamtinkerlaunch otr --exe="/path/to/valid/exe" --proton="invalid-proton"` - This was previously _meant_ to work by resolving with `fixProtonVersionMismatch`, but this would have usually failed because the function was not running at all if `ISGAME` was not `3`. Now it will attempt to resolve the mismatch.

<hr>

There are some things left to do in this PR, mainly I want to check to see if `OTPROTON` needs to be updated after a Proton version mismatch. Logically it seems to me that it should, since it is only set to the value given by `--proton=""` in the form of `RUNOTPROTON`, but as we can update this value to resolve mismatches if it is invalid, that means `OTPROTON` could be left being an invalid Proton version; that is, whatever `RUNOTPROTON` was originally set to.

TODO:
- [x] Further testing
- [x] Check if `OTPROTON` and other saving behaviour needs updated after `RUNOTPROTON` gets updated for a Proton version mismatch resolution
- [x] Version bump


sonic2kk:
`OTPROTON` does need updated. It is blank if `--proton` is not given, and is incorrect if there is a mismatch resolution. It is set to the Proton name and not the path unlike `RUNOTPROTON`, so we need a way to get the `RUNOTPROTON` name and assign it to `OTPROTON` after we have confirmed a valid Proton version.

We can probably get the RUNOTPROTON name from the path using the ProtonCSV.

sonic2kk:
Fixed `OTPROTON` issue. Now I need to verify that the UI still works fine with the selected Proton version.

sonic2kk:
One-Time Run GUI still works as expected. Mismatch still works as expected on the UI as well. Tested using Proton Experimental on the UI which works.

Also tested the mismatch resolution on the UI by manually setting an invalid Proton version on the config which displayed on the UI and then resolved correctly when launching the game. The GUI defaults to the Proton version set in the Per-Game Config (`USEPROTON`).

<hr>

Remaining work for this PR is to update the help screen to note that if `--proton` is left out, it will use the Per-Game Proton version by default. We also need to document this [on the One-Time Run wiki](https://github.com/sonic2kk/steamtinkerlaunch/wiki/One-Time-Run). After that, a rebase, version bump, and merge.

sonic2kk:
Wiki page has been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/One-Time-Run/_compare/364375a3fb9f2280228d1a292056b3eb5545400d

sonic2kk:
This change has not been noted on the changelog because it pertains to commandline One-Time Run usage, which is is not available in a release.

PR #1131, opened 2024-06-23
ModOrganizer 2: Add option to change Silent Mode Executable

### LINKED ISSUE #1130: MO2 Silent mode only loads game executable
## System Information
- SteamTinkerLaunch version: Latest (auto updates every launch)
- Distribution: Steam Deck
- Installation Method: CLI

## Issue Description
When I launch Skyrim Special Edition with MO2 (via steam tinker launch) with GUI selected, then tap "play" it works flawlessly - loads SKSE64 as it should, all good! However, when I set the MO2 launch to "Silent" it doesn't appear to load SKSE64 and instead launches the game directly or possibly uses some unknown profile? Regardless - no mods are loaded *at all*.

Anyone have any ideas?

sonic2kk:
Fixes #1130.

## Overiview
Adds the option to choose which ModOrganizer 2 Executable Configuration gets launched in Silent Mode.

image

*(Note: The default value is `none`, but for illustration purposes, I chose this an alternative value from the dropdown.)*

## Background
To launch in Silent Mode, we pass the MO2EXE `moshortcut://:exe_name` (note that this has *two* colons). By default, this is the Executable configuration matching the INI/default internal MO2 Instance name (i.e. `Oblivion`, `Skyrim Special Edition`, `Fallout New Vegas`). So by default for SkyrimSE, we would launch `molaunch://:Skyrim Special Edition`. This name should correspond to an executable configuration defined in MO2 which points to a real executable, working directory, etc. MO2 creates these for Bethesda games in order to skip the launcher usually, so this would launch the SkyrimSE EXE without using the launcher.

However some games, such as SkyrimSE, need to use SkyrimSE Script Extender (SKSE64) in order to launch the game with mods. This launch configuration will be defined inside of MO2 in order to launch the game with mods in GUI mode, but to use it in Silent Mode, we need a way to give `moshortcut://:` a different executable configuration name. In this case, instead of using `moshortcut://:Skyrim Special Edition`, we need a way to give it `moshortcut://:SKSE64`.

## Implementation
To achieve this, I added a combobox entry dropdown on the Game Menu to optionally override the Executable Configuration name given to the `moshortcut://:` command. This is under the ModOrganizer 2 settings so will be hidden if ModOrganizer 2 is not used. This dropdown controls the value of `MO2SILENTMODEEXEOVERRIDE` (which is `none` by default).

By default, dropdown is this is `none`, meaning we will do nothing by default, we won't override anything, and any users using Silent Mode today will not be impacted. If the dropdown is blank we also won't do anything.

However a user can click on this dropdown and choose a list of executable configurations parsed out of the ModOrganizer 2 instance INI at `/path/to/mo2compdata/pfx/drive_c/users/steamuser/AppData/Local/ModOrganizer/<GAMENAME>/ModOrganizer.ini`. The list of available executable configurations is stored in the INI like this:

<details>
<summary>Oblivion INI example</summary>

```ini
[customExecutables]
size=3
1\arguments=
1\binary=Z:/path/to/steamapps/common/Oblivion/OblivionLauncher.exe
1\hide=false
1\ownicon=true
1\steamAppID=
1\title=Oblivion Launcher
1\toolbar=false
1\workingDirectory=Z:/path/to/steamapps/steamapps/common/Oblivion
2\arguments=
2\binary=Z:/path/to/steamapps/common/Oblivion/Oblivion.exe
2\hide=false
2\ownicon=true
2\steamAppID=
2\title=Oblivion
2\toolbar=false
2\workingDirectory=Z:/path/to/steamapps/common/Oblivion
3\arguments=\"Z:\\path\\to\\steamapps\\common\\Oblivion\\data\"
3\binary=C:/Modding/MO2/explorer++/Explorer++.exe
3\hide=false
3\ownicon=true
3\steamAppID=
3\title=Explore Virtual Folder
3\toolbar=false
3\workingDirectory=C:/Modding/MO2/explorer++
```

</details>

The part we have to give to `moshortcut://:` is the `title` property. So we can parse out the executable titles using `sed`, getting every line that starts with a number and a backlash containing `title=`. We also have to remove the carriage return, I tried to do this with `sed` but ended up just doing it with parameter expansion.

To populate the dropdown with these values we create a new function called `createMO2SilentModeExeProfilesList` that runs on the Main Menu, similar to how we run a function on the Main Menu to populate the ReShade versions list. This function will exit early if the current game is not in `mo2games.txt` (and similarly if `mo2games.txt` doesn't  exist). Then, if the INI file we need doesn't exist, we also return early. This should hopefully minimize loading times, as the Main Menu can take quite a while to load and this would be an IO bound task.

Assuming we are able to find the INI, we loop through the matches returned by `sed -n 's/^[0-9]\\title=//p'` (matches everything after patterns line `\1title=` on each line). The order of these matches is determined by their order in the ModOrganizer 2 executable configurations list. You can re-order entries, so the order we use here follows that as the INI also stores these sequentially.

We store these entries in an exclamation-point-delimited string, because that's what `cleanDropDown` expects. If an executable name contained an exclamation point, this could be problematic, so we should document this potential problem. Likewise we should document that although there is a preset list, the user can enter their own custom value, but if they enter a value that doesn't exist then ModOrganizer 2 may not work.

The final string, which could look something like `none!Oblivion Launcher!Oblivion!Explore Virtual Folder`, is then used on the Game Menu to display the items in the dropdown. The value for `MO2SILENTMODEEXEOVERRIDE` can also be manually set in the Per-Game Config File, as usual.

Once selected, when we try to launch ModOrganizer 2 in Silent Mode, if we have a defined `MO2SILENTMODEEXEOVERRIDE` that is not `none`. then we use that as the value to pass to `moshortcut://:` instead of hardcoding to the INI name. If `MO2SILENTMODEEXEOVERRIDE` is blank or `none`, then we fall back to the INI name that we're currently using.

<hr>

I tested this one and it seems to work as expected, resolving the linked issue.

TODO:
- [x] Test non-root installs to make sure `GLOBALMISCDIR` works as expected
- [x] Code cleanup
- [x] Update langfiles
- [ ] Update wiki
    - [ ] Note how Silent Mode works by default (uses the default executable profile based on INI / instance internal name)
    - [ ] Note that it can be changed and why (to use other executables to load mods, such as `SKSE64`)
    - [ ] Note that executable names with exclamation points may have issues and that it is best to avoid using this character in executable names
    - [ ] Note that executable names can also be entered, but if they don't exist, ModOrganizer 2 may crash
- [ ] Version bump


sonic2kk:
If you set a executable name that doesn't exist, ModOrganizer 2 simply tells you that the executable doesn't exist for this instance.

sonic2kk:
Cleaned up code and bumped langfiles. Verified that `GLOBALMISCDIR` should be defined for local installs.

Have to get some wiki updates ready but after that this should be ready to go pending a version bump!

sonic2kk:
Did a quick sanity check to make sure this does not impact non-MO2 games and non-MO2 native games. `createMO2SilentModeExeProfilesList` correctly exits early.

sonic2kk:
I have the wiki update ready to go. This PR can be merged.

sonic2kk:
Wiki has been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Mod-Organizer-2/_compare/0f5e1172f8d306fa13ec1e9b28ca472c790662b7

sonic2kk:
This PR was not noted on the changelog because Silent Mode is not available in a release.

PR #1136, opened 2024-07-04
Steam Deck: Bump innoextract version to 1.9-9

### LINKED ISSUE #1135: SteamOS 3.6 Requires `innoextract` 1.9-9
## System Information
- SteamTinkerLaunch version: 'v14.0.20240624-1'
- Distribution:  Steam OS
- Installation Method: Manual / git clone

## Issue Description
Today I updated my steam deck and innoextract issues have started up again. I have read through all the other posts regarding innoextract and I understand what it is used for and that the dependencies change as the OS/STL Update.  Last time this happened, simply removing /home/deck/stl/deps/usr/bin/innoextract and running STL again resolved.  This time that doesn't seem to work.  If I run STL mo2 start from desktop mode , it opens and I can launch my game no problem.  But running from steam/game mode causes it to try and reinstall innoextract and it fails due to missing file. 

What does seam to work for running from steam, is making sure that /home/deck/stl/deps/innoextract-1.9-8-x86_64.pkg.tar.zs doesn't get removed when running STL.  I made the file immutable and it resolved my issue. 

## Logs

[before setting +i innoextract.txt](https://github.com/user-attachments/files/16076858/before.setting.%2Bi.innoextract.txt)

[after setting +i innoextract.txt](https://github.com/user-attachments/files/16076860/after.setting.%2Bi.innoextract.txt)

sonic2kk:
Fixes #1135.

SteamOS 3.6 updates the Arch snapshot so we need to do this again.

Opening as draft because SteamOS 3.6 is not yet in Stable.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240705-1 (steamos-3.6)"
+PROGVERS="v14.0.20241025-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
SteamOS 3.6 is out now, this is probably needed. See: #1190

sonic2kk:
Rebased and bumped version, should be good to merge still.

PR #1159, opened 2024-08-31
Editor: Move `GLOBCUSTVARS` in to `EDITORCFGLIST` section to show in EDITOR menu

HanPrower:
A simple change which allows `global-custom-vars.conf` to show up in the EDITOR dialog.


sonic2kk:
Huh, I never used that dog, but change makes sense to me.

Tested and it appears as expected with this PR.

image

I guess the per-game environment variables path already shows up here (I believe it's `gamecfgs/customvars/<appid>.conf`), so the global variables were likely just missed.

Not that it invalidates these changes as all, but the global custom variables file is also available under the "Game Files" menu as well.

<hr>

Looks good, I think this can be merged as-is. Thanks! :-)

Issue #1158, opened 2024-08-30
Setting `LD_PRELOAD` in a customvars conf file does not work as expected


HanPrower:
## System Information
- SteamTinkerLaunch version: v14.0.20240829-2
- Distribution: Arch Linux
- Installation Method: Flatpak

## Issue Description
With my recent forays in to messing with gamescope I have bumped in to a known issue where after ~24 minutes the game will begin to lag (gamescope seems to rapidly flip between 2D and 3D power levels) (and yes! having to wait 24 minutes each time to debug it _was_ fun 😭).

There are [2 documented workarounds](https://wiki.archlinux.org/title/Gamescope#Launching_gamescope_from_steam,_stuttering_after_~24_minutes):
1. Using `-e` (which only works in certain scenarios - chiefly, seemingly, when launching gamescope externally from Steam)
2. Passing a blank `LD_PRELOAD=` in to gamescope (i.e. `LD_PRELOAD="" gamescope -f -w 2560 -h 1440 --mangoapp -- %command%`

For me only the second way works, however, not when using STL.

The assumption I have arrived at rests with how `loadCfg` works.

The function `export`s into the environment, which makes sense logically. Prepare the environment, run the things in the environment.

However, this doesn't seem to pass the environment variable correctly to whatever is being run when compared to how Steam seems to do it when you put `envvars` in the "Launch Options" box.

I don't think `export` is wrong in general. It just doesn't seem to reflect how Steam works?

I'm aware overriding `LD_PRELOAD` is an extreme case (because doing so breaks the Steam Overlay) and usually any other set `envvars` aren't going to be re-manipulated by Steam, but it's worth a discussion if STL is functioning as expected.

My initial thought is that an option is added where you can either choose to `export` or to prefix the variables to the launch command.

## Logs
None are really relevant.

sonic2kk:
If I understand, I think SteamTinkerLaunch is functioning correctly, but that for some specific cases where Steam will override the variables (like it does for `LD_PRELOAD`) doesn't work, is that right?

For example if you were to manually export the variable to enable MangoHUD, would that work? Or some other variables, like Zink or something? I can't think of any more explicit examples off the top of my head 😅

If that above understanding is wrong, then you're absolutely right and we should figure out what is going on!

> I don't think `export` is wrong in general. It just doesn't seem to reflect how Steam works?

[...]

PR #1161, opened 2024-09-01
Add some checks to `restoreOrgVars` on `startGame`

HanPrower:
My tentative and ugly attempt at fixing #1158 

Simply when we're in `startGame` check the relevant custom-vars to see if any of the envvars that would normally be reset to original values are defined, and just don't reset them.

Probably needs more testing to be safe, but so far so good.

I wasted so much time on this :crying_cat_face:

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7208,7 +7208,7 @@ function extProtonRun {
 		# Hopefully unsetting is safe and doesn't mean places that need the SLR will lose it from this 'unset'
 		if [ "$EXTPROTUSESLR" -eq 1 ]; then
 			writelog "INFO" "${FUNCNAME[0]} - EXTPROTUSESLR is '$EXTPROTUSESLR' -- Attempting to find and use SLR with extProtonRun"
-			
+
 			unset "${SLRCMD[@]}"
 			setSLRReap
 		fi
@@ -7782,7 +7782,7 @@ function getGameWindowPID {
 
 		return
 	fi
-	
+
 	MAXWAIT=20
 	COUNTER=0
 	TESTPID="$NON"
@@ -12971,7 +12971,7 @@ function launchCustomProg {
 					writelog "INFO" "${FUNCNAME[0]} - FORK_CUSTOMCMD is set to 1 - forking the custom program in background and continue"
 					extProtonRun "FC" "$LACO" "$CUSTOMCMD_ARGS" "" "${CUSTOMCMD_USESLR}"
 				elif [ "$ONLY_CUSTOMCMD" -eq 1 ] && [ -n "${FINALOUTCMD[*]}" ]; then
-					writelog "INFO" "${FUNCNAME[0]} - ONLY_CUSTOMCMD is set to 1 and we have some arguments in FINALOUTCMD - passing to extProtonRun to build a valid start command"	
+					writelog "INFO" "${FUNCNAME[0]} - ONLY_CUSTOMCMD is set to 1 and we have some arguments in FINALOUTCMD - passing to extProtonRun to build a valid start command"
 					extProtonRun "R" "$LACO" "$CUSTOMCMD_ARGS" "$FINALOUTCMD" "${CUSTOMCMD_USESLR}"  # extProtonRun will handle adding the FINALOUTCMD args to
 				else
 					extProtonRun "RC" "$LACO" "$CUSTOMCMD_ARGS" "" "${CUSTOMCMD_USESLR}"
@@ -13017,7 +13017,7 @@ function launchCustomProg {
 
 				writelog "INFO" "${FUNCNAME[0]} - Steam Linux Runtime enabled, attempting to fetch Steam Linux Runtime for native Custom Command"
 				# "2" is the FORCESLRTYPE, meaning we want to force to get the native SLR -- We do this in case we are trying to launch a native custom command with a Proton title
-				# In this case, setSLRReap is going to have the Proton SLR vars set from the game launch, so we need to force it here to use the native SLR 
+				# In this case, setSLRReap is going to have the Proton SLR vars set from the game launch, so we need to force it here to use the native SLR
 				setNonGameSLRReap "2"
 			fi
 
@@ -17859,7 +17859,7 @@ function prepareMO2 {
 }
 
 function createMO2SilentModeExeProfilesList {
-	# Get all of the ModOrganizer 2 executables launch configurations in the instance's INI 
+	# Get all of the ModOrganizer 2 executables launch configurations in the instance's INI
 	# The user can use this to override which 'moshortcut://' is launched in Silent Mode
 	MO2SILENTMODEEXEPROFILES="$NON"
 	MO2GAMES="$GLOBALMISCDIR/mo2games.txt"
@@ -20308,7 +20308,7 @@ function startGame {
 		writelog "INFO" "${FUNCNAME[0]} - ## STL LAUNCH COMMAND: '${RUNCMD[*]}'"
 		writelog "INFO" "${FUNCNAME[0]} - ## GAMESTART HERE ###"
 
-		restoreOrgVars # restore original LC_ and friends for the game as long as not set in custom-vars
+		restoreOrgVars
 
 		GRUNLOG="$STLGLLOGDIRID/${AID}.log"
 		if [ "$ISORIGIN" -eq 1 ]; then
@@ -25433,7 +25433,7 @@ function restoreOrgVars {
 
 		# Check if the envvar has been defined in a custom-vars conf file, if so don't restore it and continue
 		if grep -q "^${RESTOREVAR}" "$GLOBCUSTVARS" || grep -q "^${RESTOREVAR}" "$GAMECUSTVARS"; then
-			writelog "WARN" "${FUNCNAME[0]} - Not restoring '${RESTOREVAR}' to original value as it is defined in a custom-vars conf file"
+			writelog "WARN" "${FUNCNAME[0]} - Not restoring '${RESTOREVAR}' to original value as it is defined in a custom-vars conf file. This may potentially cause issues."
 
 			continue
 		fi
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -20308,7 +20308,7 @@ function startGame {
 		writelog "INFO" "${FUNCNAME[0]} - ## STL LAUNCH COMMAND: '${RUNCMD[*]}'"
 		writelog "INFO" "${FUNCNAME[0]} - ## GAMESTART HERE ###"
 
-		restoreOrgVars # restore original LC_ and friends for the game
+		restoreOrgVars "startGame" # restore original LC_ and friends for the game as long as not set in custom-vars
 
 		GRUNLOG="$STLGLLOGDIRID/${AID}.log"
 		if [ "$ISORIGIN" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


sonic2kk:
Sorry it took me so long to get around to taking a look at this!

I think I see the solution here, essentially check for some known variables that Steam will override, and if they're defined in either the Global or Per-Game Custom Environment Variables config file, don't overwrite them.

The issue in #1158 was then, based on the comment in that issue, not actually caused by Steam but by some funky SteamTinkerLaunch behaviour, which this prevents.

I think calling your solution "ugly" is a stretch :sweat_smile: It makes sense, although I think we can improve it a little bit.

We could make an array with these variables and loop through them, something like `STLNOOVERRIDEENVVARS`, or something with less repeating characters ideally. Something like this `ABETTERVARIABLENAME=( "LD_PRELOAD" "LD_LIBRARY_PATH" "LC_ALL" "PATH" )`.

Then we can loop through with something like this, where we can dynamically build the variable name to assign to using `eval`, which will allow us to extract the variable _name_ rather than the _value_ and runs it as a command. We can create a string representing the variable name we want to assign to based on the current iteration in the loop, so if `LOOPVAR` was `LD_PRELOAD`, we could write `ORG_${LOOPVAR}` which would become `ORG_LD_PRELOAD`. Then we can use `eval` to assign `LOOPVAR`, which is `LD_PRELOAD`, the value of `ORG_${LOOPVAR}`, which would be the `ORG_LD_PRELOAD` variable, assuming this exists and is defined (which is an assumption we can make).

Basically using `eval` lets us build a string command to run for some dynamic assignment. I wrote up a code snippet, I didn't test it, I just wrote it up in a text file and tested the concept in a Bash prompt. Hopefully it gives the general idea though :-)

```bash
if [ "$1" != "startGame" ]; then
    # Return early if not startGame, I think this is safe to do as nothing else in the function should run if $1 != startGame
    # This should also simplify the condition in the loop
    return
fi

# STLRESTOREVARS could be a potential name, but open to better suggestions
STLRESTOREVARS=( "LD_PRELOAD" "LD_LIBRARY_PATH" "LC_ALL" "PATH" )
for RESTOREVAR in "${STLRESTOREVARS[@]}"; do

    # e.g. 'ORG_LD_PRELOAD' for 'LD_PRELOAD' on first iteration of the loop
    BUILTORGVARNAME="\$ORG_${RESTOREVAR}"

    # I think we can use grep like this? I think sed is fussy about using variables, not sure about grep!
    if grep -q "^${RESTOREVAR}" "$GLOBCUSTVARS" || grep -q "^${RESTOREVAR}" "$GAMECUSTVARS"; then
	    # Since `eval` extracts the value and runs the command, the part inside the quotes comes out to e.g. 'LC_ALL="$ORG_LC_ALL"'
	    # The dollarsign comes from the '\$' earlier, it's also possible to add it here but I would prefer to build it all in BUILTORGVARNAME
	    #
	    # Since this runs as a command, this is equivalent to typing in a Bash prompt 'LC_ALL=$ORG_LC_ALL'
	    # Everything in the quotes is a string that gets built as a command to run, and 'eval' then runs it
	    #
	    # It's all a bit weird, I hope that makes sense. I also only tested this with some test code in a Bash prompt, so it may need some more refining
	    eval "${RESTOREVAR}=\"${BUILTORGVARNAME}\""
	fi
done
```

Something like the above should be a bit more concise and allow us to more easily extend this if there are other variables to include, or if we ever want to make this list user-facing.

As well as this, in the loop we should make the log a big more explicit that these are not being overridden and that, in some cases, this could be problematic. Any user overriding these should ideally know what they are doing and should be doing it for good reason (locale could be overwritten for testing purposes, for example).

<hr>

Apart from the above potential improvement, I think this approach is fine, provided it doesn't cause any regressions for cases where the user doesn't override these. In other words, things should continue as normal if this is not defined in any of the custom environment variable files, but if they are defined, your issue in #1158 should be resolved :-)

If you can give some explicit testing steps I would also be happy to test this out, just to make sure we're on the same page.

Thanks a lot for your work and patience here to figure this out!

HanPrower:
Don't apologise for taking some time to review things. It's a hobby project, you likely have better things to do and stressing over replying here will probably lead to burn out :)

My fix _was_ ugly it was just a throw-my-hands-up-and-get-something-that-works-and-go-cry fix. What's funny is I asked my g/f how she'd fix it (she's a better coder than me) and she said she'd build an array and check against it :sweat_smile: Probably not quite the same as what you came up with, but still.

---

Just to address something: I specifically put in a check for `startGame` just to make sure it worked for me specifically without causing issues whilst launching a game (because of the gamescope issue).

However, `restoreOrgVars` is actually called 10 other times:
- 8 of them are in `extProtonRun` just before a custom program is run.
- - I'm not sure we want to restore all the envvars here as a user may expect custom vars to be loaded for whatever custom programs they're running.
- 1 is in `checkFWS` (for running FlawlessWideScreen)
- - Again, probably don't want to restore them here as is this launching proton and fws, so the user again will likely expect what they set to survive here
- 1 is in `installWinetricksPaks`
- - We may want to restore them to original here... I'm not sure what winetricks cares about as environment, but as it's trying to do something specific within a wine prefix it may want to be "default". Then again... if there's an edge issue that's fixed by overriding the envvar... :man_shrugging: 

So instead of `$1` being "startGame" it may want to be "checkCustomVars" or, inversely, "forceOriginal".
Then if `checkCustomVars` is set that's the only time we check the greps and if the greps are true we just `continue` otherwise the original is restored.

Alternatively we don't have an `$1` and always set things as the user defines. Then have a message in the log about the user has overridden things and it may cause issues.

I may lean towards no `$1`, as always stated STL is for advanced users, and if people are overriding those sorts of envvars they must have a good reason. It's not like things will explode... at worst locales will be odd, or the Steam overlay will break, or programs simply won't be found and just deleting or commenting the line or closing STL and relaunching will reset the envvars to Steam's defaults anyway.


---

As for a way to test you can simply try `LD_PRELOAD=""` in a custom-vars conf file. That will cause the Steam Overlay to not function even if enabled in Steam :smile:
You'll also notice the lack of this spam in your system's journal when you launch a game / program from STL:
`Sep 03 10:33:17 smokey steam[20078]: ERROR: ld.so: object '/home/hanfox/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.`

sonic2kk:
> My fix was ugly it was just a throw-my-hands-up-and-get-something-that-works-and-go-cry fix. What's funny is I asked my g/f how she'd fix it (she's a better coder than me) and she said she'd build an array and check against it 😅 Probably not quite the same as what you came up with, but still.

Heh, I like that expression. :smile: 

And well, it is still an array, we just dynamically assign them. In other languages you could probably use a hash/dictionary or something similar, but in Bash we can make do with this :-)

<hr>

You raise some good points here about when we should and should not use this logic. I think in most cases a user is going to expect what they define in their custom environment variables file to carry over, the benefit of this PR is to allow those variables to survive and to not have SteamTinkerLaunch override them.

> I'm not sure we want to restore all the envvars here as a user may expect custom vars to be loaded for whatever custom programs they're running.

First, I guess the 8 calls are for different code paths? Or is this called 8 times consectutively when starting a custom program? If so, it's not something for this PR but yikes... I'll need to investigate that!

Apart from that, I would agree that custom environment variables should be passed to custom commands and that this logic should apply for them as well. 

If a user runs into problems here, there is a warning log that the values aren't being restored because they're already defined in the custom environments variable file.

> Again, probably don't want to restore them here as is this launching proton and fws, so the user again will likely expect what they set to survive here

I agree with you here as well.

> We may want to restore them to original here... I'm not sure what winetricks cares about as environment, but as it's trying to do something specific within a wine prefix it may want to be "default". Then again... if there's an edge issue that's fixed by overriding the envvar... 🤷‍♂️

This is a tricky one for sure. And there have historically been cases where Winetricks has benefitted from overriding, for example, locale. On the other hand this could cause unexpected behaviour with Winetricks that is actually needed for a game.

I think for now we should keep still allow the current behaviour of this PR to apply to Winetricks, where variables won't get restored if they're defined in the custom environment variables file. This means a user can apply those overrides if they need to.

> I may lean towards no $1, as always stated STL is for advanced users, and if people are overriding those sorts of envvars they must have a good reason. It's not like things will explode... at worst locales will be odd, or the Steam overlay will break, or programs simply won't be found and just deleting or commenting the line or closing STL and relaunching will reset the envvars to Steam's defaults anyway.

I am leaning towards this as well, based on the above. I think updating the wiki to document this will help too. Honestly, I don't think many people read the wiki (a couple people complained before when I mentioned the wiki) but still,  I think making a note of this is useful. I also refer to the wiki as well if I'm ever looking at code and wondering the use-case.

I think having no conditional check for what is calling `restoreOrgVars` is the best approach. Anywhere it is called we should just follow the logic to restore the original variables unless they are defined in the Global or Per-Game config file. We can documented this and call out the cases you have mentioned, in case users actually _want_ to take advantage of this for, say, Winetricks and override some environment stuff.

I am with you on users that do this must have a good reason. This is something users probably aren't going to do lightly, and I think it is valuable to override these. The more I consider it the more I agree that SteamTinkerLaunch not doing this is already a _bug_, as _I_ would have also expected the override to work. Although SteamTinkerLauch probably didn't do it before because we want to restore those original variables since we might modify them (in particular, PATH on SteamOS we probably modify (not on disk, just the environment variable)).

> As for a way to test you can simply try L`D_PRELOAD=""` in a custom-vars conf file. That will cause the Steam Overlay to not function even if enabled in Steam 😄
> You'll also notice the lack of this spam in your system's journal when you launch a game / program from STL:

I will give it a try ASAP, does it work on your end out of interest?

<hr>

The code looks good to me, if testing goes well I don't have any comments on the code itself. I'm glad you were able to implement it from my horrid explanation heh.

ShellCheck is grumpy because now it thinks the `ORG_` variables are unused in `saveOrgVars`. The "real" fix is to apply similar logic to what you implemented in this PR to that function also, although because that introduces duplicate logic it would probably be best to make a function to do it, and I think that's way out of scope for this PR. For now, we can ignore the ShellCheck warning with a directive comment `# shellcheck disable=2034`. This tells ShellCheck to be quiet because we know better than it does :wink: you might have to put it above each line, it might get grumpy if you try to put it beside each line, but you can try putting it beside first and seeing if that works.

If you could also add a comment in `saveOrgVars`: `# TODO resolve the unused warnings by dynamically assigning the variables like we do in 'restoreOrgVars'`.

<hr>

Thanks again for your work on this! Provided it works in my own testing (which from looking at the code I expect it should), I have no trouble merging this. Resolving the ShellCheck complaints would be great but it won't block merging.

sonic2kk:
Tested this PR, works as expected with `LD_PRELOAD=""` in the Per-Game and the Global config file., as this is the most visually obvious to test. There is no reason to expect the other variables don't work as well. Thanks a lot for your investigation work to discover this issue and fix it! :smile: 

Once the ShellCheck warnings are addressed, and a `TODO` is added to `saveOrgVars` to clean it up later this PR can be merged.

sonic2kk:
Needs a rebase because of the version bumps in the commits made since this PR was opened. I can do the rebasing if you want, or feel free to do it when adding the ShellCheck directives. :slightly_smiling_face:

HanPrower:
> First, I guess the 8 calls are for different code paths? Or is this called 8 times consectutively when starting a custom program? If so, it's not something for this PR but yikes... I'll need to investigate that!

Basically from https://github.com/sonic2kk/steamtinkerlaunch/blob/master/steamtinkerlaunch#L7247 onwards there are a load of if blocks that all `restoreOrgVars` and then `emptyVars "O" "X"`. It's already marked as a TODO to refactor though.

---

Changes done. Hopefully right :crossed_fingers:

sonic2kk:
The new changes address the comments above, thanks for bumping the version as well!

This can be merged.

sonic2kk:
Thanks again for the work on your recent PRs! It's still a part of being a "maintainer" that I'm getting used to, I really appreciate you working with me :-)

PR #1162, opened 2024-09-03
GameScope: Add support for GameScope `--backend` option

sonic2kk:
Adds a combobox entry field onto the GameScope GUI to configure the `--backend` flag for GameScope. This allows users to choose a GameScope rendering backend, which can be useful to force the SDL backend on Wayland compositors in some instances for example.

Regular usage of this command is shown as, for example, `--backend=sdl`, but `--backend sdl` without the equals sign is also valid. `gamescope --backend=sdl -- glxgears` and `gamescope --backend sdl -- glxgears` are, currently, equivalent. SteamTinkerLaunch, for simplicity, currently uses the latter to forego the equals sign. If this usage ever changes we will need to revisit this.

This is under the "Advanced" section and defaults to `auto`, which is the same as what GameScope currently defaults to. This selects the backend based on the environment. If the option is left at `auto`, SteamTinkerLaunch does not append it to the GameScope arguments, which preserves compatibility with GameScope versions that do not have this flag (although it has been around for a bit, I am late to the party with adding support for this flag).

image

I did some initial testing and this works. At time of writing, there are some obvious tells when a window is using SDL or Wayland, such as the mouse cursor behaving differently on SDL versus Wayland, and Borderless Window option (`-b`) does not work with the Wayland backend but does with the SDL backend. I did not test the other backends.

TODO:
- [x] Further testing
- [ ] Update wiki to note how GameScope chooses default backend
- [x] Langfile updates
- [ ] Version bump


sonic2kk:
Tested further, works as expected with at least `auto`, `sdl`, `wayland` backends. `auto` will choose Wayland here on KDE Plasma Wayland, matching regular GameScope behaviour and usage, and won't append anything to the GameScope arguments `GAMESCOPE_ARGS`. Using the other two backends works as expected, appending to `GAMESCOPE_ARGS` and selecting the correct backend.

This is ready to merge after a version bump and after a wiki update is prepared.

sonic2kk:
This has been tested enough I think, the langfiles have been bumped, a wiki update is ready to go, and the version has been bumped.

Once CI passes, this  can be merged.

PR #1178, opened 2024-09-30
lang: Don't create global config in `loadLanguage`

### LINKED ISSUE #1177: USELUXTORPEDA does not function
## System Information
- SteamTinkerLaunch version:  `v14.0.20240904-1`
- Distribution:  Nobara Linux 40 (KDE Plasma)
- Installation Method:  ProtonUp-Qt

## Issue Description

Sorry to be bothering you with issues today haha. Though this one is more straightforward and not as weird as the custom command issue I posted today.

This one is simple. `USELUXTORPEDA` does not function at all. It looks like SteamTinkerLaunch is looking for `/luxtorpeda/luxtorpeda` which for obvious reasons does not exist. It's not a high priority bug since I can just use Luxtorpeda manually, but since it's an integrated feature, I figured I'd report it.

## Logs

Here's the STL log. Unrelated to this issue, but Half-Life 2: DownFall seems to crash immediately when used with GE-Proton9-15.
https://gist.github.com/Jonathing/0971d56796916dc3bf93107ccaddbd52

sonic2kk:
Fix #1177.

Don't attempt to save and load the global config in `loadLanguage` unless it exists.

Prevents `loadLanguage`, which is called very early, from creating the global config before `setSteamPaths` has a chance to define some paths that are needed to create the global config.

Most notably, this fixes Luxtorpeda and Roberta variables from being incorrect when created too early by `loadLanguage`, as this function creates the global config before `setSteamPaths` has a chance to define `STEAMCOMPATOOLS` which is needed in order to build a valid path for `LUXTORPEDACMD` and `ROBERTACMD`. SteamTinkerLaunch tries to auto-detect Luxtorpeda installations and this has not worked for probably the last 2 years or so, since `loadLanguage` created the global config too soon. However, Luxtorpeda could always be and still can be manually selected by the user, so the only breakage was for the auto-detection.

This may possibly fix other similar cases as well, where Steam paths are needed in the global config but it is created before paths exist.

This fix **will not** retroactively apply to global configs, as we don't know whether the variables are incorrectly generated by us or the user. Bad global configs will need re-generated.

<hr>

This PR needs some testing to ensure the `-f "$STLDEFGLOBALCFG"` check is safe. I think it is, but this will need a bit of further testing.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20241003-1"
+PROGVERS="v14.0.20240930-1 (dont-create-globalconfig-paths-without-steamvars)"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
This does not appear to introduce any regressions with generating the `global.conf` file. I tested a few scenarios going back from this branch to `master`, and with using the `--lang` option. It seems to work fine. I think this is safe to merge, after a rebase and version bump.

PR #1179, opened 2024-09-30
Luxtorpeda: Update default `LUXTORPEDACMD` path

sonic2kk:
Came up in #1177.
Related: #1178.

Luxtorpeda now uses a different launch command; `luxtorpeda.sh` instead of `luxtorpeda`. This PR updates the default path so that it uses this newer one.

This won't retroactively affect global configs, it will only affect newly-created global configs. Not all users may be using a newer Luxtorpeda version that uses this newer entry point, but I think it makes sense for our default value to be what upstream uses as the default. If the path doesn't exist, then it won't show up in the Global Menu Yad box for it, and the user can surmise that the executable was not found and set the path themselves from here or in `global.conf` manually (similar to what I described for the invalid path in #1178).

TODO:
- [x] Check if the same applies for Roberta.
- [ ] Update [Luxtorpeda wiki](https://github.com/sonic2kk/steamtinkerlaunch/wiki/Luxtorpeda) to note the new default path.
- [x] Version bump

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20240930-1 (update-default-luxtorpeda-path)"
+PROGVERS="v14.0.20241003-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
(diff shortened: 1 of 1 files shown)


sonic2kk:
Roberta still uses the same path, I downloaded it with ProtonUp-Qt and confirmed it still uses `run-vm`.

This PR needs a rebase and a version bump, and it is good to be merged.

sonic2kk:
I have an update to the wiki page ready to publish after this is merged. This is ready to go.

sonic2kk:
Wiki page has been updated: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Luxtorpeda/_compare/fef44657e55362f0d094f217fbe33ab83ad6ebaf

PR #1189, opened 2024-10-24
HMM: *VERY* rudimentary support file for Shadow Generations

MaddieMewmews:
Updated misc/hmmgames.txt to add rudimentary support _for Shadow Generations only_. Sonic Generations from Sonic X Shadow Generations (SonicGenerations2024/hedgemmmillersonic) is not yet implemented in HMM.

### DIFF
--- a/misc/hmmgames.txt
+++ b/misc/hmmgames.txt
@@ -6,3 +6,4 @@
 "Sonic Generations";"71340";"32";"hedgemmgens"
 "Sonic Lost World";"329440";"32";"hedgemmlw"
 "Sonic Origins";"1794960";"64";"hedgemmhite"
+"Shadow Generations";"2513280";"64";"hedgemmmillershadow"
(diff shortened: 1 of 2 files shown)


sonic2kk:
Thanks for coming forward and helping with the HedgeModManager stuff! I really appreciate it.

This looks good to me, the sames share the same AppID and prefix but there is a separate executable for Sonic Generations (`SONIC_GENERATIONS.exe`). But since the mime handler is apparently going to be different (`hedgemmmillersonic`) then I think this is fine, we will probably just need a separate entry with the same AppiD. The `dotnet48` stuff should still be fine since Winetricks will just skip the verb if the prefix already has it.

Didn't test but change is correct. Will merge :-) Thanks again!

sonic2kk:
Changelog has been updated with credit: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog/_compare/d7fbb83b159efcef403cbb7185bcd6e37c3f302a

PR #1204, opened 2024-12-26
Avoid bash error when CUPROTOCOMPAT is unset

sparr:
```
% steamtinkerlaunch vortex install
/bin/steamtinkerlaunch: line 4406: [: : integer expression expected
```

This error occurs when `$CUPROTOCOMPAT` is unset. The error does not seem to prevent the script from working.

This PR checks whether it's empty before checking its value.

I'm a bit fuzzy on how to update `PROGVERS` here and would appreciate some guidance if this is to be accepted.

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20241025-1"
+PROGVERS="v14.0.20241228-1"
 PROGCMD="${0##*/}"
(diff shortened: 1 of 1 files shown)


sonic2kk:
This is great, thank you! Sorry for the delay with getting back, holidays and such :smile: 

About `PROGVERS`, just bump it to the current day. For example, `v14.0.20241228-1` would be perfectly fine :-)

sonic2kk:
Many thanks for your contribution, it is much appreciated! Changelog has also been updated to give credit: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Changelog/_compare/a262ad1ba229bd9b161a38ad136a27059185a4f1

PR #1222, opened 2025-02-12
Fix typo in README

Vysp3r:
ProtonPlus is named ProtonUp-Plus which is not right. This commit fixes the typo in question.


sonic2kk:
That's what I get for committing too late at night :sweat_smile: Thanks!

Issue #1208, opened 2024-12-30
Vortex Fails to Launch


vvvilife:
## System Information
- SteamTinkerLaunch version: 12.12.r273.g89af3c8-1
- Distribution: Arch Linux
- Installation Method: aur

## Issue Description
Vortex fails to launch. I have deleted .config/steamtinkerlaunch/vortex, redownloaded, installed, launch fails. as part of this process i noticed stl isn't finding new versions of proton that are installed. output from terminal and logs below.

## Logs
```
❯ steamtinkerlaunch vortex start
/usr/bin/steamtinkerlaunch: line 19523: [: X: integer expression expected
Mon Dec 30 03:44:19 PM EST 2024 SKIP - setModWine - No download URL found for requested 'GE-Proton9-16' - skipping
Mon Dec 30 03:44:19 PM EST 2024 ERROR - setModWine - DNPROTON was not found - can't continue
sed: -e expression #1, char 0: no previous regular expression
/usr/bin/steamtinkerlaunch: line 19523: [: X: integer expression expected
/usr/bin/steamtinkerlaunch: line 19523: [: X: integer expression expected
Mon Dec 30 03:44:20 PM EST 2024 INFO - startVortex - Starting Vortex without options
```

[stl.log](https://github.com/user-attachments/files/18277690/stl.log)
[installVortex_protonrun.log](https://github.com/user-attachments/files/18277689/installVortex_protonrun.log)

HalHatesDave:
NTA but I'd suggest starting with manually installing the newest release. The AUR package was last updated to 12.12 which is >1y old. The version (as of today) after building from master is steamtinkerlaunch-v14.0.20241228-1.

More details here: https://github.com/sonic2kk/steamtinkerlaunch/wiki/Installation#manual-installation

EDIT: Also, if Vortex fails to install dotnet6 while using v14 even after setting Vortex to use Steam Linux Runtime (either by using the STL global menu or setting `VORTEXUSESLR="1"` in global.conf) as happened to me. You can try to run `WINEPREFIX=/home/$USER/.config/steamtinkerlaunch/vortex/compatdata/pfx winetricks dotnet6` to install it manually.

sonic2kk:
Closing as per the above, issues pertaining to v12.12 are invalid.

sparr:
Twice in two weeks I've installed 12.12 from arch binary package, found a bug report about this problem, then remembered I need to install from git.

[...]

PR #1254, opened 2025-09-06
[Refactor] Syntax Correction: single to double square bracket, missing ws, inconsistent tabbing.

ninetailedtori:
This should fix all the bracket syntax issues, given that the file is being declared to run in bash, we should conform to using the widely superior native syntax of `[[]]`, rather than an invocation of test with single brackets, as can be seen here failing as well (https://github.com/sonic2kk/steamtinkerlaunch/issues/1208#issuecomment-3263263341). It also regularises spacing after `; then` and similar snippets, as well as tabbing.

Do let me know if there are any issues that need to be patched! ❤️

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -171,7 +171,7 @@ SGDBRETRIES="5"
 
 ### default vars ###
 
-if [ -z "$XDG_CONFIG_HOME" ]; then
+if [[ -z "$XDG_CONFIG_HOME" ]]; then
 	STLCFGDIR="$HOME/.config/${PROGNAME,,}"													# either hardcoded config dir
 else
 	STLCFGDIR="$XDG_CONFIG_HOME/${PROGNAME,,}"												# or in XDG_CONFIG_HOME if the user set the variable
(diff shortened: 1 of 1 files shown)


ninetailedtori:
Have tested on my end, the syntax corrections seem to be functioning flawlessly, everything functions fine. Would love a double check though c:

ninetailedtori:
This one needs a bit of debugging, I'm just going through it right now to make sure everything is fixed.

ninetailedtori:
This has the file fully functioning for me, but I'd love a double check, because I may well have missed some.

ninetailedtori:
In all honesty, it would be good to create an automatic workflow to swap all tabs with 4-spaces, in order to preserve uniformity among editors, and thus increase future compatibility with commits. Or, if you'd want the other way round, this would work for tabbing:
1. any 4-spaces replaced with a tab.
2. any tabs with 2 spaces following should be rounded down to a tab
3. any tabs with 3 spaces rounded up to 2 tabs.
4. any tabs with 2 spaces preceding should be rounded down to a tab
5. any tabs with 3 spaces preceding should be rounded up to 2 tabs

Issue #1259, opened 2025-09-18
[FEATURE] Toggle for Proton Wayland


Nano-Ocelot:
## System Information
- SteamTinkerLaunch version: v14.0.20250602
- Distribution: OpenSUSE Tumbleweed 

## Feature Description
It would be nice to have a quick and easy toggle somewhere in Proton Options for enabling Proton Wayland support for those using Proton versions that support Wayland. The ENV I'm referencing is `PROTON_ENABLE_WAYLAND=1`

sonic2kk:
This would be a really nice feature. Not planning to work on this and if that changes I'll post an updated comment, so in the meantime anyone is welcome to contribute this!

ninetailedtori:
idm tossing this up rq! it's a "spring cleaning" feature request :]

PR #1267, opened 2025-10-31
Addition of PROTON_ENABLE_WAYLAND toggle

ninetailedtori:
Requires translations for all languages aside from EN_US, EN_GB, and FR_FR.


ninetailedtori:
Ref: #1259

ninetailedtori:
I'll be honest. Uncertain why my side has absolutely no change to the menu, it is a single bash file isn't it? But I'm fairly certain I added it EVERYWHERE it needed to be? Is there smth I have to call to update the steamtinkerlaunch settings list?

PR #1271, opened 2025-11-21
Add pacman post-update hook - wipe protoncsv

ninetailedtori:
Wipes protoncsv.txt every time it's updated by pacman, to force recaching of proton versions.

Also, ignore .idea (jetbrains IDE project folder.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .vscode/
+.idea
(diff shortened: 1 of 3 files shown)


ninetailedtori:
We could possibly add it to steamtinkerlaunch directly, but it's hard to know when a package updates, unless we decide, say, "if we hit a mismatch, we should delete the protoncsv file and rebuild it once, and if it still mismatches, it's the user's fault".

Issue #1275, opened 2025-12-13
Shaders are not being downloaded from the shader repos.


SiAshbery:
## System Information
- SteamTinkerLaunch version: v12.12
- Distribution: SteamOS
- Installation Method: Protonup-QT

## Issue Description
When I installed STL the only available shader repo was `astrayfx`. I assume that, as outlined here https://github.com/sonic2kk/steamtinkerlaunch/wiki/Shader-Repositories, the `repocustomlist.txt` was not found and so the template was copied over.

I manually added the repos provided in the example from the above page to the `repocustomlist.txt` using desktop mode.

When I restarted steam in gaming mode and ran the game I am trying to reshade I found that the `Shader Repo` list was fully populated, but in `Game Shaders`, the only option available was `astrayfx`

I've tried restarting, enabling/disabling various repos, enabling/disabling shader downloads as well as other settings, but the shaders will not download.

Thanks in advanced for any help.

## Logs

![Image](https://github.com/user-attachments/assets/e8451551-9053-45bb-bda1-e8ba61719fdb)
![Image](https://github.com/user-attachments/assets/07e5a42f-c8b0-41d7-999b-c96206fd4894)
![Image](https://github.com/user-attachments/assets/fb1e7a23-e13f-4d2a-8126-a8b16b10fc01)

zany130:
your using an outdated version latest is 14.0.2.250602 


but anyway the issue is still present even on the latest git it looks like the code for downloading the shader repos broke... ill see if I can take a look

PR #1279, opened 2025-12-27
Fix ReShade shader repo fetching via MediaWiki API

zany130:
### Summary
Fixes shader repository list generation by switching the ReShade repo fetch
to the MediaWiki API instead of scraping the HTML page.

### Background
PCGamingWiki no longer reliably allows direct HTML fetching of the ReShade page.
This causes `repolist.txt` to remain empty and results in only the hardcoded
`astrayfx` entry being available (see #1275 ).

### Changes
- Replace the existing repo-fetch logic inside `createShaderRepoList`
  with a MediaWiki API–based implementation.
- No new files or dependencies added.
- Bash-only change limited to the createShaderRepoList function.
- Uses jq (already an optional dependency) to parse MediaWiki API JSON.

### Testing
- Verified locally that `repolist.txt` is populated correctly.
- Custom entries in `repocustomlist.txt` are now respected.


zany130:
Works great and actually fixed both the repos not being grabbed from PC gaming wiki and also fixes repos add to the custom repo file not downloading

PR #1314, opened 2026-07-08
yad: remove --image-on-top and --decorated/--undecorated (Yad >=15.0 compat)

starter3xd:
## Problem

Yad 15.0 no longer silently ignores unknown command-line options. Instead, it exits
with code 255 and prints `Unable to parse command line: Unknown option`. This breaks
every Yad dialog in SteamTinkerLaunch.

STL was passing two options that were removed from Yad long ago:

- **`--image-on-top`** — 32 occurrences across most dialog functions
- **`--decorated` / `--undecorated`** — 51 occurrences via the `$WINDECO` variable, set in `prepareGUI()`

On Yad ≤14.2 these were silently ignored. On Yad ≥15.0 every dialog crashes.

## Fix

1. Removed all `--image-on-top` from Yad command lines (32×)
2. Removed the `$WINDECO` mechanism from `prepareGUI()` — both `--decorated` and
   `--undecorated` are no longer passed to Yad (51×)
3. Cleaned up two `echo`-generated dynamic Yad functions that also contained these options
4. `USEWINDECO` config variable is preserved as a no-op for backward compatibility

## Testing

Tested on both old and new Yad with Proton dropdown, form fields, checkboxes, and buttons:

| Yad version | Before fix | After fix |
| --- | --- | --- |
| **14.2** (stable/LTS distros) | Exit 0 (options silently ignored) | Exit 0 |
| **15.0** (rolling/bleeding edge) | Exit 255 — crashes on `--image-on-top` | Exit 0 |

- `bash -n steamtinkerlaunch` passes cleanly
- Dialog tested with dropdown (CB), checkboxes (CHK), text fields (TXT), and buttons

## Notes

- Users who previously used `USEWINDECO=0` to disable window decorations on X11
  will need to configure this at the window manager level instead
- No logic changes, no new dependencies
- 58 insertions, 61 deletions across ~50 hunks

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20250602"
+PROGVERS="v14.0.20260708-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -784,7 +784,7 @@ function OpenWikiPage {
 		else
 			TITLE="${PROGNAME}-Wiki"
 			pollWinRes "$TITLE"
-			"$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --center "$WINDECO" --html --uri="$WIKURL" "$GEOM" >/dev/null 2>/dev/null
+			"$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --center --html --uri="$WIKURL" "$GEOM" >/dev/null 2>/dev/null
 		fi
 	fi
 }
@@ -805,7 +805,7 @@ function StatusWindow {
     $RUNFUNC |
     while read -r line; do
 		echo "# ${line}";
-	done | "$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --progress --progress-text="$1..." --pulsate --center --no-buttons --auto-close "$WINDECO" "$GEOM"
+	done | "$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --progress --progress-text="$1..." --pulsate --center --no-buttons --auto-close "$GEOM"
 }
 
 function setColGui {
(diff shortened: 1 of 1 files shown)


Danielduel:
comment to:
starter3xd/yad15-fix HEAD 7b5d45e64e2d98e6cbc09cac95140a411f48de53,
master (HEAD-1) 8550ab26a712b7f5f6d0947070181446b9de61fd

I encounter this problem on arch after general system update.
I've `which steamtinkerlaunch` and it is linked for me to `/usr/bin/steamtinkerlaunch`
I've removed AUR steamtinkerlaunch and
```
$ git clone https://github.com/sonic2kk/steamtinkerlaunch
$ cd steamtinkerlaunch
$ git remote add starter3xd https://github.com/starter3xd/steamtinkerlaunch
$ git checkout -b fix-test
$ git pull starter3xd yad15-fix
```

From what I can see - there are redactions of `$WINDECO` and `--image-on-top` + a notice that these options don't work anymore.

My god, bash is scary.

Testing this fix.

```
$ chmod +x
# ln -s ..../steamtinkerlaunch /usr/bin/steamtinkerlaunch
```

Tested `editor`, `launcher`, `settings`.
It looks like it works fine.
Launched Beat Saber through this fix, works great.

LGTM

Marks20125:
I really hope this gets merged, I need stl :(

OzzyHelix:
LGTM

KoJesko:
Thanks for tracking this down — I hit the same crash on Arch (Yad 15.0) and opened #1320 before
spotting this PR, which got here well before mine. Apologies for the duplicate.

One thing I wanted to flag directly, since it affects what this PR drops: `--undecorated` does not
appear to have been removed from Yad. Only `--decorated` was — decorated is the Yad 15 default now.
On 15.0 here:

```console
$ yad --version
15.0 (GTK+ 3.24.52)
$ yad --help-all | grep -i decor
  --undecorated                                  Set window undecorated
```

and it still works, rather than just parsing:

```console
$ yad --form --undecorated --field=A --field=B v1 v2   # -> v1|v2|
$ yad --form --decorated   --field=A --field=B v1 v2   # -> Unable to parse command line: Unknown option --decorated
```

So `USEWINDECO=0` could keep working on X11 instead of becoming a no-op — it'd only need
`WINDECO="--undecorated"` retained and the `--decorated` branch dropped. That would also let the
"no longer supported" warning stay as the existing Wayland-only warning.

Not trying to push my version over yours — this is a small change on top of what you already have,
and it'd save the behaviour change called out in your Notes section. Happy to close #1320 either
way.

CyanEngineer:
A quick workaround while we're waiting for an official fix:
- Open a terminal and run `yad --write-settings` (creates a settings file at `$HOME/.config/yad/settings.conf`)
- Open the settings file
- Set `IgnoreUnknownOptions` to `true` (this was the default behaviour before yad 15.0)
- Set `Debug` to `true` (for some reason yad ignores `IgnoreUnknownOptions` when it isn't in debug mode)

Now you should be good to go!

PR #1320, opened 2026-07-30
Fix all Yad dialogs failing to open on Yad 15.0

KoJesko:
## Problem

Yad 15.0 removed the `--image-on-top` and `--decorated` options. Yad refuses to start when it is
given an option it does not recognise:

```
Unable to parse command line: Unknown option --image-on-top
```

SteamTinkerLaunch passes `--image-on-top` in 32 places and `--decorated` via `WINDECO`, so on
Yad 15.0 **every** SteamTinkerLaunch window closes the instant it is opened — the main menu, the
game settings menu, the category menus, all of them. To the user it looks like STL simply does
nothing when a game is launched.

It shows up in the log as the menu opening and closing within the same second, followed by the
window poll timing out:

```
INFO - prepareMenu - Opening Menu
INFO - goBackToPrevFunction - 'MainMenu' closed
SKIP - updateWinRes - Timeout waiting for Window 'SteamTinkerLaunch-MainMenu'
```

This affects any distribution that has moved to Yad 15.0 (Arch and derivatives already have).

## Fix

`prepareGUI` now asks Yad which options it supports and only passes the ones that exist.

Feature detection is used rather than a version comparison so the behaviour is also correct for
distro patches, forks, and future Yad releases. If the query fails for any reason both options are
simply left out — that costs only cosmetics and never prevents a window from opening.

Decorated windows are the default in Yad 15.0, so omitting `--decorated` there is equivalent to the
previous behaviour.

## Why arrays

`WINDECO` becomes an array, and the new `YADIMGTOP` array replaces the hardcoded `--image-on-top`.

This matters for correctness, not style. An unsupported option has to disappear from the command
line *entirely*. Substituting an empty string would keep Yad running but silently corrupt every form
dialog, because Yad treats the empty argument as a positional one and consumes it as the first
`--field` value, shifting every following value by one. In the main menu the `--field` values are the
commands the buttons run, so the buttons would quietly start invoking each other's commands.

Measured on Yad 15.0, `--form --field=A --field=B v1 v2`:

| `WINDECO` | Result |
| --- | --- |
| `("--decorated")` — current code | `Unable to parse command line: Unknown option --decorated` |
| `""` — naive fix | `\|v1\|` — **field values shifted** |
| `()` — this PR | `v1\|v2\|` — correct |
| `("--undecorated")` — this PR | `v1\|v2\|` — correct |

An empty array contributes zero arguments, which is exactly equivalent to not passing the option.

As a side effect this also removes a latent version of the same bug: previously, any code path that
reached a Yad form before `prepareGUI` had run would have passed `""` and shifted the field values.

## Testing

Tested on Arch (CachyOS), Yad 15.0 (GTK+ 3.24.52), Wayland session.

- Before: `steamtinkerlaunch settings <appid>` exits immediately, one `Unable to parse command line`
  error, no window.
- After: menus render and stay open, zero parse errors, and the log records what was detected:
  ```
  INFO - prepareGUI - Yad has no '--image-on-top' option (removed in Yad 15.0) - omitting it
  INFO - prepareGUI - Yad has no '--decorated' option (removed in Yad 15.0) - using the Yad default instead
  ```
- Form field values verified unshifted (table above).
- `shellcheck --extended-analysis=false steamtinkerlaunch` — clean, v0.10.0 as used by CI.
- `bash -n` — clean.

I do not have a pre-15.0 Yad to hand to test the other branch of the detection, so a second pair of
eyes on that path would be welcome. The logic keeps passing both options whenever Yad advertises
them, so older Yad should be unaffected.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/steamtinkerlaunch
+++ b/steamtinkerlaunch
@@ -7,7 +7,7 @@
 PREFIX="/usr"
 PROGNAME="SteamTinkerLaunch"
 NICEPROGNAME="Steam Tinker Launch"
-PROGVERS="v14.0.20250602"
+PROGVERS="v14.0.20260730-1"
 PROGCMD="${0##*/}"
 PROGINTERNALPROTNAME="Proton-stl"
 SHOSTL="stl"
@@ -784,7 +784,7 @@ function OpenWikiPage {
 		else
 			TITLE="${PROGNAME}-Wiki"
 			pollWinRes "$TITLE"
-			"$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --center "$WINDECO" --html --uri="$WIKURL" "$GEOM" >/dev/null 2>/dev/null
+			"$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --center "${WINDECO[@]}" --html --uri="$WIKURL" "$GEOM" >/dev/null 2>/dev/null
 		fi
 	fi
 }
@@ -805,7 +805,7 @@ function StatusWindow {
     $RUNFUNC |
     while read -r line; do
 		echo "# ${line}";
-	done | "$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --progress --progress-text="$1..." --pulsate --center --no-buttons --auto-close "$WINDECO" "$GEOM"
+	done | "$YAD" --window-icon="$STLICON" --title="$TITLE" --on-top --progress --progress-text="$1..." --pulsate --center --no-buttons --auto-close "${WINDECO[@]}" "$GEOM"
 }
 
 function setColGui {
@@ -1615,7 +1615,7 @@ function getSteamGridDBArtworkGUI {
 	# It will use the Global Menu default, but also allows the user to specify a different action this time
 	#
 	# FSGDBAW = Fetcch SteamGridDB ArtWork :-)
-	FSGDBAWGUISET="$("$YAD" --f1-action="$F1ACTION" --window-icon="$STLICON" --form --scroll --center --on-top "$WINDECO" \
+	FSGDBAWGUISET="$("$YAD" --f1-action="$F1ACTION" --window-icon="$STLICON" --form --scroll --center --on-top "${WINDECO[@]}" \
 	--title="$TITLE" --separator="|" --image="$SHOWPIC" \
 	--text="$(spanFont "$(strFix "$GUI_FSGDBAW" "$FSGDBAW_HEADERTITLE")" "H")\n${DESC_FSGDBAW}" \
 	--field=" ":LBL " " \
@@ -2128,7 +2128,7 @@ function openGameLauncher {
 	function GLgui {
 		LISTDIR="$DFDIR/$1"
 		if [ -d "$LISTDIR" ] && [ "$(find "$LISTDIR" -name "*.desktop" | wc -l)" -ge 1 ]; then
-			"$YAD" --f1-action="$F1ACTION" --icons --window-icon="$STLICON" --read-dir="$LISTDIR" "$WINDECO" --title="$TITLE" --single-click --keep-icon-size --center --compact --sort-by-name "$GEOM"
+			"$YAD" --f1-action="$F1ACTION" --icons --window-icon="$STLICON" --read-dir="$LISTDIR" "${WINDECO[@]}" --title="$TITLE" --single-click --keep-icon-size --center --compact --sort-by-name "$GEOM"
 		else
 			writelog "SKIP" "${FUNCNAME[0]} - No games found in '$LISTDIR' or directory itself not found"
 		fi
@@ -2165,7 +2165,7 @@ function openGameLauncher {
 	else
 		if [ "$1" == "menu" ]; then
 			createCollectionList
-			CATMENU="$("$YAD" --f1-action="$F1ACTION" --window-icon="$STLICON" --center "$WINDECO" --form --scroll --separator="\n" --quoted-output \
+			CATMENU="$("$YAD" --f1-action="$F1ACTION" --window-icon="$STLICON" --center "${WINDECO[@]}" --form --scroll --separator="\n" --quoted-output \
 			--text="$(spanFont "$GUI_CHOOSECAT" "H")" \
 			--field="$GUI_CHOOSECATS":CB "$(cleanDropDown "installed" "$CATLIST")" \
 			--title="$TITLE" "$GEOM"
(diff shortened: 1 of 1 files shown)


KoJesko:
Heads up: this overlaps with #1314, which I only spotted after opening this — apologies for the
duplicate. #1314 got here first and does fix the crash, so I'm happy to close this in its favour.

Before that, one substantive difference worth putting in front of whoever reviews.

#1314 removes the `WINDECO` mechanism entirely and turns `USEWINDECO` into a no-op, on the stated
basis that `--decorated`/`--undecorated` were both removed from Yad. Only `--decorated` was removed.
`--undecorated` is still there in Yad 15.0:

```console
$ yad --version
15.0 (GTK+ 3.24.52)
$ yad --help-all | grep -i decor
  --undecorated                                  Set window undecorated
```

And it still functions — passing it to a Yad 15.0 form dialog works and returns the expected field
values:

```console
$ yad --form --undecorated --field=A --field=B v1 v2   # -> v1|v2|   (no error)
$ yad --form --decorated   --field=A --field=B v1 v2   # -> Unable to parse command line: Unknown option --decorated
```

So `USEWINDECO=0` can keep working on X11 instead of becoming a no-op. This PR passes each option
only when the installed Yad advertises it, which drops exactly what Yad rejects and nothing else.

The reason the options are passed via arrays here rather than being blanked to `""` is that Yad
takes an empty argument as a positional one and consumes it as the first `--field` value, shifting
every later value by one — in the main menu those values are the commands the buttons run. #1314
avoids this by deleting the option text outright, so it isn't affected; noting it only because it
constrains any fix that tries to keep the options conditional.

Happy to go whichever way you prefer — close this, or rebase it onto #1314 as a follow-up that
restores `--undecorated`.

Aeryn1023:
I was able to download this and replace the original steamtinkerlaunch file, and I can confirm it works. Here's to hoping it gets merged

KoJesko:
For now, I created my own fix at https://github.com/KoJesko/stl-yad15-fix

CyanEngineer:
A quick workaround while we're waiting for an official fix:
- Open a terminal and run `yad --write-settings` (creates a settings file at `$HOME/.config/yad/settings.conf`)
- Open the settings file
- Set `IgnoreUnknownOptions` to `true` (this was the default behaviour before yad 15.0)
- Set `Debug` to `true` (for some reason yad ignores `IgnoreUnknownOptions` when it isn't in debug mode)

Now you should be good to go!

KoJesko:
> A quick workaround while we're waiting for an official fix:
> 
>     * Open a terminal and run `yad --write-settings` (creates a settings file at `$HOME/.config/yad/settings.conf`)
> 
>     * Open the settings file
> 
>     * Set `IgnoreUnknownOptions` to `true` (this was the default behaviour before yad 15.0)
> 
>     * Set `Debug` to `true` (for some reason yad ignores `IgnoreUnknownOptions` when it isn't in debug mode)
> 
> 
> Now you should be good to go!

That's a much simpler fix than the patch I drafted, thanks for that!

Do you want me to add this to my repo? As instructions instead of a patch?

CyanEngineer:
> That's a much simpler fix than the patch I drafted, thanks for that!
> 
> Do you want me to add this to my repo? As instructions instead of a patch?

I just shared the workaround on a few issues and PRs to give people a quickfix that doesn't require building an STL fork - I actually hadn't realized that you had also shared a patch script that makes your solution really accessible as well.
Feel free to share the workaround around if you want, it was by no means an attempt to "one-up" your patch :)

KoJesko:
> > That's a much simpler fix than the patch I drafted, thanks for that!
> > 
> > Do you want me to add this to my repo? As instructions instead of a patch?
> 
> I just shared the workaround on a few issues and PRs to give people a quickfix that doesn't require building an STL fork - I actually hadn't realized that you had also shared a patch script that makes your solution really accessible as well.
> Feel free to share the workaround around if you want, it was by no means an attempt to "one-up" your patch :) 

Did not think it was ;)

But anyways, if you do not use yad for anything besides stl, reverting to 12 is also an option!

I havent tested the patch on all oses, so im likely going to suggest this. A bit late now so.

QguAr71:
Additional confirmation on Arch + a caveat about the `yad-gtk2` workaround that may save others some time.

### Patch verification on Arch Linux

Tested on Arch, `yad 15.0 (GTK+ 3.24.52)`, Steam via Steam Linux Runtime, two games (Cyberpunk 2077, appid 1091500) and re-tested the Wait Requester path.

Result after applying this PR on top of `master`:

- All `SteamTinkerLaunch-*` dialogs now open. MainMenu renders at the expected geometry (1692x687) instead of timing out.
- `updateWinRes` reports `Window 'SteamTinkerLaunch-MainMenu' is running - polling the resolution` instead of the previous `Timeout waiting for Window`.
- No `Unknown option --image-on-top` / `--decorated` errors left in the log.

```
# before
updateWinRes - Timeout waiting for Window 'SteamTinkerLaunch-MainMenu'
X Error: Unable to parse command line: Unknown option --image-on-top (exit 255)

# after
updateWinRes - Window 'SteamTinkerLaunch-MainMenu' is running - polling the resolution
```

### The patch applies to `master`, not to 12.12 (AUR)

Worth flagging, because the PR base is `master` (`PROGVERS v14.0.20250602`). On the current AUR release `12.12` the patch **does not apply cleanly** — I got 28/55 hunks rejected, and the surviving 14 `image-on-top` sites are in completely different locations. Anyone on the AUR package needs to pull `master` first:

```bash
curl -sL https://raw.githubusercontent.com/sonic2kk/steamtinkerlaunch/master/steamtinkerlaunch -o steamtinkerlaunch
curl -sL https://github.com/sonic2kk/steamtinkerlaunch/pull/1320.diff -o pr1320.diff
sed -i 's|a/steamtinkerlaunch|a/steamtinkerlaunch|;s|b/steamtinkerlaunch|b/steamtinkerlaunch|' pr1320.diff
patch -p0 steamtinkerlaunch < pr1320.diff
```

After patching, `PROGVERS` reads `v14.0.20260730-1` and `grep -c image-on-top` drops from 32 to 4 (all inside the new `YADHELP` gated block).

### `yad-gtk2` (0.42) is *not* a viable workaround

I tried this first to avoid touching STL. `yad-gtk2` 0.42 does still ship `--image-on-top` (`src/option.c`), so it looks like a drop-in fix — but STL enforces `MINYAD="7.2"` and `0.42.0` fails that version comparison (`sort -V`), so STL aborts with *"Your Yad version is too old"*. There is currently **no yad release that satisfies both constraints**: the option was added in 0.40 and the version string stayed `0.x` until it jumped to 7.x, by which point `--image-on-top` was already being phased out. So the version-gated patch / `YADHELP` check is the only clean path.

Thanks for the fix — confirmed working here.

QguAr71:
Follow-up from testing this PR on Arch (yad 15.0): with it applied **all** dialogs open correctly — MainMenu, Wait Requester, settings — so the crash described here is fully fixed.

One **separate** sizing bug remains after patching, in the launcher only: `GLgui` passes `$GEOM` (`--geometry=WxH`) to `yad --icons`, and yad 15 ignores `--geometry` for icon boxes, so the icon grid collapses to ~231x135. `--width`/`--height` do work. I split this out as #1324 so it does not muddy this PR:

https://github.com/sonic2kk/steamtinkerlaunch/issues/1324

Not a blocker for this PR — just flagging it so reviewers know the launcher still needs a follow-up.

